PR #24, opened 2016-08-08
Add GCM_ENABLED flag to manifest.

zbsz:
Disabled GCM for all internal builds to test web socket fallback.

With GCM_ENABLED set to 'false' client will not register for GCM notifications, but will always use our own web socket connection. 
This connection is meant to be used in cases where Play Services are not available, but for now we want to force it on internal builds to test how it affects the device.


hakonbo:
build again

PR #44, opened 2016-08-16
[AN-4391] Disable tracking

hakonbo:
#### APK
[Download build #7349](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7349/artifact/build/artifact/wire-dev-PR44-7349.apk)
[Download build #7351](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7351/artifact/build/artifact/wire-dev-PR44-7351.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/MainActivity.java
+++ b/app/src/main/java/com/waz/zclient/MainActivity.java
@@ -18,6 +18,7 @@
 package com.waz.zclient;
 
 import android.app.NotificationManager;
+import android.content.Context;
 import android.content.DialogInterface;
 import android.content.Intent;
 import android.content.res.Configuration;
@@ -62,6 +63,7 @@ import com.waz.zclient.controllers.tracking.events.exception.ExceptionEvent;
 import com.waz.zclient.controllers.tracking.events.otr.VerifiedConversationEvent;
 import com.waz.zclient.controllers.tracking.events.profile.SignOut;
 import com.waz.zclient.controllers.tracking.screens.ApplicationScreen;
+import com.waz.zclient.controllers.userpreferences.UserPreferencesController;
 import com.waz.zclient.core.api.scala.AppEntryStore;
 import com.waz.zclient.core.controllers.tracking.attributes.RangedAttribute;
 import com.waz.zclient.core.controllers.tracking.events.Event;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/waz/zclient/MainActivity.java
```diff
@@ -220,9 +222,14 @@ protected void onResume() {
...
-                                             getControllerFactory().getUserPreferencesController().getDeviceId(),
-                                             getControllerFactory().getTrackingController());
+        final boolean trackingEnabled = getSharedPreferences(UserPreferencesController.USER_PREFS_TAG, Context.MODE_PRIVATE)
+            .getBoolean(getString(R.string.pref_advanced_analytics_enabled_key), true);
+
+        if (trackingEnabled) {
+            HockeyCrashReporting.checkForCrashes(getApplicationContext(),
+                                                 getControllerFactory().getUserPreferencesController().getDeviceId(),
+                                                 getControllerFactory().getTrackingController());
+        }
```
zbsz: So when it's disabled then we won't upload crashes, but will ever delete the generated files? We may end up with multiple crash reports in files dir that are never deleted.

hakonbo: We may get some exceptions there during current app session, but from next restart the UncaughtExceptionHandler won't be registered and no more exceptions stored.

hakonbo: Ah, I guess you might refer to the manually saved exceptions? I'll have another look and see if we can address that too.

zbsz: SE also saves some exceptions using `ExceptionHandler.saveException`.
We also have `HockeyCrashReporter` class, it was supposed to be used to upload crashes, but seems it's not used

hakonbo: Good point. I'll add a delete call for every startup.

hakonbo:
@zbsz Added stack trace purge

zbsz:
Should we also add a check in `NativeCrashReporter`?

hakonbo:
Not sure I follow regarding NativeCrashReporter. But if you suggest we should delete the native crash dumps created by breakpad, I agree. Will add a delete method to NativeCrashManager. Or were you thinking of something else?

zbsz:
I was thinking about making sure that it doesn't upload when tracking is disabled.
Didn't notice that this is only done inside of `checkFroCrashes`, so its fine.
👍

hakonbo:
Cool, pushed the native crash deletion stuff now...

PR #45, opened 2016-08-16
Update play services verification dialog

zbsz:
Changed to not display error dialog if play services are missing or disabled.
Most of the time this means that user doesn't want or can't run Play Services.
We don't want to bother those users with a dialog. 
Alternative solution would be to show a dialog with 'never again' checkbox, 
but it seems like too much work for a small gain.
#### APK
[Download build #7350](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7350/artifact/build/artifact/wire-dev-PR45-7350.apk)


hakonbo:
lgtm

PR #58, opened 2016-08-17
[AN-4300] Show infinite loading bar when there is no server connectivity

deanrobertcook:
#### Ticket:
https://wearezeta.atlassian.net/browse/AN-4300

#### Test description:
Pro tip, try and connect to hm-wifi-guest to trick your device into thinking it has internet, but to prevent it from being able to connect to the servers. Then you can test both use cases.
#### APK
[Download build #7388](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7388/artifact/build/artifact/wire-dev-PR58-7388.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/pages/main/connectivity/ConnectivityFragment.java
+++ b/app/src/main/java/com/waz/zclient/pages/main/connectivity/ConnectivityFragment.java
@@ -23,18 +23,22 @@ import android.view.View;
 import android.view.ViewGroup;
 import android.widget.ImageView;
 import com.waz.zclient.R;
+import com.waz.zclient.controllers.accentcolor.AccentColorObserver;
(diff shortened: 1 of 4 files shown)


mykola-mokhnach:
On tablet this connecting progress bar is never replaced with NO INTERNET bar. This works properly on mobile phone through.

deanrobertcook:
@mykola-mokhnach That doesn't make a lick of sense. Lemme check it out...

deanrobertcook:
@mykola-mokhnach Yeah, I think that's just a question of the device's connectivity state. Sometimes the device figures out it's not actually connected to the internet, in which case you get the yellow bar. Other times it keeps on trying and believes it has a connection (sometimes for a while) and so you might see the loading bar for a long time.

mykola-mokhnach:
Then OK if you say this is normal behaviour

PR #88, opened 2016-08-25
[AN-4439] Fix for NPE from Hockey

shumengye:
#### Ticket
https://wearezeta.atlassian.net/browse/AN-4439
#### APK
[Download build #7444](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7444/artifact/build/artifact/wire-dev-PR88-7444.apk)
[Download build #7532](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7532/artifact/build/artifact/wire-dev-PR88-7532.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/pages/main/pickuser/PickUserFragment.java
+++ b/app/src/main/java/com/waz/zclient/pages/main/pickuser/PickUserFragment.java
@@ -364,6 +364,8 @@ public class PickUserFragment extends BaseFragment<PickUserFragment.Container> i
         }
 
         loadStartUi();
+        usersSearchModelObserver.resumeListening();
+        usersSearchModelObserver.forceUpdate();
     }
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/waz/zclient/pages/main/pickuser/PickUserFragment.java
```diff
@@ -364,6 +364,7 @@ public void run() {
         }
 
         loadStartUi();
+        usersSearchModelObserver.resumeListening();
```
deanrobertcook: It might be possible that we would get an update in between `onStop` and `onResume` that the modelObserver would miss. 

Not sure if that's a big issue, but if it is, then we might need to `forceUpdate()`

shumengye:
@deanrobertcook I pushed change

PR #294, opened 2016-11-08
[AN-4659] Added possibility to set no sound

JPonte:
Added the possibility for each of the 3 possible notification sounds to
be silent.
Ticket: https://wearezeta.atlassian.net/browse/AN-4659
#### APK
[Download build #7851](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7851/artifact/build/artifact/wire-dev-PR294-7851.apk)
[Download build #7854](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7854/artifact/build/artifact/wire-dev-PR294-7854.apk)
[Download build #7855](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7855/artifact/build/artifact/wire-dev-PR294-7855.apk)
[Download build #7856](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7856/artifact/build/artifact/wire-dev-PR294-7856.apk)
[Download build #7857](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7857/artifact/build/artifact/wire-dev-PR294-7857.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/api/scala/ScalaMediaStore.java
+++ b/app/src/main/java/com/waz/zclient/api/scala/ScalaMediaStore.java
@@ -86,43 +86,46 @@ public class ScalaMediaStore extends MediaStore {
 
     @Override
     public void setCustomSoundUrisFromPreferences(SharedPreferences preferences) {
-        String value = preferences.getString(context.getString(R.string.pref_options_ringtones_ringtone_key), null);
-        if (value != null) {
-            setCustomSoundUri(R.raw.ringing_from_them, value);
-            if (RingtoneUtils.isDefaultValue(context, value, R.raw.ringing_from_them)) {
+        //The value of the preference is the uri for the sound resource
+        //If this value is an empty string, then it should be silent
+        //If the preference doesn't exist yet (null) we assume the default value instead
+        String customSoundUri = preferences.getString(context.getString(R.string.pref_options_ringtones_ringtone_key), null);
+        if (customSoundUri != null) {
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/waz/zclient/api/scala/ScalaMediaStore.java
```diff
@@ -85,7 +87,7 @@ public void setCustomSoundUri(@RawRes int resourceId, String uri) {
     @Override
     public void setCustomSoundUrisFromPreferences(SharedPreferences preferences) {
         String value = preferences.getString(context.getString(R.string.pref_options_ringtones_ringtone_key), null);
-        if (!TextUtils.isEmpty(value)) {
+        if (value != null) {
```
shumengye: `TextUtils.isEmpty()`is checking for null

JPonte: But it's also checking for empty string, which should be valid in this case because it's the value returned by the Preference for the silent option.

shumengye: ok, perhaps add a comment clarifying the values

xuanzhaopeng:
Should not set "None" for Sounds of Ringtone , Text tone , Ping as default.

xuanzhaopeng:
Issue 2: In the sounds selection popup, the item called "Silent" , but after I select it, in the Options menu, we show it as "None",  we should keep to use same name

PR #299, opened 2016-11-09
Fix for crash when receiving a video call when permission for video c…

jspittka:
…amera was denied. AUDIO-1069, AN-4421.


vova-lohika:
👍

PR #337, opened 2016-11-24
[Asset V3] Use Asset V2 as default option in Developer Settings

xuanzhaopeng:
Please review @deanrobertcook 

cc @vova-lohika  @red-avtovo 
#### APK
[Download build #7966](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7966/artifact/build/artifact/wire-dev-PR337-7966.apk)

### DIFF
--- a/app/src/main/res/xml/preferences_developer.xml
+++ b/app/src/main/res/xml/preferences_developer.xml
@@ -54,7 +54,7 @@
         />
 
     <CheckBoxPreference
-        android:defaultValue="true"
+        android:defaultValue="false"
         android:key="@string/pref_dev_assets_v3_key"
         android:title="@string/pref_dev_assets_v3_title"
         />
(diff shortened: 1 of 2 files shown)


deanrobertcook:
@xuanzhaopeng we should be able to leave this as set to true. This will ensure that debug (and internal builds) use v3 by default. However, in release builds, where this checkbox does not appear, SE will look for this preference in the system, and if it does not exist, set it to false by default. 

Please just keep an eye out in the candidate build and let me know if this mechanism doesn't work as intended - then we need to change it.

xuanzhaopeng:
@deanrobertcook , We build exp build by : (from Android Jenins)
./gradlew --refresh-dependencies clean 
./gradlew assembleExperimentalRelease || true
./gradlew assembleExperimentalRelease

Thus, I don't have permission to :
1) Change the preference from adb (As this build is in Release Mode)
2) I'm not sure to change this preference for each test case is good idea, cause it will take extra time.

The possible solution:
1) If it is release build, I should not see Developer settings at all.
2) We use build config in in build.gradle to set default value for the preference(Like Testing Gallery)

deanrobertcook:
Ah yeah, I see - it's actually independent of whether it's a debug build or not...

As discussed - make a broadcast receiver that can listen for a custom intent - much like what we have with candidate builds and auto-answer calls

PR #358, opened 2016-11-30
[AN-4684] Usernames in search results

JPonte:
The previous PR didn’t include the the row views from the search result, which are different from the ones in the contact list.


JPonte:
build again

PR #363, opened 2016-12-01
[AN-4738] Send on enter with BT keyboard

hakonbo:
#### APK
[Download build #8020](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8020/artifact/build/artifact/wire-dev-PR363-8020.apk)


hakonbo:
build again

xuanzhaopeng:
Test it on bt keyboard, works fine when I switch to "Disable Send button",  and enter in BT keyboard works.

BUT, Text sketch will never works when I use BT keyboard. I add bug label, let @bjoernherbig to decide whether it should be fixed inside of this PR.

bufke:
Should that apk download link work publicly? I'm happy to test it out.

hakonbo:
@xuanzhaopeng We'll sketch out of it for now.
@bufke Thanks, but our build server and artefacts are internal only for the time being. This should make it into our next public release though.

PR #380, opened 2016-12-06
[AN-4745] Fix take over text, tablet drawable and size of attempts

JPonte:
#### APK
[Download build #8053](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8053/artifact/build/artifact/wire-dev-PR380-8053.apk)


xuanzhaopeng:
I see conversation view instead of Shield view in Tablet. 

STR:
1.Tablet Portrait
2. Login with a user without User Name
3. I see Takeover in left part

ER:
I see shield in right part
AR:
I see first conversation

PR #399, opened 2016-12-12
Forces user detail view update on common connections change

JPonte:
Since the user model observer doesn’t update itself for common
connections.


shumengye:
@JPonte 
sure will merge for RC but can you create ticket to move this common connection logic to SE?

shumengye:
@JPonte 
`UserDetailsView` is used in more places like conversation options menu

PR #403, opened 2016-12-13
Corrects value of secondary text color in dark theme

shumengye:
Makes secondary text color value same as in user's device tab
#### APK
[Download build #8122](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8122/artifact/build/artifact/wire-dev-PR403-8122.apk)

### DIFF
--- a/wire-core/src/main/res/values/colors.xml
+++ b/wire-core/src/main/res/values/colors.xml
@@ -127,7 +127,7 @@
     <color name="separator_light">@color/light_graphite_40</color>
 
     <color name="text__secondary_dark__40">@color/white_40</color>
-    <color name="text__secondary_dark">@color/white_24</color>
+    <color name="text__secondary_dark">@color/white_40</color>
     <color name="text__primary_dark_40">@color/white_40</color>
     <color name="text__primary_dark">@color/white</color>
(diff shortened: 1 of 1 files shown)


wire-core/src/main/res/values/colors.xml
```diff
@@ -127,7 +127,7 @@
     <color name="separator_light">@color/light_graphite_40</color>
 
     <color name="text__secondary_dark__40">@color/white_40</color>
-    <color name="text__secondary_dark">@color/white_24</color>
+    <color name="text__secondary_dark">@color/white_40</color>
```
hakonbo: The `text__secondary_dark` field is used quite a few places. Are you sure you want to update all of them?

shumengye: @hakonbo Yes, confirmed with @markesaaremets. The correct value is actually white with 56% opacity, but we are using 40% in registration and a lot of other places so this would be the quickest way to at least make the text more readable.

shumengye:
build again

wire-core/src/main/res/values/colors.xml
hakonbo: cool

PR #405, opened 2016-12-13
[AN-4772] Usernames issues

JPonte:
Ticket: https://wearezeta.atlassian.net/browse/AN-4772
Also closes the take over screen if the user eventually gets an handle
(from other device for example).
#### APK
[Download build #8129](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8129/artifact/build/artifact/wire-dev-PR405-8129.apk)


xuanzhaopeng:
Checked (Works perfect)
1) Emoji keyboard works perfect now(it will reject all keyword emoji)
2) Checked a user who has 2 devices, and see takeover screen in both devices, and takeover will be disappeared immediately when I set Handle from one device.

Issue #233, opened 2016-10-12
Bring the app to f-droid


bootofood:
Please make it possible to compile the app without the non-free dependencies.
In this way it could be included in f-droid, finally beating Signal!

someoneEsle:
https://github.com/wireapp/wire-android/issues/5
Please close.

bootofood:
Thanks. That issue is closed, that's why I didn't find it. Good to read that they're interested in publishing the app in f-droid, certainly a more friendly and reasonable approach than what we got for Signal.

Shouldn't a "f-droid issue" stay open until we've got it in there, complying with the inclusion policy?

someoneEsle:
I don't know what the policy as for keeping issues open is, but it looks like something they're aware of, therefore there's no need to let the ticket open (which may be why they closed the other one). I'm pretty sure they'll remember even if this issue gets closed.
(I understand your point about leaving an issue open though, maybe they could leave the original one open as it has the official reply too)

bjoernherbig:
Thanks guys for bringing this up again and yes, we will remember it for sure. As for closing it, we will leave this one open since there are likely to come in more requests for f-droid.

h-2:
Out of curiosity: what are the proprietary dependencies blocking f-droid acceptance? Seems like if you already have a fallback for GCM you are almost there or not? Maybe other people could help in this process?

legovini:
@h-2 take a look here:

https://github.com/wireapp/wire-android/issues/5#issuecomment-234947466

and compiling without GCM is probably non trivial, even if there is already a fallback mechanism.

h-2:
I am posting a 50€ bounty for someone who creates a version of the Android Wire client that fully meets F-Droid’s criteria, i.e. no proprietary dependencies. Preferably the changes would also meet upstream’s criteria for integration into the official repository, e.g. via a Flavor or build flag or something like that.

I can send the money via Paypal or WireTransfer/SEPA.

lipis:
💶 💵 💶 💵 💶 💵

grote:
The first step, before anybody can take this on would be to get rid of the [custom maven repository](https://dl.bintray.com/wire-android/third-party/com/wire/) and get those dependencies submitted to jcenter.

Currently, these libraries are affected:
* gradle-android-scala-plugin - non-upstreamed fork?
* icu4j-shrunk - non-upstreamed fork?
* robotest_2.11 - non-upstreamed fork?
* spotify-auth - non-free software?
[...]

PR #411, opened 2016-12-16
Introduce free (libre) buildTypes

bgermann:
Two buildTypes releaseFree and debugFree are added to
have support for differentiating between builds with
and without non-free dependencies.

This is needed as a first step to resolve issue #233.


AndroidBob:

eighthave:
From what I've seen, this solves the last issue preventing Wire going into F-Droid.  @wireapp: does it work for you to have this version, without Google Play Services jars and Spotify, built and distributed via F-Droid?

PR #457, opened 2017-01-09
[AN-4828] Double tap to like message

hakonbo:
#### APK
[Download build #8209](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8209/artifact/build/artifact/wire-dev-PR457-8209.apk)
[Download build #8216](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8216/artifact/build/artifact/wire-dev-PR457-8216.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/messages/MessageViewPart.scala
+++ b/app/src/main/scala/com/waz/zclient/messages/MessageViewPart.scala
@@ -1,20 +1,20 @@
 /**
- * Wire
- * Copyright (C) 2016 Wire Swiss GmbH
- *
- * This program is free software: you can redistribute it and/or modify
- * it under the terms of the GNU General Public License as published by
- * the Free Software Foundation, either version 3 of the License, or
- * (at your option) any later version.
- *
- * This program is distributed in the hope that it will be useful,
- * but WITHOUT ANY WARRANTY; without even the implied warranty of
- * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
- * GNU General Public License for more details.
- *
- * You should have received a copy of the GNU General Public License
- * along with this program.  If not, see <http://www.gnu.org/licenses/>.
- */
+  * Wire
+  * Copyright (C) 2017 Wire Swiss GmbH
+  *
+  * This program is free software: you can redistribute it and/or modify
+  * it under the terms of the GNU General Public License as published by
+  * the Free Software Foundation, either version 3 of the License, or
+  * (at your option) any later version.
+  *
+  * This program is distributed in the hope that it will be useful,
+  * but WITHOUT ANY WARRANTY; without even the implied warranty of
+  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+  * GNU General Public License for more details.
+  *
+  * You should have received a copy of the GNU General Public License
+  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+  */
 package com.waz.zclient.messages
 
 import android.content.Context
@@ -45,11 +45,14 @@ import org.threeten.bp.{Instant, LocalDateTime, ZoneId}
 
 trait MessageViewPart extends View {
   val tpe: MsgPart
+  protected val messageAndLikes = Signal[MessageAndLikes]()
 
   def set(msg: MessageData, part: Option[MessageContent], opts: MsgBindOptions): Unit
 
-  def set(msg: MessageAndLikes, part: Option[MessageContent], opts: MsgBindOptions): Unit =
+  def set(msg: MessageAndLikes, part: Option[MessageContent], opts: MsgBindOptions): Unit = {
     set(msg.message, part, opts)
+    messageAndLikes ! msg
+  }
 
   //By default disable clicks for all view types. There are fewer that need click functionality than those that don't
   this.onClick {}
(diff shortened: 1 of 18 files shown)


app/src/main/scala/com/waz/zclient/messages/MessageViewPart.scala
```diff
@@ -45,27 +62,64 @@ import org.threeten.bp.{Instant, LocalDateTime, ZoneId}
...
+  }
+
+  this.onClick (
+    {
+      // Single click
+      onClicked ! ({})
+      getParent.asInstanceOf[View].performClick()
+    },
+    {
+      // Double click
```
hakonbo: This section needs a cleanup...

app/src/main/scala/com/waz/zclient/utils/package.scala
```diff
@@ -62,6 +63,11 @@ package object utils {
       override def onClick(v: View): Unit = f
     })
 
+    def onClick(f1: => Unit, f2: => Unit): Unit = view.setOnClickListener(new OnDoubleClickListener {
```
deanrobertcook: best to give more meaningful names to the functions if there are a couple of them, like `onSingleClick` or `onDoubleClick`

deanrobertcook: And then, if you use [currying](http://danielwestheide.com/blog/2013/01/30/the-neophytes-guide-to-scala-part-11-currying-and-partially-applied-functions.html), you can avoid having to wrap the two functions in another set of parenthesis, 

```
def onClick(f1: => Unit)(f2: =>Unit) = {...}
```

can then be called like so:
```
this.onClick { myOnClickFunction } { myOnDoubleClickFunction }
```
(without wrapping both function calls in another separate set of parenthesis separated by a comma - pretty small improvement, but just a nice style thing)

Off the top of my head though, I'm not sure how that'll behave with overloading...

app/src/main/scala/com/waz/zclient/messages/MessageViewPart.scala
```diff
@@ -45,27 +62,64 @@ import org.threeten.bp.{Instant, LocalDateTime, ZoneId}
...
+  private val likable = message map {
+    case m if m.msgType == Message.Type.ANY_ASSET => true
+    case m if m.msgType == Message.Type.ASSET => true
+    case m if m.msgType == Message.Type.AUDIO_ASSET => true
+    case m if m.msgType == Message.Type.LOCATION => true
+    case m if m.msgType == Message.Type.TEXT => true
+    case m if m.msgType == Message.Type.TEXT_EMOJI_ONLY => true
+    case m if m.msgType == Message.Type.RICH_MEDIA => true
+    case m if m.msgType == Message.Type.VIDEO_ASSET => true
+    case _ => false
```
deanrobertcook: What you can do is pull these types out and make a set of them in a companion object of `ClickableViewPart`, just under the definition of the `ClickableViewPart` trait, like so: 

```
object ClickableViewPart {
  import Message.Type._
  val likableTypes = Set(
    ANY_ASSET,
    ASSET,
    AUDIO_ASSET,
    LOCATION,
    TEXT,
    TEXT_EMOJI_ONLY,
    RICH_MEDIA,
    VIDEO_ASSET
  )

  def isLikeable(mt: Message.Type) = likableTypes.contains(mt)
}
```

Then, back in the trait itself, you can do:
```
trait ClickableViewPart extends MessageViewPart with ViewHelper {
  import ClickableViewPart._
  ...
  protected val message = Signal[MessageData]()
  message.disableAutowiring() //important to ensure the signal keeps updating itself in the absence of any listeners
  ...
  // Double click
      if (message.currentValue.map(_.msgType).exists(isLikeable)) {
        for (reacts <- reactions.head) {
          if (likedByMe.currentValue.get) reacts.unlike(message.currentValue.get.convId, message.currentValue.get.id)
          else reacts.like(message.currentValue.get.convId, message.currentValue.get.id)
          getParent.asInstanceOf[View].performClick()
        }
      } 
 ...
}
```

deanrobertcook: P.S. - I like that you're trying to use signals though!

hakonbo:
build again

app/src/main/scala/com/waz/zclient/messages/MessageViewPart.scala
```diff
@@ -45,30 +45,73 @@ import org.threeten.bp.{Instant, LocalDateTime, ZoneId}
...
+  }
+
+  def onSingleClick = {
     onClicked ! ({})
     getParent.asInstanceOf[View].performClick()
   }
 
+  def onDoubleClick = {
+    if (message.currentValue.map(_.msgType).exists(isLikeable)) {
+      for (reacts <- reactions.head) {
```
deanrobertcook: I'm not sure if relying on the future (`.head`) here is a good idea. Basically, if the user double clicks, using a future will guarantee that the action will be carried out, even if it takes an arbitrary amount of time for the future to complete (potentially blocking the UI thread)

Instead, in order to keep the action responsive, it might be better to use `currentValue`. That said, if the value's not there, then nothing will happen, but that - in general - is probably better for things that respond to user interactions. 

Also, to avoid the `.get`s further down in the block, we can chain monadic stuff (futures or options) together in the same for-comprehension:

```
for {
      msg <- message.currentValue
      if isLikeable(msg.msgType)
      reacts <- reactions.currentValue
      likedByMe <- likedByMe.currentValue
    } {
      if (likedByMe) reacts.unlike(msg.convId, msg.id)
      else reacts.like(msg.convId, msg.id)
      getParent.asInstanceOf[View].performClick() //perform click to change focus
    }
```

This also has the advantage of not evaluating later parts of the for-comprehension if any level of it returns an empty option

deanrobertcook: Actually, I just noticed that we already use `.head` in other parts to achieve similar functionality, so I guess it's fine, but the for-comprehension can still be used with `.head` instead to make things a little tidier :)

red-avtovo:
https://wearezeta.atlassian.net/browse/AN-4828
Heart doesn't become red on 4.4.2

deanrobertcook:
@red-avtovo that issue isn't related to the changes made here - will need to be fixed in another commit

hakonbo:
build again

hakonbo:
build again

PR #459, opened 2017-01-10
[AN-ROBIN] Add id for items introduced from msg stream

xuanzhaopeng:
Copy the id which already implemented in collection branch into master. Which will make some conflicts when you merge collection back into master for them.

All those ids I added are all introduced in Collection branch. The reason why I copy it back to master because maybe we will have RC only for message stream without Collections, and also no needs to ask @Joao to rebase collection on master everyday. 
#### APK
[Download build #8204](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8204/artifact/build/artifact/wire-dev-PR459-8204.apk)
[Download build #8207](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8207/artifact/build/artifact/wire-dev-PR459-8207.apk)

### DIFF
--- a/app/src/main/res/layout/message_audio_asset.xml
+++ b/app/src/main/res/layout/message_audio_asset.xml
@@ -19,6 +19,7 @@
 
 -->
 <com.waz.zclient.messages.parts.assets.AudioAssetPartView xmlns:android="http://schemas.android.com/apk/res/android"
+                                                          android:id="@+id/message_audio_asset"
                                                           android:layout_width="match_parent"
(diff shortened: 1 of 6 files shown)


deanrobertcook:
I don't think we need the suffixes in front of each id, i.e. the `ytpv__` part in front of each id. We've not been keeping up with this style throughout the rest of `message_stream`

xuanzhaopeng:
@deanrobertcook  Hey Dean,  I remove all prefix

PR #479, opened 2017-01-14
[AN-4024][AN-4848] Fix marking message as read.

wire-android:
Messages should not be marked as read when messages list is not actually open.


AndroidBob:

hakonbo:
Cherry-picked to rel/2.25.x

PR #482, opened 2017-01-14
[AN-4873] Fix MatchError in AudioAssetPartView.

wire-android:
Events subscriber can not be a partial function, it has to be exhaustive (handle all cases).


AndroidBob:

hakonbo:
Cherry-picked to rel/2.25.x

PR #491, opened 2017-01-16
[AN-SHIELD] Wrong shield icon for unverify system messages

hakonbo:
#### APK
[Download build #8279](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8279/artifact/build/artifact/wire-dev-PR491-8279.apk)


xuanzhaopeng:
Now the shield works properly,  the shield in conversation view and in people details infor view is only shown when his devices and my devices are all verified.

PR #492, opened 2017-01-17
Single image lower toolbar

JPonte:
#### APK
[Download build #8283](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8283/artifact/build/artifact/wire-dev-PR492-8283.apk)
[Download build #8286](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8286/artifact/build/artifact/wire-dev-PR492-8286.apk)
[Download build #8290](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8290/artifact/build/artifact/wire-dev-PR492-8290.apk)
[Download build #8300](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8300/artifact/build/artifact/wire-dev-PR492-8300.apk)
[Download build #8301](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8301/artifact/build/artifact/wire-dev-PR492-8301.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/controllers/Base$$ControllerFactory.java
+++ b/app/src/main/java/com/waz/zclient/controllers/Base$$ControllerFactory.java
@@ -21,6 +21,7 @@ import android.app.Activity;
 import android.content.Context;
 import android.view.View;
 
+import com.waz.zclient.Injectable;
 import com.waz.zclient.WireContext;
 import com.waz.zclient.controllers.accentcolor.AccentColorController;
 import com.waz.zclient.controllers.accentcolor.IAccentColorController;
@@ -88,8 +89,6 @@ import com.waz.zclient.pages.main.conversationpager.controller.ISlidingPaneContr
 import com.waz.zclient.pages.main.conversationpager.controller.SlidingPaneController;
 import com.waz.zclient.pages.main.pickuser.controller.IPickUserController;
 import com.waz.zclient.pages.main.pickuser.controller.PickUserController;
-import java.lang.IllegalStateException;
-import java.lang.Override;
 
 public abstract class Base$$ControllerFactory implements IControllerFactory {
   protected IAccentColorController accentColorController;
@@ -633,7 +632,7 @@ public abstract class Base$$ControllerFactory implements IControllerFactory {
     public ICollectionsController getCollectionsController() {
         verifyLifecycle();
         if (collectionsController == null) {
-            collectionsController = new CollectionController(((WireContext) this.context).injector());
+            collectionsController = ((Injectable) this.context).inject(CollectionController.Manifest(), ((WireContext) this.context).injector());
         }
         return collectionsController;
     }
--- a/app/src/main/java/com/waz/zclient/controllers/Base$$ControllerFactory.java
+++ b/app/src/main/java/com/waz/zclient/controllers/Base$$ControllerFactory.java
@@ -162,6 +162,7 @@ public abstract class Base$$ControllerFactory implements IControllerFactory {
   protected boolean isTornDown;
 
   protected Context context;
+  protected Activity activity;
 
   public Base$$ControllerFactory(Context context) {
     this.context = context;
@@ -549,6 +550,7 @@ public abstract class Base$$ControllerFactory implements IControllerFactory {
 
   @Override
   public void setActivity(Activity activity) {
+    this.activity = activity;
     getGlobalLayoutController().setActivity(activity);
     getOrientationController().setActivity(activity);
     getSpotifyController().setActivity(activity);
@@ -632,7 +634,7 @@ public abstract class Base$$ControllerFactory implements IControllerFactory {
     public ICollectionsController getCollectionsController() {
         verifyLifecycle();
         if (collectionsController == null) {
-            collectionsController = ((Injectable) this.context).inject(CollectionController.Manifest(), ((WireContext) this.context).injector());
+            collectionsController = CollectionController.injectedCollectionController(((Injectable) activity), ((WireContext) activity).injector());
         }
         return collectionsController;
     }
--- a/app/src/main/scala/com/waz/zclient/conversation/CollectionController.scala
+++ b/app/src/main/scala/com/waz/zclient/conversation/CollectionController.scala
@@ -123,6 +123,7 @@ class StubCollectionController extends ICollectionsController{
 object CollectionController {
 
   val GridColumns = 4
+  val Manifest = manifest[CollectionController]
 
   trait ContentType {
     val msgTypes: Seq[Message.Type]
--- a/app/src/main/scala/com/waz/zclient/conversation/CollectionController.scala
+++ b/app/src/main/scala/com/waz/zclient/conversation/CollectionController.scala
@@ -31,9 +31,6 @@ import com.waz.zclient.{Injectable, Injector}
 
 trait ICollectionsController {
 
-  val focusedItem: SourceSignal[Option[MessageData]]
-  val conversationName: Signal[String]
-
   def openCollection(): Unit
 
   def closeCollection(): Unit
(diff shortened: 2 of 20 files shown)


JPonte:
build again

app/src/main/scala/com/waz/zclient/views/SingleImageViewToolbar.scala
```diff
@@ -0,0 +1,117 @@
...
+  likeButton.setPressedBackgroundColor(ContextCompat.getColor(getContext, R.color.light_graphite))
+  downloadButton.setPressedBackgroundColor(ContextCompat.getColor(getContext, R.color.light_graphite))
+  shareButton.setPressedBackgroundColor(ContextCompat.getColor(getContext, R.color.light_graphite))
+  deleteButton.setPressedBackgroundColor(ContextCompat.getColor(getContext, R.color.light_graphite))
+  viewButton.setPressedBackgroundColor(ContextCompat.getColor(getContext, R.color.light_graphite))
+
+  likeButton.onClick( message.foreach(msg => messageActionCallback.onAction(MessageAction.LIKE, msg)))
+  downloadButton.onClick( message.foreach(msg => messageActionCallback.onAction(MessageAction.SAVE, msg)))
+
+  shareButton.onClick( message.foreach(msg => messageActionCallback.onAction(MessageAction.FORWARD, msg)))
```
wire-android: It's a bit weird that you use `messageActionCallback` here, why don't you access `messageActionsController.onMessageAction` directly?
`Callback` class is legacy stuff that's only needed for java intergration, so it would be better to avoid it when possible. In this case I would only create instance of it when creating `MessageBottomSheetDialog`.

wire-ui/src/main/res/layout/single_image_view_toolbar_layout.xml
```diff
@@ -0,0 +1,123 @@
...
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
wire-android: I would suggest to use `<merge>` here. Using `LinearLayout` will result in two nested linear layouts.
https://developer.android.com/training/improving-layouts/reusing-layouts.html

app/src/main/scala/com/waz/zclient/views/SingleImageViewToolbar.scala
```diff
@@ -0,0 +1,117 @@
...
+    case _ => Signal.empty[MessageAndLikes]
+  }
+
+  messageSig.on(Threading.Ui){ updateLikeButton }
+
+  zms.on(Threading.Ui){ z =>
+    z.msgAndLikes.onUpdate{
+      _ => updateSig.mutate(!_)
+    }
+  }
```
wire-android: This is wrong on so many levels :(
1. Why `Threading.Ui` ? you should only enforce UI thread when you really need it, otherwise use background thread or don't specify thread at all. No point in using precious UI time in here.
2. `msgAndLikes.onUpdate` will fire every time any msg changes, you don't need to reload your message so often.
3. `CollectionController.focusedItem` should either care about msg updates itself, or should be changed to store `MessageId`, this way it would be explicit that clients need to care about it

Seems that the only info that you actually need to have updated is `likedBySelf` property, and you could retrieve it directly from reactions storage, like that:
```
  val likedBySelf = collectionController.focusedItem flatMap {
    case Some(m) => zms.flatMap { z => 
      z.reactionsStorage.signal((m.id, z.selfUserId)).map(_.action == Liking.like).orElse(Signal const false) 
    }
    case None => Signal.const(false)
  }
```

app/src/main/scala/com/waz/zclient/views/SingleImageViewToolbar.scala
```diff
@@ -0,0 +1,117 @@
...
+
+  messageActionsController.onDeleteConfirmed.on(Threading.Ui){
+    _ => collectionController.focusedItem ! None
+  }
+
+  private def message: Option[Message] = messageSig.currentValue.map(ZMessaging.currentUi.messages.cachedOrUpdated(_))
```
wire-android: instead of creating the message every time you could just have a signal which is updated when focused msg changes, `Message` will handle updates internally.
```
  val message = collectionController.focusedItem.map(_.map(_.id)) collect {
    case Some(id) => ZMessaging.currentUi.messages.cachedOrNew(id)
  }
  message.disableAutowiring() // to make sure the value is available in `message.currentValue`

File: app/src/main/scala/com/waz/zclient/conversation/CollectionController.scala [hunk 5]
```diff
@@ -69,9 +66,11 @@ class CollectionController(implicit injector: Injector) extends Injectable with
 
   val conversation = zms.zip(currentConv) flatMap { case (zms, convId) => zms.convsStorage.signal(convId) }
 
-  override val conversationName = conversation map (data => if (data.convType == IConversation.Type.GROUP) data.name.filter(!_.isEmpty).getOrElse(data.generatedName) else data.generatedName)
+  val conversationName = conversation map (data => if (data.convType == IConversation.Type.GROUP) data.name.filter(!_.isEmpty).getOrElse(data.generatedName) else data.generatedName)
+
+  val focusedItem: SourceSignal[Option[MessageData]] = Signal(None)
 
-  override val focusedItem: SourceSignal[Option[MessageData]] = Signal(None)
+  val targetItem: SourceSignal[Option[MessageData]] = Signal(None)
```
wire-android: This should rather be an `EventStream` judging by how you use it in messages list. 
If you modify a signal inside of it's listener then there is something wrong.

app/src/main/java/com/waz/zclient/controllers/Base$$ControllerFactory.java
```diff
@@ -633,7 +634,7 @@ public IUsernamesController getUsernameController() {
     public ICollectionsController getCollectionsController() {
         verifyLifecycle();
         if (collectionsController == null) {
-            collectionsController = new CollectionController(((WireContext) this.context).injector());
+            collectionsController = CollectionController.injectedCollectionController(((Injectable) activity), ((WireContext) activity).injector());
```
wire-android: You could use `BaseScalaActivity.injectJava(CollectionController.class)`

wire-ui/src/main/res/layout/single_image_view_toolbar_layout.xml
JPonte: Not in this case. The parent linear layout is vertical and we need this one.

app/src/main/java/com/waz/zclient/controllers/Base$$ControllerFactory.java
JPonte: Right, that was already fixed in another PR.

app/src/main/scala/com/waz/zclient/conversation/CollectionController.scala
JPonte: Makes sense, will do.

wire-ui/src/main/res/layout/single_image_view_toolbar_layout.xml
wire-android: I don't mean the layout in fragment where this toolbar is used, I mean the `SingleImageViewToolbar` itself.
`inflate` method inside toolbar will create whatever is devined in xml (LinearLayout in this case) and insert it as a child of `SingleImageViewToolbar`. Use hierarchy viewer and inspect the toolbar to see what I mean.

app/src/main/scala/com/waz/zclient/views/SingleImageViewToolbar.scala
JPonte: Yea, not proud of this one...
I think I was updating the view inside that one, hence the Ui thread, but then changed it and forgot about it.
Will fix it.

PR #499, opened 2017-01-18
[AN-4804] Add tracking for collections

JPonte:
Ticket: https://wearezeta.atlassian.net/browse/AN-4804
#### APK
[Download build #8298](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8298/artifact/build/artifact/wire-dev-PR499-8298.apk)
[Download build #8320](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8320/artifact/build/artifact/wire-dev-PR499-8320.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/tracking/Event.scala
+++ b/app/src/main/scala/com/waz/zclient/tracking/Event.scala
@@ -112,3 +112,39 @@ case class PlayedAudioMessageEvent(fileMimeType: String, durationSec: Int, playe
 
 case class PlayedYouTubeMessageEvent(playedByReceiver: Boolean, isOtto: Boolean, convType: ConversationType)
   extends MessageEvent("media.played_youtube_message", playedByReceiver, isOtto, convType)
+
+abstract class CollectionsEvent(name: String, conversationType: ConversationType, withBot: Boolean) extends Event(name){
+  protected val baseAttributes = Map(
+    WITH_BOT          -> String.valueOf(withBot),
+    CONVERSATION_TYPE -> conversationType.name
+  )
+}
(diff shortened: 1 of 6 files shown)


app/src/main/scala/com/waz/zclient/tracking/Event.scala
```diff
@@ -112,3 +112,39 @@ case class PlayedAudioMessageEvent(fileMimeType: String, durationSec: Int, playe
 
 case class PlayedYouTubeMessageEvent(playedByReceiver: Boolean, isOtto: Boolean, convType: ConversationType)
   extends MessageEvent("media.played_youtube_message", playedByReceiver, isOtto, convType)
+
+abstract class CollectionsEvent(name: String, conversationType: ConversationType, withBot: Boolean) extends Event("collections." + name){
+  protected val baseAttributes = Map(
```
deanrobertcook: haha, for some reason doing this didn't occur to me when I needed it above thanks!

app/src/main/scala/com/waz/zclient/tracking/Event.scala
```diff
@@ -112,3 +112,39 @@ case class PlayedAudioMessageEvent(fileMimeType: String, durationSec: Int, playe
 
 case class PlayedYouTubeMessageEvent(playedByReceiver: Boolean, isOtto: Boolean, convType: ConversationType)
   extends MessageEvent("media.played_youtube_message", playedByReceiver, isOtto, convType)
+
+abstract class CollectionsEvent(name: String, conversationType: ConversationType, withBot: Boolean) extends Event("collections." + name){
```
deanrobertcook: I think it might be a bit clearer (and easier to search) if the name is just written in full for each specific Event type (that is, just moving the `"collections."` to the individual case classes) , or else we should go back and do the same for the ones above - this file will get very big very quickly as we start to move more stuff from Java to Scala and we should be consistent

PR #519, opened 2017-01-22
Changed wifi download pref check.

wire-android:
Moved preference access logic to a val in 'AssetsController' and disabled
autowiring for that signal, this way value will be precomputed.

Calling `Signal.currentValue` on freshly created signal will often fail,
most of the time signals will be initially empty. In this case preference
is loaded in background, so it would only work if value is already cached.

We should pay more attention to warning logs generated in `Signal.currentValue`, 
calling that on unwired signal is a bit risky. 
Actually `currentValue` should be avoided whenever possible, use `Signal.head` instead.
#### APK
[Download build #8324](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8324/artifact/build/artifact/wire-dev-PR519-8324.apk)


AndroidBob:

hakonbo:
build again

PR #530, opened 2017-01-24
use different text for generic error

tomkukral:
Use different error for generic registration error and invalid
registration error.

Same error can confuse users and make find error cause more difficult.


AndroidBob:

bjoernherbig:
@infotexture it's used

PR #540, opened 2017-01-26
Sync generic registration error strings

infotexture:
Align with copy from phone registration variants:
* `new_reg_phone_generic_error_header`
* `new_reg_phone_generic_error_message`

Closes #530.
#### APK
[Download build #8350](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8350/artifact/build/artifact/wire-dev-PR540-8350.apk)


infotexture:
To facilitate re-use, we could instead consider directly referencing these variants via `@string`, but I’m not sure whether such dependencies are desirable in this case, nor which variants should be source/target.

Please advise.

PR #571, opened 2017-02-03
[AN-4950] Closes preview only if conversation changes to another

JPonte:
Ticket: https://wearezeta.atlassian.net/browse/AN-4950
#### APK
[Download build #8410](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8410/artifact/build/artifact/wire-dev-PR571-8410.apk)
[Download build #8412](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8412/artifact/build/artifact/wire-dev-PR571-8412.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/pages/main/conversation/ConversationFragment.java
+++ b/app/src/main/java/com/waz/zclient/pages/main/conversation/ConversationFragment.java
@@ -698,7 +698,7 @@ public class ConversationFragment extends BaseFragment<ConversationFragment.Cont
             conversationModelObserver.setAndUpdate(toConversation);
         }
 
-        if (isPreviewShown) {
+        if (isPreviewShown && fromConversation != null && !toConversation.getId().equals(fromConversation.getId())) {
             onCancelPreview();
         }
 


app/src/main/java/com/waz/zclient/pages/main/conversation/ConversationFragment.java
```diff
@@ -698,7 +698,7 @@ public void onCurrentConversationHasChanged(final IConversation fromConversation
             conversationModelObserver.setAndUpdate(toConversation);
         }
 
-        if (isPreviewShown) {
+        if (isPreviewShown && fromConversation != null || !toConversation.getId().equals(fromConversation.getId())) {
```
hakonbo: I guess we can get a NPE from `fromConversation.getId()` and would be good to add some more parentheses to make it explicitly clear the the two sides of `||` are evaluated separately.

JPonte: Oops... actually it's supposed to be a && instead of a ||...

PR #575, opened 2017-02-05
[AN-4894] Remove `DelayedViewEventContext`.

wire-android:
Replaced with `ViewEventContext` without delay.
Delay here was meant to improve performance during view recycling, we wanted to avoid context restarts if view is only temporarily detached.
This way, some parts of view didn't need to be refreshed, as context restart forces all signals to fire.
Unfortunately with this delay some updates are executed when view is detached, and on some Android versions those views would not be properly updated (views ignore updates when detached). Removing delay avoids this issue, makes UI more predictable.

Note: this will affect performance in messages stream, from my testing it seems fine, but we should check on slower devices and monitor that. If it turns out to be a problem then we can reimplement view recycling to avoid detaching views when not needed.  
#### APK
[Download build #8416](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8416/artifact/build/artifact/wire-dev-PR575-8416.apk)
[Download build #8418](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8418/artifact/build/artifact/wire-dev-PR575-8418.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/WireContext.scala
+++ b/app/src/main/scala/com/waz/zclient/WireContext.scala
@@ -54,34 +54,21 @@ trait ViewFinder {
   def stub[V <: View](id: Int) : V = findById[ViewStub](id).inflate().asInstanceOf[V]
 }
 
-trait DelayedViewEventContext extends View with EventContext {
-  import com.waz.ZLog.ImplicitTag._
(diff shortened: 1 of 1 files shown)


AndroidBob:

hakonbo:
build again

hakonbo:
build again

PR #730, opened 2017-03-21
Convince Benny to use Wire Internal

hakonbo:
#### APK
[Download build #8685](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8685/artifact/build/artifact/wire-dev-PR730-8685.apk)


red-avtovo:
do we really need to push that code into production?

bennyn:
The Android system does not think it is a good decision to download this build. 😂

a

PR #736, opened 2017-03-22
Updated to AVS 3.3.22

hakonbo:
#### APK
[Download build #8698](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8698/artifact/build/artifact/wire-dev-PR736-8698.apk)
[Download build #8699](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8699/artifact/build/artifact/wire-dev-PR736-8699.apk)

### DIFF
--- a/testing_gallery/build.gradle
+++ b/testing_gallery/build.gradle
@@ -5,8 +5,8 @@ android {
     buildToolsVersion rootProject.ext.buildToolsVersion
 
     signingConfigs {
-        if (file("../app/signing.gradle").exists()) {
-            apply from: '../app/signing.gradle'
+        if (file("signing.gradle").exists()) {
+            apply from: 'signing.gradle'
         }
     }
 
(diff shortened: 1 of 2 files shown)


testing_gallery/build.gradle
```diff
@@ -5,8 +5,8 @@ android {
     buildToolsVersion rootProject.ext.buildToolsVersion
 
     signingConfigs {
-        if (file("../app/signing.gradle").exists()) {
-            apply from: '../app/signing.gradle'
```
red-avtovo: What's wrong with that?

red-avtovo:
C1 @ Android 5.1.1 / Android 6.0.1
C1 @ Android 6.0.1 / Android 5.1.1

C1 makes a video call to C2
C2 answeres
C2 sees label (C1 turned of the video), but both of them see the video

Happens in 40% of cases

testing_gallery/build.gradle
hakonbo: This makes builds fail unless you have the certificates in the testing_gallery as well. Feel free to use the signing.file, but it's better to copy it to testing_gallery and it will only be used for those who want/need it.

PR #763, opened 2017-04-03
[AN-5148] Avoid oom when creating big bitmaps

JPonte:
This should avoid getting oom with big bitmaps. These should be eventually replaced with the AssetDrawable.
Ticket: https://wearezeta.atlassian.net/browse/AN-5148

#### APK
[Download build #8732](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8732/artifact/build/artifact/wire-dev-PR763-8732.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/pages/main/backgroundmain/views/BackgroundDrawable.java
+++ b/app/src/main/java/com/waz/zclient/pages/main/backgroundmain/views/BackgroundDrawable.java
@@ -136,6 +136,9 @@ public class BackgroundDrawable extends Drawable {
     private void createBackgroundShader() {
         if (bitmap == null) {
             bitmap = createBitmap((int) width, (int) height);
+            if (bitmap == null) {
+                return;
+            }
         }
 
         float imageWidth = (float) bitmap.getWidth();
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/waz/zclient/pages/main/backgroundmain/views/BackgroundDrawable.java
```diff
@@ -151,6 +149,18 @@ private void createBackgroundShader() {
...
 
+    private Bitmap createBitmap(int width, int height) {
+        Bitmap ret;
+        try {
+            int[] colors = new int[width * height];
+            Arrays.fill(colors, DEFAULT_BACKGROUND_COLOR);
+            ret = Bitmap.createBitmap(colors, width, height, Bitmap.Config.ARGB_8888);
+            return ret;
+        } catch (OutOfMemoryError e){
+            return createBitmap(width / 2, height / 2);
```
makingthematrix: I think we should stop at some smallest practical width and height and if we still get OutOfMemory error, just throw it then. Otherwise we could end up in a situation where a real OOM will be covered with a stack overflow, as Wire will constantly try to create a bitmap of width ==0 and height == 0, but it will still need a few bytes of memory, so another OOM will be raised.

Please create constants `SMALLEST_WIDTH` and `SMALLEST_HEIGHT`. They can be hardcoded to 1. And please check them in the catch block before making a new recursive call to `createBitmap`.

JPonte: It actually generates a IlegalArgumentException, made a check for that instead.

PR #770, opened 2017-04-05
Set send assets v3 preference true by default

JPonte:
#### APK
[Download build #8740](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8740/artifact/build/artifact/wire-dev-PR770-8740.apk)

### DIFF
--- a/app/src/main/res/xml/preferences_developer.xml
+++ b/app/src/main/res/xml/preferences_developer.xml
@@ -47,7 +47,7 @@
         />
 
     <CheckBoxPreference
-        android:defaultValue="false"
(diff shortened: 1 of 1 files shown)


app/src/main/res/xml/preferences_developer.xml
```diff
@@ -47,7 +47,7 @@
         />
 
     <CheckBoxPreference
-        android:defaultValue="false"
```
hakonbo: You'll need to change the default value where it's read in code. Ping me if you need more details.

PR #805, opened 2017-04-20
Removing spotify in-app player

hakonbo:
#### APK
[Download build #8771](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8771/artifact/build/artifact/wire-dev-PR805-8771.apk)


hakonbo:
build again

hakonbo:
build again

levinus:
Why is the inapp player being removed?

hakonbo:
@levinus it's not been there for a while, the circular player was not done when we re-implemented the conversation content in Scala a while back (end of last year), mostly due to time constraint Spotify and Soundcloud tracks are displayed similar to link previews. We may or may not add a new in-app player, but for now it's a bit down on the priority list. It seems like people prefer to play content in the native apps instead of Wire.

bgermann:
Will the spotify dependencies also be removed in wire-android-sync-engine?

PR #808, opened 2017-04-21
Fixed NPE when opening app from quick reply

hakonbo:
#### APK
[Download build #8778](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8778/artifact/build/artifact/wire-dev-PR808-8778.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/utils/IntentUtils.java
+++ b/app/src/main/java/com/waz/zclient/utils/IntentUtils.java
@@ -321,7 +321,12 @@ public class IntentUtils {
     }
 
     private static List<Uri> getParcelableArrayListExtra(Intent intent) {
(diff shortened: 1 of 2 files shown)


app/src/main/java/com/waz/zclient/utils/IntentUtils.java
```diff
@@ -321,7 +321,12 @@ public static String getLaunchConversationSharedText(Intent intent) {
     }
 
     private static List<Uri> getParcelableArrayListExtra(Intent intent) {
-        return intent.getParcelableArrayListExtra(EXTRA_LAUNCH_CONVERSATION_FILES);
+        List<Uri> files = intent.getParcelableArrayListExtra(EXTRA_LAUNCH_CONVERSATION_FILES);
```
mykola-mokhnach: return (files == null) ? Collections.emptyList() : files;

PR #821, opened 2017-04-27
No internet bar, use new Unknown mode as initial value

hakonbo:
#### APK
[Download build #8799](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8799/artifact/build/artifact/wire-dev-PR821-8799.apk)


hakonbo:
build again

PR #835, opened 2017-05-03
Fix the logic to show "You are logged in for the first time" fragment

mykola-mokhnach:
This PR fixes a bug when First time login fragment is shown if the same user logs out and logs in again on the same device and "You are logged in for the first time" dialog is still shown for him.

#### APK
[Download build #8817](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8817/artifact/build/artifact/wire-dev-PR835-8817.apk)

### DIFF
--- a/wire-core/src/main/java/com/waz/zclient/core/api/scala/AppEntryStore.java
+++ b/wire-core/src/main/java/com/waz/zclient/core/api/scala/AppEntryStore.java
@@ -107,7 +107,7 @@ public class AppEntryStore implements IAppEntryStore, ErrorsList.ErrorListener {
     private Invitations.PersonalToken invitationToken;
     private RegistrationEventContext registrationEventContext;
 
-    private boolean isFirstSignInOnDevice = false;
+    private boolean isFirstSignInOnDevice = true;
 
     private UpdateListener selfUpdateListener = new UpdateListener() {
         @Override
@@ -522,7 +522,11 @@ public class AppEntryStore implements IAppEntryStore, ErrorsList.ErrorListener {
                 setRegistrationContext(RegistrationEventContext.PERSONAL_INVITE_PHONE);
                 break;
             case FIRST_LOGIN:
-                appEntryStateCallback.onShowFirstLaunchPage();
+                if (isFirstSignInOnDevice) {
+                    appEntryStateCallback.onShowFirstLaunchPage();
+                } else {
+                    setState(AppEntryState.LOGGED_IN);
+                }
                 break;
         }
     }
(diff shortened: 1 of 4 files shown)


wire-core/src/main/java/com/waz/zclient/core/api/scala/AppEntryStore.java
```diff
@@ -510,7 +522,11 @@ private void setStateInternal(AppEntryState state, boolean forceUpdate) {
                 setRegistrationContext(RegistrationEventContext.PERSONAL_INVITE_PHONE);
                 break;
             case FIRST_LOGIN:
-                appEntryStateCallback.onShowFirstLaunchPage();
+                if (isFirstSignInOnDevice) {
```
JPonte: I think that if it's not the first login it shouldn't reach this state in the first place, so this check should be on the place where the `setState(AppEntryState .FIRST_LOGIN)` is.

mykola-mokhnach: I just didn't want to expose **isFirstSignInOnDevice** property. Also, state change has been already implemented like that if you check the previous implementation of _onShowFirstLaunchPage_

JPonte: What do you mean expose it? The `FIRST_LOGIN` state is being set on `EMAIL_SIGNED_IN` and `PHONE_SIGNED_IN_RESUMING` on that switch. It doesn't seem right to switch to the `FIRST_LOGIN` state if it's not in fact the first login.

mykola-mokhnach: not only. Check the EmailAddPhoneFragment->onNotNow method

JPonte: Right, makes sense then. I think it's a bit ugly to expose the setState outside that class but ok.

wire-core/src/main/java/com/waz/zclient/core/api/scala/AppEntryStore.java
```diff
@@ -605,6 +621,14 @@ public void onConfirmationCodeSendingFailed(KindOfAccess kindOfAccess,
 
     @Override
     public void setSignInPhone(final String countryCode, final String phone, final ErrorCallback errorCallback) {
+        try {
+            AccountsStorage storage = getAccountsStorage();
+            isFirstSignInOnDevice = storage == null || !Await.result(storage.findByPhone(phone),
+                Duration.create(1, TimeUnit.SECONDS)).isDefined();
+        } catch (Exception e) {
```
wire-android: What exception do you actually expect here? Catching unspecified exception is weird, it kind of says: "I have no idea what I'm doing, will catch all" :)
Besides that it seems that you are using exception to pass some logic, it's not recommended, would be better to check before if there will be exception.

wire-core/src/main/java/com/waz/zclient/core/api/scala/AppEntryStore.java
```diff
@@ -1045,8 +1069,23 @@ public void onRegistrationFailed(int i, String s, String s1) {
...
+        return global.accountsStorage();
+    }
+
     @Override
     public void signInWithEmail(String email, String password, final ErrorCallback errorCallback) {
+        try {
+            AccountsStorage storage = getAccountsStorage();
+            isFirstSignInOnDevice = storage == null ||
+                !Await.result(storage.findByEmail(email), Duration.create(1, TimeUnit.SECONDS)).isDefined();
+        } catch (Exception e) {
```
wire-android: again, catching anything

wire-core/src/main/java/com/waz/zclient/core/api/scala/AppEntryStore.java
mykola-mokhnach: This is the definition of Await.result:

```java
    public static <T> T result(Awaitable<T> var0, Duration var1) throws Exception {
}
```

can it be implemented differently?

wire-android: oh, missed that;
waiting for db results on UI thread is even bigger issue :/ 
guess there is no easy way of implementing this with Await

hakonbo:
Closing for now, please re-open when ready for merging

PR #843, opened 2017-05-06
Mark asset message as read only when asset is available.

wire-android:
Added check for asset state when reporting messages as read, this is needed to make sure ephemeral msg counter is not started if asset is not available.

Asset availability here is different for different assets.
For regular image assets we wait until image is downloaded (and displayed) as this is started automatically.
For other assets we only wait until upload is done (or failed), we don't want to wait until download as download is user triggered, and we want messages to expire even if user didn't request to download.

#### APK
[Download build #8831](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8831/artifact/build/artifact/wire-dev-PR843-8831.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/messages/MessagesListView.scala
+++ b/app/src/main/scala/com/waz/zclient/messages/MessagesListView.scala
@@ -27,10 +27,12 @@ import android.util.AttributeSet
 import android.view.WindowManager
 import com.waz.ZLog.ImplicitTag._
 import com.waz.ZLog._
+import com.waz.api.{AssetStatus, Message}
 import com.waz.model.{ConvId, Dim2, MessageData}
 import com.waz.service.messages.MessageAndLikes
 import com.waz.threading.Threading
 import com.waz.utils.events.{EventContext, Signal}
+import com.waz.zclient.controllers.AssetsController
 import com.waz.zclient.controllers.global.SelectionController
 import com.waz.zclient.messages.MessageView.MsgBindOptions
 import com.waz.zclient.messages.ScrollController.Scroll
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/messages/MessagesListView.scala
```diff
@@ -171,13 +174,28 @@ case class MessageViewHolder(view: MessageView, adapter: MessagesListAdapter)(im
...
+      case Some(convId) =>
+        message.filter(_.convId == convId) flatMap {
+          case msg if msg.isAssetMessage && msg.state == Message.Status.SENT =>
+            // received asset message is considered read when its asset is available,
+            // this is especially needed for ephemeral messages, only start the counter when message is downloaded
+            assets.assetSignal(msg.assetId) flatMap {
+              case (_, AssetStatus.DOWNLOAD_DONE) if msg.msgType == Message.Type.ASSET =>
+                // image assets are considered read only once fully downloaded
+                Signal const msg
+              case (_, AssetStatus.UPLOAD_DONE | AssetStatus.UPLOAD_CANCELLED | AssetStatus.UPLOAD_FAILED) if msg.msgType != Message.Type.ASSET =>
```
deanrobertcook: I just checked with how iOS does it, and they wait until the download is done before they start the ephemeral countdown, so I guess we should remain consistent with what they do.

Also (and I guess we don't do this) they make sure the player doesn't disappear if the user is still playing the asset, even after it has expired.

deanrobertcook: I just checked with how iOS does it, and they wait until the download is done before they start the ephemeral countdown, so I guess we should remain consistent with what they do.

Also (and I guess we don't do this) they make sure the player doesn't disappear if the user is still playing the asset, even after it has expired.

deanrobertcook: never mind!

PR #862, opened 2017-05-12
Include an old hack for the name updater that was in the Ui Conversation

JPonte:
#### APK
[Download build #8867](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8867/artifact/build/artifact/wire-dev-PR862-8867.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/views/conversationlist/ConversationListRow.scala
+++ b/app/src/main/scala/com/waz/zclient/views/conversationlist/ConversationListRow.scala
@@ -97,6 +97,11 @@ class NormalConversationListRow(context: Context, attrs: AttributeSet, style: In
     if (conv.convType == ConversationType.Incoming) {
       (conv.id, getInboxName(memberCount))
     } else {
+      if (conv.displayName == "") {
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/views/conversationlist/ConversationListRow.scala
```diff
@@ -97,6 +97,11 @@ class NormalConversationListRow(context: Context, attrs: AttributeSet, style: In
     if (conv.convType == ConversationType.Incoming) {
       (conv.id, getInboxName(memberCount))
     } else {
+      if (conv.displayName == "") {
+        // This hack was in the UiModule Conversation implementation
+        // XXX: this is a hack for some random errors, sometimes conv has empty name which is never updated
+        zms.head.map {_.conversations.forceNameUpdate(conv.id)}
+      }
```
deanrobertcook: We should probably fix the naming issues in the `ConversationData` object itself. If this is for (another) hotfix, then cool, but we should make a ticket and sort this out. I think it gets us in a few places.

PR #867, opened 2017-05-17
Various bug fixes from crash reporting

hakonbo:
[AN-5239] Fix possible NoSuchMethodError
[AN-5243] Fixed possible NPE
[AN-5238] Fix possible IndexOutOfBoundsException
[AN-5244] Fix possible NPE
[AN-5249] Fixed possible NPE in ShareActivity
[AN-5250] Fixed possible NPE in ConversationFragment
[AN-5251] Fixed possible NPE in DeviceDetailPreferences
[AN-5252] Fixed possible NPE in QuickReplyFragment
#### APK
[Download build #8873](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8873/artifact/build/artifact/wire-dev-PR867-8873.apk)
[Download build #8875](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8875/artifact/build/artifact/wire-dev-PR867-8875.apk)
[Download build #8878](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8878/artifact/build/artifact/wire-dev-PR867-8878.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/adapters/ConversationListAdapter.scala
+++ b/app/src/main/scala/com/waz/zclient/adapters/ConversationListAdapter.scala
@@ -40,7 +40,7 @@ class ConversationListAdapter(context: Context)(implicit injector: Injector, eve
 
   lazy val zms = inject[Signal[ZMessaging]]
   val currentMode = Signal[ListMode]()
-  lazy val conversations = for {
+  lazy val conversationsSignal = for {
     z <- zms
     conversations <- z.convsStorage.convsSignal
     mode <- currentMode
@@ -49,7 +49,10 @@ class ConversationListAdapter(context: Context)(implicit injector: Injector, eve
       .filter(mode.filter).toSeq
       .sorted(mode.sort)
 
-  lazy val incomingRequests = for {
+  var conversations: Option[Seq[ConversationData]] = None
+  var incomingRequests: Option[(Seq[ConversationData], Seq[UserId])] = None
+
+  lazy val incomingRequestsSignal = for {
     z <- zms
     conversations <- z.convsStorage.convsSignal.map(_.conversations.filter(Incoming.filter).toSeq)
     members <- Signal.sequence(conversations.map(c => z.membersStorage.activeMembers(c.id).map(_.find(_ != z.selfUserId))):_*)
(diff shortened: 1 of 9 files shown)


app/src/main/scala/com/waz/zclient/adapters/ConversationListAdapter.scala
```diff
@@ -49,7 +49,10 @@ class ConversationListAdapter(context: Context)(implicit injector: Injector, eve
       .filter(mode.filter).toSeq
       .sorted(mode.sort)
 
-  lazy val incomingRequests = for {
+  var conversations: Option[Seq[ConversationData]] = None
+  var incomingRequests: Option[(Seq[ConversationData], Seq[UserId])] = None
```
deanrobertcook: I don't think the `Option` is necessary here. `Seq` and `Options` can both be thought of as containers for things (just that `Option` is limited to size 1), so having a container of a container just adds a bit more complexity. There are plenty of methods on `Seq` that allow you to check if it's empty etc.

Just as a side note, there are some cases where an Option of another collection does make sense: usually whenever an empty collection is semantically different from having no data at all. The most recent example I can think of is user permissions (a `Set`) for teams. If the user is not part of a team, then having a `Option[Set[Permission]]` set to `None` signifies he's not in a team, but having a `Some(Set.empty[Permission]]` would mean he's on the team, but doesn't have any permissions. 

In this case however, if we have no data, that's pretty much it. It can't really be achieved in two ways.

app/src/main/scala/com/waz/zclient/adapters/ConversationListAdapter.scala
```diff
@@ -65,19 +65,17 @@ class ConversationListAdapter(context: Context)(implicit injector: Injector, eve
   var maxAlpha = 1.0f
 
   Signal(conversationsSignal, incomingRequestsSignal).on(Threading.Ui) { data =>
```
deanrobertcook: ever so slightly nicer is also:

```
... { 
  case (convs, requests) => 
    conversations = convs
    incomingRequests = requests
    notifyDataSetChanged()
}
```

app/src/main/scala/com/waz/zclient/adapters/ConversationListAdapter.scala
```diff
@@ -65,19 +65,17 @@ class ConversationListAdapter(context: Context)(implicit injector: Injector, eve
...
   private def getConversation(position: Int): Option[ConversationData] = {
-    conversations.fold(Option.empty[ConversationData]){ convs =>
-      convs.lift(position)
-    }
+    conversations.lift(position)
   }
 
   private def getItem(position: Int): Option[ConversationData] = {
-    if (incomingRequests.exists(_._2.nonEmpty)) {
+    if (incomingRequests._2.nonEmpty) {
```
deanrobertcook: I think we can get rid of every curly bracket in this method :) Slightly more in Scala style

Also, I think a case might look slightly nicer?

```
incomingRequests._2 match {
  case Seq.empty => getConversation(position)
  case _ => if (position == 0) None else getConversation(position -1)
}
```

app/src/main/scala/com/waz/zclient/adapters/ConversationListAdapter.scala
```diff
@@ -90,11 +88,11 @@ class ConversationListAdapter(context: Context)(implicit injector: Injector, eve
 
   override def getItemCount = {
     val incoming =
-      if (incomingRequests.exists(_._2.nonEmpty))
+      if (incomingRequests._2.nonEmpty)
```
deanrobertcook: I think it's nicer to one-line these small `if-else`s as well: `if (empty) 1 else 0`

app/src/main/scala/com/waz/zclient/adapters/ConversationListAdapter.scala
```diff
@@ -156,7 +154,7 @@ class ConversationListAdapter(context: Context)(implicit injector: Injector, eve
   }
 
   override def getItemViewType(position: Int): Int = {
-    if (position == 0 && incomingRequests.exists(_._2.nonEmpty)) {
+    if (position == 0 && incomingRequests._2.nonEmpty) {
```
deanrobertcook: brackets again

hakonbo:
build again

PR #889, opened 2017-05-26
[AN-5146] Ask to share contacts when opening personal start ui

hakonbo:
Always ask use for permission to sync hashed contact details for user matching. 
For Android  4.x and 5.x we already have permission to access contact details, but we want to explicitly ask user for this. Once user opens start UI a dialog is shown asking for permission to access contacts. For Android 4.x and 5.x we use our own dialog, for Android 6.x and up, we just trigger the default permission dialog.
The dialog will be shown on every start UI launch, until user either says yes to sharing, or ticks the "do not show again" checkbox. 
For existing users that haven't touched the share contacts setting, we will show this dialog. For existing users who have set the share contacts preference to true, we will not show the dialog.
Please see ticket for more details and reasoning for this behaviour. 
#### APK
[Download build #8923](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8923/artifact/build/artifact/wire-dev-PR889-8923.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/controllers/userpreferences/UserPreferencesController.java
+++ b/app/src/main/java/com/waz/zclient/controllers/userpreferences/UserPreferencesController.java
@@ -20,10 +20,8 @@ package com.waz.zclient.controllers.userpreferences;
 import android.annotation.SuppressLint;
 import android.content.Context;
 import android.content.SharedPreferences;
-
 import com.waz.zclient.R;
 import com.waz.zclient.utils.StringUtils;
-
 import org.json.JSONArray;
 import org.json.JSONException;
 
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/waz/zclient/controllers/userpreferences/UserPreferencesController.java
```diff
@@ -290,7 +288,7 @@ public boolean hasCheckedForUnsupportedEmojis(int version) {
 
     @Override
     public boolean hasShareContactsEnabled() {
-        return userPreferences.getBoolean(context.getString(R.string.pref_share_contacts_key), true);
+        return userPreferences.getBoolean(context.getString(R.string.pref_share_contacts_key), false);
```
red-avtovo: Should we really disable that by default?

hakonbo: Yes

red-avtovo:
👍

PR #901, opened 2017-06-02
Remove V2 Calling code

deanrobertcook:
#### APK
[Download build #8959](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8959/artifact/build/artifact/wire-dev-PR901-8959.apk)


deanrobertcook:
@hakonbo I imagine this would be an existing bug... at least removing V2 shouldn't have had an effect on that...

PR #936, opened 2017-06-14
Move preference dialogs to Scala: part 1

deanrobertcook:
This moves the `ChangeUsernamePreferenceFragment` to scala (and now also renamed to `ChangeHandleFragment`. 

As part of this, it was necessary to also move the `AccountPreferences` fragment to scala, which has the nice side effect of removing a lot of the usage from the `Self` java API.

All preferences and settings should work as they did before, and now they even update in real time if the values are changed on another client.

#### APK
[Download build #9026](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9026/artifact/build/artifact/wire-dev-PR936-9026.apk)
[Download build #9027](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9027/artifact/build/artifact/wire-dev-PR936-9027.apk)
[Download build #9046](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9046/artifact/build/artifact/wire-dev-PR936-9046.apk)
[Download build #9047](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9047/artifact/build/artifact/wire-dev-PR936-9047.apk)


deanrobertcook:
will get QA review after all preferences are done

app/src/main/scala/com/waz/zclient/preferences/AccountPreferences.scala
```diff
@@ -0,0 +1,358 @@
...
+        case Some(h) =>
+          p.setTitle(StringUtils.formatHandle(h.string))
+          p.setSummary(getString(R.string.pref_account_username_title))
+        case None =>
+          p.setTitle(getString(R.string.pref_account_username_empty_title))
+          p.setSummary("")
+      }
+    },
+
+    //Reset password
```
makingthematrix: Since this code repeats itself so much, how about a utility method:
```
  private def onClick(keyId: Int)(body: => Boolean) = {
    val p = findPref[Preference](keyId)
    p.setOnPreferenceClickListener(new OnPreferenceClickListener() {
      override def onPreferenceClick(preference: Preference): Boolean = body
    })
    p
  }
```

and then:
```
onClick(R.string.pref_account_password_key){
      tracking.tagEvent(new ResetPassword(ResetPassword.Location.FROM_PROFILE))
      false
}
```

You can even use it inside `returning(...)` for the username.

app/src/main/scala/com/waz/zclient/preferences/AccountPreferences.scala
```diff
@@ -0,0 +1,358 @@
...
+            .add(new AccentColorPickerFragment, AccentColorPickerFragment.fragmentTag)
+            .addToBackStack(AccentColorPickerFragment.fragmentTag)
+            .commit
+          true
+        }
+      })
+      selfUser.map(_.accent).map(AccentColor(_).getColor()).on(Threading.Ui)(p.setAccentColor)
+    },
+
+    phonePref,
```
makingthematrix: And it would be even more readable if there was a `lazy val blahblahPref` for each of the above and only a seq of them here.

deanrobertcook:
rebasing to master

PR #957, opened 2017-06-23
[AN-5069] fixes choosing the best fit picture size supported by the camera

makingthematrix:
Just before taking a photo we specify the picture size, choosing one from the list of ones supported by the camera. So far we assumed the first one on the list is the biggest and we choose that. It is not true for some phones. Besides, we can be smarter and choose not the biggest picture size, but the one that fits the best our default size for photos sent over Wire.
#### APK
[Download build #9077](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9077/artifact/build/artifact/wire-dev-PR957-9077.apk)
[Download build #9080](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9080/artifact/build/artifact/wire-dev-PR957-9080.apk)


makingthematrix:
build again

PR #1042, opened 2017-07-25
[AN-5406] User search functionality moved to SE

makingthematrix:
Exact match moved to the top of results
*Requires new SE*
https://github.com/wireapp/wire-android-sync-engine/pull/200
#### APK
[Download build #9279](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9279/artifact/build/artifact/wire-dev-PR1042-9279.apk)
[Download build #9280](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9280/artifact/build/artifact/wire-dev-PR1042-9280.apk)


app/src/main/scala/com/waz/zclient/controllers/SearchUserController.scala
```diff
@@ -20,44 +20,39 @@ package com.waz.zclient.controllers
...
-}
+import scala.concurrent.duration._
+import scala.language.postfixOps
 
 class SearchUserController(initialState: SearchState)(implicit injector: Injector, ec: EventContext) extends Injectable {
   implicit private val uiStorage = inject[UiStorage]
 
   private val zms = inject[Signal[ZMessaging]]
 
   val searchState = Signal(initialState)
```
deanrobertcook: Do we use the `searchState` anywhere else? If we always want to ensure that it's throttled, maybe just inline it's definition with the throttled version.

red-avtovo:
Searching for user takes for half a minute on Staging

PR #1071, opened 2017-08-09
[AN-5494] fixes initial default ringtone being silent

makingthematrix:
If the ringtone is not set explicitly, the `SoundController` receives an empty uri. This should be replaced by the default ringtone when registering `ringing_from_them` sound, but it wasn't. Fixed.

https://wearezeta.atlassian.net/browse/AN-5494

Also, I splitted the `SoundController.setCustomSoundUrisFromPreferences` into two to make it more readable.
#### APK
[Download build #9327](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9327/artifact/build/artifact/wire-dev-PR1071-9327.apk)


someoneEsle:
looks good to me

PR #1072, opened 2017-08-09
[AN-5523] Update SE and AVS for new multiple-accounts API

deanrobertcook:
https://wearezeta.atlassian.net/browse/AN-5523
#### APK
[Download build #9328](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9328/artifact/build/artifact/wire-dev-PR1072-9328.apk)


yahaverlich:
The bugs reported were UI notification based, which was not the purpose of this build, my apologies. The only calling bug that seems prominent, is when the calls get stuck on joining. Otherwise everything works properly.

PR #1150, opened 2017-09-28
[AN-5680] Use finish instead of finishAfterTransition on sdk below 21

JPonte:
#### APK
[Download build #9527](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9527/artifact/build/artifact/wire-dev-PR1150-9527.apk)


someoneEsle:
We are currently investigating this issue -- will update once we have figured out exactly what was happening

someoneEsle:
ok looks good

PR #1283, opened 2017-11-22
[AN-5711] Keyboard flickering fix

makingthematrix:
`KeyboardVisibilityListener.onLayoutChange` could be called in a loop sometimes:
1. It's called during init.
2. It computes the dimensions of the screen for the keyboard and its visibility.
3. It calls listeners for these values.
4. The listeners trigger displaying or hiding the keyboard.
5. That triggers `onLayoutChange` and recomputing of the screen dimensions.
Eventually the loop stops if two consecutive calls give the same outcome, but depending on the device it may take a while (initialization of other layout elements gets in the way).

I delayed the computations and synchronized it. Now the recomputation should happen at most once (ie. changes triggered by the initial call to `onLayoutChange` should result in at most one more call to it).

Now, if the user switches to another conversation and then back to Wire, the keyboard hides, but does not flicker. In the logs I see that there are other methods which hide the keyboard when we switch back and forth, so remembering the state and reopening the keyboard is much more tricky than just fixing the flickering.
#### APK
[Download build #9825](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9825/artifact/build/artifact/wire-dev-PR1283-9825.apk)
[Download build #9826](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9826/artifact/build/artifact/wire-dev-PR1283-9826.apk)
[Download build #9833](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9833/artifact/build/artifact/wire-dev-PR1283-9833.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/utils/keyboard/KeyboardVisibilityListener.java
+++ b/app/src/main/java/com/waz/zclient/utils/keyboard/KeyboardVisibilityListener.java
@@ -25,7 +25,6 @@ import com.waz.zclient.utils.ContextUtils;
 public class KeyboardVisibilityListener {
     private final View contentView;
     private final int statusAndNavigationBarHeight;
-    private int lastKeyboardHeight;
     private int keyboardHeight;
 
     public interface Callback {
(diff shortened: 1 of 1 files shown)


someoneEsle:
The loop is fixed, but the behaviour is still not optimal (you can see the keyboard close, and sometimes it looks like a weird flash). I'll show you tomorrow

PR #1326, opened 2017-12-12
Re-enable missed notification tracking for internal

CalumMcCall:
#### APK
[Download build #9950](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9950/artifact/build/artifact/wire-dev-PR1326-9950.apk)
[Download build #9957](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9957/artifact/build/artifact/wire-dev-PR1326-9957.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
+++ b/app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
@@ -121,8 +121,8 @@ class GlobalTrackingController(implicit inj: Injector, cxt: WireContext, eventCo
       }
 
       event match {
-        case _: MissedPushEvent =>
-        //TODO - re-enable this event when we can reduce their frequency a little. Too many events for mixpanel right now
+        case _: MissedPushEvent if !BuildConfig.FLAVOR.equals("internal") =>
+          //This event is high volume, so we limit it to only internal clients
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
```diff
@@ -121,8 +121,8 @@ class GlobalTrackingController(implicit inj: Injector, cxt: WireContext, eventCo
       }
 
       event match {
-        case _: MissedPushEvent =>
-        //TODO - re-enable this event when we can reduce their frequency a little. Too many events for mixpanel right now
+        case _: MissedPushEvent if BuildConfig.DEBUG =>
```
deanrobertcook: Pretty sure internal has `BuildConfig.DEBUG` set to false, since we build with `assembleInternalRelease` (debug builds can't be uploaded to the play store). 

It might be better to use the `BuildConfig.FLAVOR` field instead

deanrobertcook: Also, this will send the event even if tracking is disabled, which is probably not a big deal for internal users, but maybe you should invert the check and do nothing just before the final `case` (i.e., let the event you want to re-enable fall into the final `case`)

PR #1334, opened 2017-12-18
[AN-5846] fix keyboard not closing

makingthematrix:
I moved `notifyKeyboardVisibilityChanged` from `CursorView` to `CursorController` and call it `ConversationPagerFragment` when the page is switch to the conversation list. The name `notifyKeyboardVisibilityChanged` is misleading right now - it shows/hides the keyboard, not just reacts to the keyboard being already shown/hidden. I focused only on fixing the bug, so I left it as it is. I'd like to come back to it later on.

Also, in the `DrawingFragment`, the text bubble wasn't appearing correctly because it relied on the old listener. I fixed that, but again, I didn't want to remove the listener completely, because I didn't want to risk new bugs. 
#### APK
[Download build #9983](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9983/artifact/build/artifact/wire-dev-PR1334-9983.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/pages/main/conversationpager/ConversationPagerFragment.java
+++ b/app/src/main/java/com/waz/zclient/pages/main/conversationpager/ConversationPagerFragment.java
@@ -35,8 +35,8 @@ import com.waz.zclient.controllers.navigation.Page;
 import com.waz.zclient.controllers.navigation.PagerControllerObserver;
 import com.waz.zclient.conversation.ConversationController;
 import com.waz.zclient.collection.controllers.CollectionController;
+import com.waz.zclient.cursor.CursorController;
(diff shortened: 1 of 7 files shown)


someoneEsle:
Bug 1
The keyboard should close as soon as I start swiping to the right, and not wait to close until the conversation has been swiped completely

makingthematrix:
@someoneEsle : 
~"The keyboard should close as soon as I start swiping to the right, and not wait to close until the conversation has been swiped completely"
I created https://wearezeta.atlassian.net/browse/AN-5849 for that.

someoneEsle:
Let's try to remember to fix this in the next release. It's not critical of course, but it shouldn't be forgotten :)

PR #1453, opened 2018-03-07
Fix incorrect comparison of non-option call state

deanrobertcook:
#### APK
[Download build #10461](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10461/artifact/build/artifact/wire-dev-PR1453-10461.apk)


someoneEsle:
Looks good after a quick manual test, I'll be waiting for the result of automation on experimental (which points to this branch) before removing the QA label

PR #1478, opened 2018-03-16
Fix like and unlike in the single image toolbar

JPonte:
#### APK
[Download build #10578](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10578/artifact/build/artifact/wire-dev-PR1478-10578.apk)
[Download build #10591](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10591/artifact/build/artifact/wire-dev-PR1478-10591.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/collection/views/SingleImageViewToolbar.scala
+++ b/app/src/main/scala/com/waz/zclient/collection/views/SingleImageViewToolbar.scala
@@ -32,6 +32,7 @@ import com.waz.zclient.utils.ContextUtils._
 import com.waz.zclient.utils.RichView
 import com.waz.zclient.{R, ViewHelper}
 import com.waz.ZLog.ImplicitTag._
+import MessageAction._
 
 class SingleImageViewToolbar(context: Context, attrs: AttributeSet, style: Int) extends LinearLayout(context, attrs, style) with ViewHelper {
   def this(context: Context, attrs: AttributeSet) = this(context, attrs, 0)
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/collection/views/SingleImageViewToolbar.scala
```diff
@@ -77,7 +77,12 @@ class SingleImageViewToolbar(context: Context, attrs: AttributeSet, style: Int)
   Seq(likeButton, downloadButton, shareButton, deleteButton, viewButton)
     .foreach(_.setPressedBackgroundColor(getColor(R.color.light_graphite)))
 
-  likeButton.onClick( message.head.foreach(msg => messageActionsController.onMessageAction ! (MessageAction.Like, msg)))
+  likeButton.onClick( Signal(message, likedBySelf).head.foreach{
```
deanrobertcook: be careful of using the `ZipSignal` too much - the way it works is that it always has a subscription to all of the signals you pass it, which in a lot of cases is fine, but it might lead to unnecessary expensive calculations in the worst case. Generally a `FlatmapSignal` is safer, which is why I prefer using `for` comprehensions

app/src/main/scala/com/waz/zclient/collection/views/SingleImageViewToolbar.scala
```diff
@@ -77,7 +77,12 @@ class SingleImageViewToolbar(context: Context, attrs: AttributeSet, style: Int)
...
+    case (msg, true) =>
+      messageActionsController.onMessageAction ! (MessageAction.Unlike, msg)
+    case (msg, false) =>
+      messageActionsController.onMessageAction ! (MessageAction.Like, msg)
```
deanrobertcook:
```
import MessageAction._
messageActionsController.onMessageAction ! (if (likedBySelf) Unlike else Like, msg)
```

PR #1491, opened 2018-03-21
Remove giphy java api

yeryomenkom:
#### APK
[Download build #10611](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10611/artifact/build/artifact/wire-dev-PR1491-10611.apk)
[Download build #10613](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10613/artifact/build/artifact/wire-dev-PR1491-10613.apk)
[Download build #10620](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10620/artifact/build/artifact/wire-dev-PR1491-10620.apk)
[Download build #10622](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10622/artifact/build/artifact/wire-dev-PR1491-10622.apk)
[Download build #10632](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10632/artifact/build/artifact/wire-dev-PR1491-10632.apk)
[Download build #10633](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10633/artifact/build/artifact/wire-dev-PR1491-10633.apk)
[Download build #10651](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10651/artifact/build/artifact/wire-dev-PR1491-10651.apk)


app/src/main/scala/com/waz/zclient/giphy/GiphySharingPreviewFragment.scala
```diff
@@ -0,0 +1,293 @@
...
+
+object GiphySharingPreviewFragment {
+
+  //TODO Move this to Signal class
+  implicit class RichSignal[T](signal: Signal[T]) {
+    def either[B](right: Signal[B]): Signal[Either[T, B]] =
+      signal.map(Left(_): Either[T, B]).orElse(right.map(Right.apply))
+
+    def pipeTo(sourceSignal: SourceSignal[T])(implicit ec: EventContext): Unit =
+      signal.foreach(sourceSignal ! _)
```
deanrobertcook: Nice ones!

app/src/main/scala/com/waz/zclient/giphy/GiphySharingPreviewFragment.scala
```diff
@@ -0,0 +1,293 @@
...
+    giphySearchEditText
+    recyclerView
+    errorView
+    confirmationMenu
+    toolbar
+
+    searchTerm.onUi(_ => spinnerController.showSpinner(LoadingIndicatorView.InfiniteLoadingBar))
+    giphySearchResults.onUi(_ => spinnerController.hideSpinner())
+    selectedGif.filter(_.isDefined).onUi(_ => spinnerController.showSpinner(LoadingIndicatorView.InfiniteLoadingBar))
+    isPreviewShown.filter(_ == false).onUi(_ => spinnerController.hideSpinner())
```
deanrobertcook: So, be careful not to treat signals like events, here we have two states we want to achieve, either hiding or showing the loading bar. In general try to avoid using `Empty` signals if possible. They should really just be used for undefined state (or in some cases to simplify downstream logic). In this case, the search term is a string, and can be the empty string `""` or some input, so I would include this in your subscription:

```
val searchTerm = Signal("")
val selectedGif = Signal(Option.empty[AssetData])

...

(for {
      st   <- searchTerm
      res  <- giphySearchResults
      sel  <- selectedGif
      prev <- isPreviewShown
    } yield st.nonEmpty && res.isEmpty && sel.isDefined && prev).onUi {
      case true => spinnerController.showSpinner(LoadingIndicatorView.InfiniteLoadingBar)
      case _    => spinnerController.hideSpinner()
    }

} 
```

for example. (Note, I didn't think very hard about the logic in my version - it's probably wrong)

app/src/main/scala/com/waz/zclient/giphy/GiphySharingPreviewFragment.scala
```diff
@@ -0,0 +1,272 @@
...
+        }
+      })
+    }
+
+    val gifDrawable = new ImageAssetDrawable(gifImage, scaleType = ScaleType.CenterInside)
+    previewImage.setImageDrawable(gifDrawable)
+
+    EventStream.union(
+      searchTerm.onChanged.map(_ => true),
+      selectedGif.filter(_.isDefined).onChanged.map(_ => true),
```
deanrobertcook: you probably want to call `onChanged` first? If you filter first, then this event stream will only fire once when the signal goes from `None` -> `Some(x)`. Maybe that's what you want, but I'm just checking

app/src/main/scala/com/waz/zclient/giphy/GiphySharingPreviewFragment.scala
```diff
@@ -0,0 +1,272 @@
...
+      case true =>
+        if (themeController.isDarkTheme) R.drawable.action_back_light
+        else R.drawable.action_back_dark
+      case false =>
+        if (themeController.isDarkTheme) R.drawable.ic_action_search_light
+        else R.drawable.ic_action_search_dark
+    }.onUi(icon => vh.foreach(_.setNavigationIcon(icon)))
+  }
+  private lazy val giphySearchEditText = returning(view[EditText](R.id.cet__giphy_preview__search)) { vh =>
+    vh.foreach { _.afterTextChangedSignal() pipeTo searchTerm }
```
deanrobertcook: just noticed - this needs to be done in `onViewCreated` as well!

someoneEsle:
- [ ] Bug 1
1. Open a conversation
2. Type something (without sending)
3. Tap GIF button
4. Now we should display search results for the text you entered, but we show the same results as if the search field were empty

PR #1526, opened 2018-04-06
Handle null pointer returned by File.listFiles in WireApplication.clearOldVideoFiles

makingthematrix:
`File.listFiles` returns `null` if it's not possible to read from the directory or the path is invalid for some reason. It's a very rare case of a very ugly Java way of doing things. I wrapped the result in an `Option`.


makingthematrix:
build again

PR #1546, opened 2018-04-20
Refactor conversation options bottom sheet dialog 

JPonte:
And use native dialogs for confirmation

https://github.com/wireapp/android-project/issues/38

#### APK
[Download build #10797](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10797/artifact/build/artifact/wire-dev-PR1546-10797.apk)
[Download build #10799](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10799/artifact/build/artifact/wire-dev-PR1546-10799.apk)
[Download build #10808](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10808/artifact/build/artifact/wire-dev-PR1546-10808.apk)


app/src/main/scala/com/waz/zclient/participants/ConversationOptionsMenuController.scala
```diff
@@ -0,0 +1,265 @@
...
+  private val users                  = inject[UsersController]
+  private val callingController      = inject[ICallingController]
+  private val cameraController       = inject[ICameraController]
+  private val screenController       = inject[IConversationScreenController]
+
+  override val onMenuItemClicked: SourceStream[MenuItem] = EventStream()
+
+  val onDeleteConv: SourceStream[ConvId] = EventStream[ConvId]()
+
+  def tag: String = if (mode.inConversationList) "OptionsMenu_ConvList" else "OptionsMenu_Participants"
```
makingthematrix: `mode.inConversationList` is a `val`, so it's not going to change. `tag` can be a `lazy val`.

app/src/main/scala/com/waz/zclient/participants/ConversationOptionsMenuController.scala
```diff
@@ -0,0 +1,265 @@
...
+import com.waz.zclient.pages.main.profile.camera.CameraContext
+import com.waz.zclient.participants.ConversationOptionsMenuController._
+import com.waz.zclient.participants.OptionsMenuController._
+import com.waz.zclient.utils.ContextUtils.{getInt, getString}
+import com.waz.zclient.{Injectable, Injector, R}
+
+import scala.concurrent.Future
+import scala.concurrent.duration._
+
+case class ConversationOptionsMenuController(convId: ConvId, mode: Mode)(implicit injector: Injector, context: Context, ec: EventContext) extends OptionsMenuController with Injectable {
```
makingthematrix: A controller should not be a case class. A standard class should be enough.

app/src/main/scala/com/waz/zclient/participants/ConversationOptionsMenuController.scala
```diff
@@ -0,0 +1,265 @@
...
+
+        case Unarchive => convController.archive(cId, archive = false)
+        case Mute   => convController.setMuted(cId, muted = true)
+        case Unmute => convController.setMuted(cId, muted = false)
+        case Leave     => leaveConversation(cId)
+        case Delete    => deleteConversation(cId)
+        case Block     => user.map(_.id).foreach(showBlockConfirmation(cId, _))
+        case Unblock   => zMessaging.head.flatMap { zms =>
+          user.map(_.id) match {
+            case Some(uId) => zms.connection.unblockConnection(uId)
```
makingthematrix: You need to access `zms` only if `user` is `Some(_)`. You can move all that `zMessaging.head.flatMap { zms =>` inside this case's block.

app/src/main/scala/com/waz/zclient/participants/ConversationOptionsMenuController.scala
```diff
@@ -0,0 +1,265 @@
...
+            }).create
+          dialog.show()
+        }
+      }
+    }
+  }
+
+  private def showBlockConfirmation(convId: ConvId, userId: UserId) =
+    (for {
+      curConvId <- convController.currentConvId.map(Option(_)).orElse(Signal.const(Option.empty[ConvId])).head
```
makingthematrix: `currentConvId` is never null, so mapping to `Option` and `orElse` is not necessary.

app/src/main/scala/com/waz/zclient/participants/ConversationOptionsMenuController.scala
```diff
@@ -0,0 +1,265 @@
...
+        override def onClick(dialog: DialogInterface, which: Int): Unit = convController.delete(convId, alsoLeave = true)
+      }).create
+    dialog.show()
+  }
+
+  def deleteConversation(convId: ConvId): Unit = {
+    isGroup.head.flatMap { isGroup =>
+      isMember.head.map { isMember =>
+        if (isGroup && isMember) {
+          val dialog = new AlertDialog.Builder(context, R.style.Theme_Light_Dialog_Alert_Destructive)
```
makingthematrix: You can move that dialog creation out of `if/else` since the only difference is the negative button.

app/src/main/scala/com/waz/zclient/participants/ConversationOptionsMenuController.scala
JPonte: That is not true. currentConvId is an empty signal after log in.

app/src/main/scala/com/waz/zclient/participants/ConversationOptionsMenuController.scala
```diff
@@ -0,0 +1,265 @@
...
+}
+
+object ConversationOptionsMenuController {
+
+  sealed trait Mode {
+    val inConversationList: Boolean
+  }
+  case class Normal(inConversationList: Boolean) extends Mode
+  case class Deleting(inConversationList: Boolean) extends Mode
+  case class Leaving(inConversationList: Boolean) extends Mode
```
deanrobertcook: I think it's a good idea to put the sub classes inside the companion object, so:
``` 
object Mode {
  case class Normal(inConversationList: Boolean) extends 
  case class Deleting(inConversationList: Boolean) extends Mode
  case class Leaving(inConversationList: Boolean) extends Mode
}
```

since by reference, `Mode.Normal` makes more sense than `ConversationOptionsMenuController.Normal`

PR #1591, opened 2018-06-04
Fix ephemeral messages handling in MessageNotificationsController

makingthematrix:
This is the code from https://github.com/wireapp/wire-android/pull/1379 (which itself was a follow-up on https://github.com/wireapp/wire-android/pull/1287). It contains a new version of `MessageNotificationsController` which produces notifications in form of instances of a non-Android case class `NotificationProps`. `NotificationProps` are then used by `NotificationManagerWrapper` to actually create a notification in Android. This way we can finally write non-Android unit tests for  `MessageNotificationsController`.
#### APK
[Download build #11382](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11382/artifact/build/artifact/wire-dev-PR1591-11382.apk)


makingthematrix:
#### Bugs (fixed and possibly still occuring) which should be unit tested:

Bug 1
(Android 6) Receive some normal messages from conversation A. A switches to ephemeral and sends a message. Now all your previous notifications for the normal messages are erased (and you won't see any notification)

Bug 2
(Android 6) receive an ephemeral message in conversation A. Receive a normal message in conversation B (which happened to be a group chat but it's probably irrelevant). Now the notifications are grouped and the sender's name is shown for the ephemeral notification.

Bug 3
(Android 6) Receive some normal messages from conversation A. A switches to ephemeral and sends a message. Now you see the notifications from A grouped under the same title, so you know that "Timed message" was sent by A.

Bug 4
If A sends me 10 ephemeral messages, I shouldn't see 10 different notifications (Android 7).

Bug 5
(Android 6) Receive an ephemeral message. After that, everything other message you receive will not be displayed in the notification.

Bug 6
(Android 6 and 8) Receive two ephemeral messages from the same conversation. The notification we display has title "Someone sent you a message" instead of "Someone". 

Bug 7
Receive ephemeral messages in conversation A. Now receive normal message in conversation A. Now receive a message from conversation B. Notifications from A are cleared.
Update: I don't get notifications from B.

Bug 8
Notifications are not cleared automatically when I open the respective conversations "manually". Basically, if you receive some notifications (with ephemeral messages, though it might happen with normal messages only as well), and then open the app from the home screen, and open the conversation(s) where the messages are, the notifications are not cleared.

Bug 9
Looks like very often the notifications fail to be bundled (note to self: run C651510)
Update: looks like very often some of the notifications that were supposed to be bundled just disappear -- especially if you send messages to the two accounts you have logged in the app.

Bug 10
I do not get notification for incoming connection request.

makingthematrix:
I made some additional changes in a separate commit to make it easier to remove them in case we wanted to go in another direction.
The changes are a step towards using signals instead of calling methods on the wrapper, although as you can see right now I use simple event streams. If there is a notification to cancel, I publish its id in one event stream, the wrapper picks it up, and calls a method on Android; if there is a notification to display, the props are published to another event stream, the wrapper picks it up, builds the notification with the Android's context, and calls the display method on Android.

This way I was able to remove the context from the controller. Only the Android wrapper uses it.

Also, a part of the processing is now taken outside of the UI thread.

JPonte:
build again

someoneEsle:
- [ ] Bug 11
(Android 8) If a person sends me 10 messages, we show the 5 most recent ones but we don't show the  "x more messages". Also, the name of the conversation turns bold the moment you have more than 5 incoming notifications -- and you also add the "." that separates the conversation name from the message that says "X more messages", though it is not displayed)

someoneEsle:
- [ ] Bug 12
A sends me message m1. I swipe the notification away. A sends me message m2. The notification that appears now should show both m1 and m2, but it doesn't.

makingthematrix:
build again

someoneEsle:
- [ ] Bug 13
(Android 8) Receive two ephemeral messages. Tapping the notification doesn't take you to the conversation

someoneEsle:
- [ ] Bug 15
We shouldn't redraw the notification every time a new message comes in.
E.g.
- A sends me a message
- I pull down the notification tray
- A sends me another message
- the notification should be updated with the new message, but it shouldn't be cleared and readded. This also prevents me from using quick replies while A sends me a message

someoneEsle:
- [ ] Bug 16
(Android 8) I'm logged in with accounts A and B. I'm looking at A's conversations list. I receive a notification from a conversation in account B. I tap it. While the conversation is opening, I see the same notification being displayed again. (this does not currently happen in internal)

app/src/main/scala/com/waz/zclient/notifications/controllers/MessageNotificationsController.scala
```diff
@@ -18,487 +18,421 @@
...
+      val groupedConvs =
+        if (bundleEnabled)
+          nots.groupBy(_.convId).map { case (convId, ns) => toNotificationConvId(userId, convId) -> ns }
+        else
+          Map(toNotificationGroupId(userId) -> nots)
+
+      if (bundleEnabled) {
+        val notIdSet = groupedConvs.keys.toSet
+        inject[NotificationManagerWrapper].getActiveNotificationIds
+          .collect { case id if !notIdSet.contains(id) => id }
```
JPonte: Don't do this here. `notIdSet` contains only the notifications from one of the users, which means that if we have two accounts active the notifications are cleared every time.
Also, `getActiveNotification` contains summary notifications as well but `notIdSet` doesn't.

app/src/main/scala/com/waz/zclient/notifications/controllers/MessageNotificationsController.scala
```diff
@@ -18,487 +18,421 @@
...
+          .foreach { c =>
+            verbose(s"a notification from active notifications ids is being cancelled: $c")
+            notificationToCancel ! c
+          }
 
-      val groupedConvs = nots.groupBy(_.convId)
-      val isGrouped = groupedConvs.keys.size > 1
+        val summaryId = toNotificationGroupId(userId)
+        if (notIdSet.contains(summaryId)) {
```
JPonte: Why are you cancelling summary notifications here? It seems `notIdSet` will never contain `summaryId` when `bundleEnabled` is `true`. And you're recreating them again right afterwards.

app/src/main/scala/com/waz/zclient/notifications/controllers/MessageNotificationsController.scala
```diff
@@ -18,487 +18,421 @@
...
+    val convIds = ns.map(_.convId).toSet
+    val isSingleConv = convIds.size == 1
 
-    var color = sharedPreferences.getInt(UserPreferencesController.USER_PREFS_LAST_ACCENT_COLOR, -1)
-    if (color == -1) {
-      color = getColor(R.color.accent_default)
-    }
+    val header =
+      if (isSingleConv) {
+        if (ns.exists(_.isEphemeral))
```
JPonte: This isn't enough. If someone sends a normal message their name is visible. If after that they send an ephemeral one it will get mixed up in the same notification and their name will change to `Someone`. Ephemeral notifications have to be filtered out at the beginning and you have to create another notification for them.

makingthematrix: I talked to Design about this, because in fact it would be easier to treat ephemeral notifications separately. They want it this way.

app/src/main/res/layout/preference_bottom_padding.xml
```diff
@@ -25,8 +25,8 @@
              android:paddingBottom="@dimen/wire__padding__big"
     >
 
-    <include layout="@layout/asp_preference"
+<!--    <include layout="@layout/asp_preference"
```
JPonte: This can be deleted.

app/src/test/scala/com/waz/testutils/TestUtils.scala
```diff
@@ -64,17 +64,19 @@ object TestUtils {
...
+
 }
 
 
 abstract class TestWireContext extends WireContext {
   override def eventContext = EventContext.Implicits.global
 }
 
-class ViewTestActivity extends FragmentActivity with ActivityHelper {
+/*class ViewTestActivity extends FragmentActivity with ActivityHelper {
```
JPonte: Remove.

app/src/test/scala/com/waz/zclient/common/views/ChatheadControllerTest.scala
```diff
@@ -17,27 +17,12 @@
...
-import org.robolectric.RobolectricTestRunner
-import org.robolectric.annotation.Config
+import org.scalatest.Ignore
 import org.scalatest.junit.JUnitSuite
 
-import scala.collection.GenSet
-
-@RunWith(classOf[RobolectricTestRunner])
-@Config(manifest=Config.NONE)
+@Ignore
```
JPonte: Unless you are going to fix the test, just remove the file.

app/src/test/scala/com/waz/zclient/messages/parts/footer/FooterViewControllerTest.scala
```diff
@@ -21,35 +21,24 @@ import java.util.concurrent.TimeUnit
...
 import org.scalatest.junit.JUnitSuite
-import org.threeten.bp.Instant
 
+import scala.concurrent.ExecutionContext
 import scala.concurrent.duration._
-import scala.concurrent.{Await, ExecutionContext}
 
-@RunWith(classOf[RobolectricTestRunner])
-@Config(manifest = "src/test/AndroidManifest.xml", resourceDir = "../../build/intermediates/res/merged/dev/debug")
+@Ignore
```
JPonte: This file can be deleted too.

app/src/test/scala/com/waz/zclient/notifications/controllers/MessageNotificationsControllerTest.scala
```diff
@@ -1,100 +1,421 @@
...
+    sendInfos(infos)
+
+    waitForOne { props =>
+      props.contentTitle.exists(_.header.args == AnyRefArgs(List(3.asInstanceOf[AnyRef], "9"))) && // nine notifications from three conversations
+      props.contentText.exists(_.body == ResString(0, 0, StringArgs(List("3:9")))) // the main message is the last one
+    }
   }
+
+  @Test
+  def receiveTwoTextsFromOneUserInABundle(): Unit = {
```
JPonte: Maybe add some tests for de cancelling of notifications too?

someoneEsle:
- [ ] Bug 17
(Android 6) Receive 3 messages in 2 conversations. The notification says "2 messages in 3 conversations" instead of "3 messages in 2 conversations"

someoneEsle:
- [ ] Bug 18
(Found on Android 6, but probably happens on 8 as well) When I receive a notification from my work account, we should display the team name in the notification (this is how it currently works in internal)

someoneEsle:
Bug 16 is still not fixed

someoneEsle:
- [ ] Bug 20
I have the app open on the conversations list. A writes me. I do not see the subtitle in the conversations list updating properly. However, the conversation *does* go to the top of the list, which tells me that I've actually received the message

someoneEsle:
- [ ] Bug 21
(Android 8, note to self: test on Android 6) Receive a normal message in a group chat. The notification contains the name of the group in bold, then a row with ContactName: message.
Now receive an ephemeral message in that group chat.
The title of the notification (where the conversation name was displayed) turns into "someone". I'm not sure what the behavior in this scenario should be, but this is not how internal behaves -- and I doubt that this is what design wants. :)

JPonte:
Regarding ephemeral messages, a separate notification should be created for each conversation that contains a timed message. These notifications should be obfuscated.

someoneEsle:
- [ ] Bug 20
(Android 8) I'm logged in with 2 accounts, A1 and A2. I have some notifications from multiple conversations in both accounts. 
I swipe away the group of notifications from A1.
I receive a notification in a conversation from A2.
All the notifications from A1 come back.

makingthematrix:
build again

PR #1595, opened 2018-06-05
Automatically set the speaker on when turning on video

JPonte:


app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
```diff
@@ -204,6 +204,7 @@ class CallController(implicit inj: Injector, cxt: WireContext, eventContext: Eve
     verbose(s"toggleVideo")
     updateCall { case (call, cs) =>
       import VideoState._
+      if (call.videoSendState == Stopped) callingZms.head.flatMap(_.mediamanager.setSpeaker(true))
```
deanrobertcook: Maybe we should be doing this in SE?

JPonte: Maybe. Not sure if we want every `setVideoSendState(Started)` to also turn on the speaker or just the calls from this button.

deanrobertcook: well yeah - if the video is on, then so the speaker should be too, right?

deanrobertcook:
logic moved to SE as discussed

PR #1605, opened 2018-06-10
Allow sharing of multiple files at once

kingjan1999:
This is achieved by setting an appropriate flag when the intent is opened. For maximum compatibility, this change was only made for file sharing and only for Android versions from version 4.4 (Kitkat).


AndroidBob:

deanrobertcook:
build again

PR #1616, opened 2018-06-12
Add information about the calling type in the notification

JPonte:


app/src/main/scala/com/waz/zclient/notifications/controllers/CallingNotificationsController.scala
```diff
@@ -108,15 +111,22 @@ class CallingNotificationsController(implicit cxt: WireContext, eventContext: Ev
     toCancel.foreach(notificationManager.cancel(CallNotificationTag, _))
 
     nots.foreach { not =>
+      val title = if (not.isGroup) not.convName else not.caller
+      val message = if (not.isGroup)
```
makingthematrix: In case of nested if/elses consider using brackets ({}) even if there's only one line of code in the expression. Improves readability and guards against mistakes if someone changes the code afterwards.
Also, in this case, a match/case could be a good idea as well.
```
(not.isGroup, not.videoCall) match {
  case (true, true) => ...
  ...
}
```

PR #1622, opened 2018-06-21
okHttp related changes

yeryomenkom:
Remove old http client imports. Refactor ChangeHandleFragment.


someoneEsle:
- [ ] Bug 1
This may have been a one-time/wifi thing but it's still worth investigating since it might be bad.
I was logged in with two accounts, A and B. 
I have the app open on the conversations list of A. 
I go into airplane mode. 
I get messages from both account A and B. 
I turn off airplane mode.
I get the messages from account B, but not from account A.
Even after replying to the conversation in A where the incoming messages should be (which works), I still don't see any of the previous messages.

someoneEsle:
- [ ] Bug 2
1. Go to settings>account>change email
2. type an email that is already associated to a wire account
3. You get an error saying the email is already taken (which is correct)
4. Now type another email, not associated with any wire account
5. You keep getting the error

PR #1632, opened 2018-06-28
Avoid calling unarchive every time a conversation is selected

JPonte:


app/src/main/scala/com/waz/zclient/conversation/ConversationController.scala
```diff
@@ -90,7 +90,8 @@ class ConversationController(implicit injector: Injector, context: Context, ec:
       lastConvId = convId
       for {
         z <- zms.head
-        _ <- z.convsUi.setConversationArchived(id, archived = false)
+        conv <- z.convsStorage.get(id)
+        _ <- if (conv.exists(_.archived)) z.convsUi.setConversationArchived(id, archived = false) else Future.successful(Option.empty[ConversationData])
         _ <- z.convsStats.selectConversation(convId)
       } yield { // catches changes coming from UI
```
deanrobertcook: we could probably move all this logic to SE?

PR #1665, opened 2018-07-12
Markdown Fixes

johnxnguyen:
# What's in this PR?
This PR contains mostly design tweaks but also some leftover todos from the previous run.

**Design fixes include:**
- Definitions of colors for code, list prefix and quotes (both in light and dark mode).
- The color of links should be the user's chosen accent color.
- If the user changes accent color (on any device), the link color should update accordingly.
- The quote stripe should be thinner and appear the same color as the quoted text.
- Increase paragraph spacing around list items and quotes.
- Code blocks should not be indented.
- Increase line-height of all text slightly.

**Bug fixes:**
- The paragraph spacing was not respecting screen density (on low-density devices, it was too large).
- HTML blocks were missing a trailing newline.
- Quotes were not rendering on multiple lines

## Note
These changes have already been shown to the design team.

### DIFF
--- a/app/src/androidTest/kotlin/com/waz/zclient/markdown/GroupSpanTest.kt
+++ b/app/src/androidTest/kotlin/com/waz/zclient/markdown/GroupSpanTest.kt
@@ -104,7 +104,7 @@ class GroupSpanTest {
     @Test
     fun testThatItConfiguresSpansAndNode_BlockQuote() {
         // given
-        val sut = BlockQuoteSpan(Color.BLUE, 4, 8, 16, 8)
+        val sut = BlockQuoteSpan(Color.BLUE, Color.GREEN, 4, 8, 16, 8)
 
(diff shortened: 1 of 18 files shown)


someoneEsle:
- [ ] Bug 1
Receive a quote on multiple lines. I can see the "```" at the end

johnxnguyen:
@someoneEsle This is actually not a bug: The triple tick marks are used to create a code block, but to correctly use them, the tick marks should be on their own line. You can verify this [here](https://markdown-it.github.io).

someoneEsle:
Thank you so much for that link, it'll make my life easier :D

johnxnguyen:
@someoneEsle also, if you use that link, make sure CommonMark Strict is enabled, since that is the specification that the client conforms to.

someoneEsle:
build again

PR #1693, opened 2018-08-01
Bump upload file size to 100MB

makingthematrix:
Requires https://github.com/wireapp/wire-android-sync-engine/pull/394
I replaced the call to `Formatter.formatShortFileSize` with a simple string interpolation, because the former treats a megabyte as a million of bytes, while in SE we define it as 2^20 (1024 * 1024) bytes. As a result, the 100 * 1024 * 1024 bytes limit was rounded up back to megabytes as 105MB.  
#### APK
[Download build #11388](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11388/artifact/build/artifact/wire-dev-PR1693-11388.apk)
[Download build #11413](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11413/artifact/build/artifact/wire-dev-PR1693-11413.apk)
[Download build #11528](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11528/artifact/build/artifact/wire-dev-PR1693-11528.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/views/ConversationFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/views/ConversationFragment.scala
@@ -24,14 +24,13 @@ import android.provider.MediaStore
 import android.support.annotation.Nullable
 import android.support.v7.widget.{ActionMenuView, Toolbar}
 import android.text.TextUtils
-import android.text.format.Formatter
 import android.view._
 import android.view.animation.Animation
 import android.widget.{AbsListView, FrameLayout, TextView}
 import com.waz.ZLog.ImplicitTag._
 import com.waz.ZLog._
 import com.waz.model.AccentColor
-import com.waz.api.{AssetFactory, AudioAssetForUpload, AudioEffect, ErrorType}
+import com.waz.api.{AudioAssetForUpload, AudioEffect, ErrorType}
 import com.waz.content.GlobalPreferences
 import com.waz.model.ConversationData.ConversationType
 import com.waz.model.{MessageContent => _, _}
@@ -671,8 +670,7 @@ class ConversationFragment extends FragmentHelper {
       )
       errorsController.dismissSyncError(err.id)
     case ErrorType.CANNOT_SEND_ASSET_TOO_LARGE =>
-      val maxAllowedSizeInBytes = AssetFactory.getMaxAllowedAssetSizeInBytes
-      if (maxAllowedSizeInBytes > 0) {
+      accountsController.isTeam.head.foreach { isTeam =>
         val dialog = ViewUtils.showAlertDialog(
           getActivity,
           R.string.asset_upload_error__file_too_large__title,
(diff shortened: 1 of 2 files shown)


app/src/main/scala/com/waz/zclient/views/ConversationFragment.scala
```diff
@@ -731,8 +731,7 @@ class ConversationFragment extends BaseFragment[ConversationFragment.Container]
       )
       errorsController.dismissSyncError(err.id)
     case ErrorType.CANNOT_SEND_ASSET_TOO_LARGE =>
-      val maxAllowedSizeInBytes = AssetFactory.getMaxAllowedAssetSizeInBytes
-      if (maxAllowedSizeInBytes > 0) {
+      if (AssetData.MaxAllowedBackendAssetSizeInBytes > 0) {
```
deanrobertcook: hmm, this seems like a pointless check?

app/src/main/scala/com/waz/zclient/views/ConversationFragment.scala
```diff
@@ -741,7 +740,7 @@ class ConversationFragment extends BaseFragment[ConversationFragment.Container]
           null,
           true
         )
-        dialog.setMessage(getString(R.string.asset_upload_error__file_too_large__message, Formatter.formatShortFileSize(getContext, maxAllowedSizeInBytes)))
+        dialog.setMessage(getString(R.string.asset_upload_error__file_too_large__message, s"${AssetData.MaxAllowedBackendAssetSizeInBytes / (1024 * 1024)}MB"))
```
deanrobertcook: nice!

makingthematrix:
build again

mykola-mokhnach:
This is what I see in the conversation if I send 20MB-large picture as a file via webapp:
http://media.corporate-ir.net/media_files/IROL/17/176060/EchoDot/Echo%20Dot-White,%20Low%20Angle,%20Front,%20On.jpg


screen

mykola-mokhnach:
Other media files (mp3/mp4, > 40MB) are never getting uploaded. I was waiting for more than 20 minutes until it finally failed to upload these files

mykola-mokhnach:
There is no visual feedback for large video download. It took me up to one minute between I clicked play button on the downloaded video preview (40 MB .mp4 file) and the playback has actually started.

mykola-mokhnach:
1. Receive some big binary file 
2. Tap it to init download
3. Tap the X icon to cancel the download

AR:
The download is still in progress. X button has no effect

ER:
The download should be stopped after STR

mykola-mokhnach:
The size of received mp4 files is always shown as 0:00, but such files can be played properly though

mykola-mokhnach:
An empty bar is shown instead of a picture preview if I send HD JPEG files (> 20 MB) from web app

mykola-mokhnach:
Big file download is restarted completely if there is a temporary problem with network connection. I would expect the download to be resumed instead.

mykola-mokhnach:
1. Start uploading some big file
2. Click X button to cancel

AR:
Upload failed error is shown

ER:
The item being uploaded should be deleted from the conversation

CalumMcCall:
So the broken cancel button is an issue I discovered yesterday also. The issue is that we actually haven't implemented cancelling at all, since it was decided that it didn't make sense with files >25MB. Implementing it will take some time, so we are planning to do that in a separate ticket.

CalumMcCall:
Hey @mykola-mokhnach so since we are currently refactoring assets, it doesn't make much sense to start trying to fix these bugs at the moment. Also, given that they all exist in current production already, can we just merge this PR so we can get this in? We are currently in a service run to work on calling, so our resources are going to be focused there for the next couple weeks. The asset refactoring will be completed after that, and then I can try reproducing these bugs then. Sound good?

mykola-mokhnach:
I would say it is not going to make it worse if the issues already existed in prod. Just merge it

PR #1705, opened 2018-08-06
Remove com.waz.api.AccentColor and the old AccentColorController

makingthematrix:
I found one more place, in `DrawingFragment` where a mixup of `com.waz.api.AccentColor` and `com.waz.api.impl.AccentColor` caused a crash on opening. I looked through the rest of UI and replaced `import com.waz.api._` in a few places with what was actually used in the given file. It's safer this way.

Requires https://github.com/wireapp/wire-android-sync-engine/pull/396
#### APK
[Download build #11416](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11416/artifact/build/artifact/wire-dev-PR1705-11416.apk)
[Download build #11421](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11421/artifact/build/artifact/wire-dev-PR1705-11421.apk)
[Download build #11422](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11422/artifact/build/artifact/wire-dev-PR1705-11422.apk)
[Download build #11428](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11428/artifact/build/artifact/wire-dev-PR1705-11428.apk)
[Download build #11430](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11430/artifact/build/artifact/wire-dev-PR1705-11430.apk)
[Download build #11433](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11433/artifact/build/artifact/wire-dev-PR1705-11433.apk)

PR #1710, opened 2018-08-07
Enable TLS 1.2 on lollipop and older devices.

yeryomenkom:


app/src/main/scala/com/waz/zclient/WireApplication.scala
```diff
@@ -284,8 +286,23 @@ class WireApplication extends MultiDexApplication with WireContext with Injectab
...
 
+  private def enableTLS12OnOldDevices(): Unit = {
+    if (Build.VERSION.SDK_INT <= Build.VERSION_CODES.LOLLIPOP) {
+      try {
+        ProviderInstaller.installIfNeeded(getApplicationContext)
+        val sslContext = SSLContext.getInstance("TLSv1.2")
+        sslContext.init(null, null, null)
+        sslContext.createSSLEngine
+      } catch {
+        case error: Throwable =>
```
deanrobertcook: should probably catch `NonFatal(e)` here

app/src/main/java/com/waz/zclient/markdown/MarkdownTextView.java
```diff
@@ -81,15 +81,15 @@ private void configureStyleSheet() {
...
-        }, eventContext());
+//        // update the link color whenever the accent color changes
+//        ((BaseActivity) getContext()).injectJava(AccentColorController.class).accentColorForJava(new AccentColorCallback() {
+//            @Override
+//            public void color(AccentColor color) {
+//                mStyleSheet.setLinkColor(color.getColor());
+//                setLinkTextColor(color.getColor());
+//                refreshLinks();
+//            }
+//        }, eventContext());
```
deanrobertcook: Let's fix this properly and in another commit

yeryomenkom: Fix is in https://github.com/wireapp/wire-android/pull/1705
So, we should merge that PR first. After that I will this branch.

app/src/main/scala/com/waz/zclient/WireApplication.scala
yeryomenkom: Any reasons?

yeryomenkom:
build again

app/src/main/scala/com/waz/zclient/WireApplication.scala
deanrobertcook: Because you don't want to stop the app from crashing if there is an OOO error or an Unsatisfied link error - in fact, you should really figure out which Exceptions these methods throw and only handle those, but at least with NonFatal(e) you're filtering out the really bad things that should crash the app

PR #1723, opened 2018-08-10
Add warning messages to add bot and self added messages

JPonte:
#### APK
[Download build #11446](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11446/artifact/build/artifact/wire-dev-PR1723-11446.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/messages/parts/MemberChangePartView.scala
+++ b/app/src/main/scala/com/waz/zclient/messages/parts/MemberChangePartView.scala
@@ -20,10 +20,11 @@ package com.waz.zclient.messages.parts
 import android.content.Context
 import android.graphics.drawable.Drawable
 import android.util.AttributeSet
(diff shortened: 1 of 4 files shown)


app/src/main/scala/com/waz/zclient/messages/parts/MemberChangePartView.scala
```diff
@@ -114,6 +116,22 @@ class MemberChangePartView(context: Context, attrs: AttributeSet, style: Int) ex
     }
   }
 
+  private val servicesPresentWarning = for {
+    zms <- zMessaging
+    msg <- message
+    members <- users.users(msg.members)
+    convMembersIds <- zms.membersStorage.activeMembers(msg.convId)
```
deanrobertcook: we've started injecting (signals of) services directly instead of mapping from zms. Should make things easier to test in the future

PR #1734, opened 2018-08-15
Start merging logic between start ui and add participant adapter

deanrobertcook:
#### APK
[Download build #11496](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11496/artifact/build/artifact/wire-dev-PR1734-11496.apk)
[Download build #11499](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11499/artifact/build/artifact/wire-dev-PR1734-11499.apk)
[Download build #11502](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11502/artifact/build/artifact/wire-dev-PR1734-11502.apk)
[Download build #11503](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11503/artifact/build/artifact/wire-dev-PR1734-11503.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/conversation/creation/AddParticipantsFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/conversation/creation/AddParticipantsFragment.scala
@@ -24,45 +24,47 @@ import android.support.design.widget.TabLayout.OnTabSelectedListener
 import android.support.v7.widget.{LinearLayoutManager, RecyclerView}
 import android.view._
 import android.view.inputmethod.EditorInfo
-import android.widget.TextView
+import android.widget.{ImageView, TextView}
 import android.widget.TextView.OnEditorActionListener
 import com.waz.ZLog.ImplicitTag._
 import com.waz.ZLog.verbose
-import com.waz.api.impl.ErrorResponse
 import com.waz.model._
-import com.waz.service.{IntegrationsService, UserSearchService, ZMessaging}
 import com.waz.service.tracking.{OpenSelectParticipants, TrackingService}
+import com.waz.service.{IntegrationsService, UserSearchService, ZMessaging}
 import com.waz.threading.Threading
 import com.waz.utils.events._
 import com.waz.utils.returning
 import com.waz.zclient._
 import com.waz.zclient.common.controllers.ThemeController.Theme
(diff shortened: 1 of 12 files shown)


someoneEsle:
- [ ] Bug 1
There should be no "Manage services" button when I have no services

app/src/main/scala/com/waz/zclient/conversation/creation/AddParticipantsFragment.scala
```diff
@@ -214,64 +234,49 @@ case class AddParticipantsAdapter(usersSelected: SourceSignal[Set[UserId]],
...
+        case Users(us) => us
+        case _ => Seq.empty
+      }
+
+      val integrationResults = res match {
+        case Services(ss) => ss
+        case _ => Seq.empty
+      }
+
+      this.results = userResults.map(u => (Left(u), usersSelected.contains(u.id))) ++ integrationResults.map(i => (Right(i), servsSelected.contains((i.provider, i.id))))
```
makingthematrix: In all cases either `userResults` or `integrationResults` is empty, so instead of summing them, we could map `this.results` directly from `res`:
```
this.results = res match {
  case Users(us) => us.map(u => Left(u), => usersSelected.contains(u.id)))
  case Services(ss) => ss.map(s => (Right(s), servsSelected.contains((s.provider, s.id))))
}
```

But this is a part of a bigger issue. 

Notice how from the new `SearchController` to `AddParticipantsFragment` to `SingleUserRowView` we constantly mix and separate `UserData` and `IntegrationData`. We `Either` and `match/case` are all over the place - because from the UI perspective, and in how we search them, rows of searched users and rows of searched services look very similar. The row is of the same size, there is a (slightly different) avatar, a name, a subtitle, and when the users clicks it, another view appears.

I would suggest to have a common case class, eg. `SearchResultInfo`, which would contain all data needed to display the row in UI, plus the id that would tell us is it a user or a service, and which one is it (eg.`Either[UserId, (ProviderId, IntegrationId)]`). This `ResultResultInfo` would be created directly both from `IntegrationData` and `UserData` in `SearchController` and used everywhere else. `AddParticipantsFragment`, `SingleUserRowView`, and `ChatheadView` could be simplified, as well as the traits `AddUserListState` and `SearchUserListState`.

PR #1786, opened 2018-09-11
fix: prevent websocket notification from flickering.

deanrobertcook:
It seems as though the signal for deciding to show the ws notification
  can flicker very quickly to true before it is then set back to false
  when the app is minimised (for those users where the WebSocketService
  should be not active).

  This commit throttles the signal to prevent this flicker.

  fixes: https://github.com/wireapp/android-project/issues/262
#### APK
[Download build #11670](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11670/artifact/build/artifact/wire-dev-PR1786-11670.apk)


deanrobertcook:
Didn't fix

PR #1812, opened 2018-09-28
feat: Highlight user's mentions in notifications

JPonte:
Mentions targeting the user should be bold in the message notification text

depends on: https://github.com/wireapp/wire-android-sync-engine/pull/430
#### APK
[Download build #11785](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11785/artifact/build/artifact/wire-dev-PR1812-11785.apk)


someoneEsle:
As agreed, we can just close this PR since the correct behavior is already in RC

PR #1815, opened 2018-10-02
Mentions notifications

JPonte:


someoneEsle:
- [ ] Bug 1
When I set "only mentions" on another device, I see the mute icon in the conversations list but I don't see the right option ticked in the bottom sheet menu. C725827

someoneEsle:
- [ ] Bug 2
Set a conversation to Notifications>Nothing. Get a call. Don't pick up. The subtitle in the conversations list says "Missed call from ..." instead of being summarised.

JPonte:
build again

someoneEsle:
- [ ] Bug 4
The following was not specified in the specs but I talked to @eike96 .
- Both 1:1 and group conversations should display the "@" glyph in the conversations list if they were set to "mentions only"
- if I'm a personal user, I mute a conversation (either 1:1 or group), I should not see the "@" glyph if I get mentioned

PR #1837, opened 2018-10-16
fix: add job for checking registered PushToken

deanrobertcook:
This commit provides an Evernote Job for, and executes, the check to see
if the current user has a correctly registered push token.

The evernote job handles scheduling, back-off and rescheduling of the
work, while the job itself will be triggered from inside the
WebSocketBroadcastReceiver (which I will now rename to
StartAndUpgradeBroadcastReceiver). This should be sufficient to
ensure that anyone with the bug will be re-registered the next time
they start the app after upgrading, and every device restart should
be a regular enough check to make sure the token is still valid.

partly fixes: https://github.com/wireapp/android-project/issues/263
#### APK
[Download build #11868](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11868/artifact/build/artifact/wire-dev-PR1837-11868.apk)
[Download build #11872](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11872/artifact/build/artifact/wire-dev-PR1837-11872.apk)

### DIFF
--- a/app/src/main/scala/com/waz/services/websocket/WebSocketService.scala
+++ b/app/src/main/scala/com/waz/services/websocket/WebSocketService.scala
@@ -26,6 +26,7 @@ import android.support.v4.app.NotificationCompat
 import com.waz.ZLog.ImplicitTag._
 import com.waz.ZLog._
 import com.waz.content.GlobalPreferences.{PushEnabledKey, WsForegroundKey}
+import com.waz.jobs.PushTokenCheckJob
 import com.waz.service.AccountsService.InForeground
 import com.waz.service.{AccountsService, GlobalModule, ZMessaging}
 import com.waz.threading.Threading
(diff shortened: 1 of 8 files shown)


app/src/main/scala/com/waz/services/websocket/WebSocketService.scala
```diff
@@ -87,10 +88,18 @@ class WebSocketBroadcastReceiver extends BroadcastReceiver {
   private lazy val controller =
     injector.binding[WebSocketController].getOrElse(throw new Exception(s"Failed to load WebSocketController")).apply()
 
+  private lazy val accounts =
+    injector.binding[AccountsService].getOrElse(throw new Exception(s"Failed to load WebSocketController")).apply()
```
makingthematrix: The exception message should be "Failed to load AccountsService".

deanrobertcook:
build again

PR #1845, opened 2018-10-18
fix: Don't crop photos taken with the internal camera

makingthematrix:
fixes https://github.com/wireapp/android-project/issues/258

When taking a photo, we choose from the list of possible resolutions the format which is equal, or only a bit bigger in height than `ImageAssetGenerator.MediumSize`. Usually it means a resolution with ratio 16:9, but in some cases the chosen one was 4:3 or even 1:1. As a result, even though the user saw a 16:9 picture on the screen, the photo taken was taken in the chosen resolution's ratio.
This PR fixes that by preferring 16:9 resolutions. Only if there is none, a different resolution will be chosen. 
Also, the preview will show all the photo, scaled down to fit the screen. Until now the photo was cropped instead, showing only the center part.
#### APK
[Download build #11884](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11884/artifact/build/artifact/wire-dev-PR1845-11884.apk)
[Download build #11885](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11885/artifact/build/artifact/wire-dev-PR1845-11885.apk)
[Download build #11949](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11949/artifact/build/artifact/wire-dev-PR1845-11949.apk)
[Download build #11989](http://192.168.120.33:8080/job/Pull%20Request%20Builder/11989/artifact/build/artifact/wire-dev-PR1845-11989.apk)
[Download build #12005](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12005/artifact/build/artifact/wire-dev-PR1845-12005.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/camera/controllers/GlobalCameraController.scala
+++ b/app/src/main/scala/com/waz/zclient/camera/controllers/GlobalCameraController.scala
@@ -313,12 +313,12 @@ class AndroidCamera(info: CameraInfo, texture: SurfaceTexture, w: Int, h: Int, c
     PreviewSize(w, h)
   }
 
-  private def getPictureSize(pms: CameraParamsWrapper) = {
-    val (bigger, smaller) = pms.get.getSupportedPictureSizes.asScala.map(new CameraSizeWrapper(_)).partition { size =>
-      size.height >= ImageAssetGenerator.MediumSize
+  private def getPictureSize(pms: CameraParamsWrapper) =
+    pms.get.getSupportedPictureSizes.asScala.map(new CameraSizeWrapper(_)).minBy { s =>
+      val is16_9 = math.abs( (s.width.toDouble / s.height) - GlobalCameraController.Ratio_16_9 ) < 0.01
+      val differenceToHeight = Math.abs(s.height - ImageAssetGenerator.MediumSize)
+      (!is16_9, differenceToHeight) //Ordering[Boolean] considers false < true, so we want to flip the is16_9 check to give them preference
     }
-    if (bigger.nonEmpty) bigger.minBy(_.width) else smaller.maxBy(_.width)
-  }
 
   /**
     * activityRotation is relative to the natural orientation of the device, regardless of how the device is rotated.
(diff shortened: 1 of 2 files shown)


someoneEsle:
build again

app/src/main/scala/com/waz/zclient/camera/controllers/GlobalCameraController.scala
```diff
@@ -314,9 +314,16 @@ class AndroidCamera(info: CameraInfo, texture: SurfaceTexture, w: Int, h: Int, c
...
-    val (bigger, smaller) = pms.get.getSupportedPictureSizes.asScala.map(new CameraSizeWrapper(_)).partition { size =>
+    def partition(sizes: Seq[CameraSizeWrapper]) = sizes.partition { size =>
       size.height >= ImageAssetGenerator.MediumSize
     }
+
+    val sizes = pms.get.getSupportedPictureSizes.asScala.map(new CameraSizeWrapper(_))
+    val sizes_16_9 = sizes.filter { s => s.width.toDouble / s.height == GlobalCameraController.Ratio_16_9 }
+
+    // if it's possible, we want to use a camera with 16:9 ratio
+    val (bigger, smaller) = partition(if (sizes_16_9.nonEmpty) sizes_16_9 else sizes)
```
deanrobertcook: What we are essentially doing here is looking for the camera size where the aspect ratio is closest to `MediumSize` (if it's bigger, getMin, if it's smaller get max), but first giving preference to those where the aspect ratio is 16:9. 

By instead ordering by the **difference** to `MediumSize`, and using `Ordering[Boolean]` to effectively partition the collection, we can do this in one method call:

```
pms.get.getSupportedPictureSizes.asScala.map(new CameraSizeWrapper(_)).minBy { s =>
      val is16_9 = s.width.toDouble / s.height == GlobalCameraController.Ratio_16_9
      val differenceToHeight = Math.abs(s.height - ImageAssetGenerator.MediumSize)
      (!is16_9, differenceToHeight) //Ordering[Boolean] considers false < true, so we want to flip the is16_9 check to give them preference
    }
    
```

deanrobertcook: (please note I didn't fix the floating point comparison)

Issue #1835, opened 2018-10-15
Disable Google Analytics introduced by FCM (short-term) and replace FCM by in-house notification solution (mid-term, see Tutanota)


baimafeima:
This was previously raised here and I have copied the template for sake of simplicity: https://github.com/siacs/Conversations/issues/3041

Since you have introduced Firebase, you might need to explicitly disable Google Analytics:

Source: https://firebase.google.com/docs/analytics/

    

> Permanently deactivate collection
> 
>     If you need to deactivate Analytics collection permanently in a version of your app, set firebase_analytics_collection_deactivated to true in your app's AndroidManifest.xml in the application tag. For example:
> 
>     <meta-data android:name="firebase_analytics_collection_deactivated" android:value="true" />
> 
>     Disable Advertising ID collection
> 
>     If you wish to disable collection of the Advertising ID in your Android app, you can set the value of google_analytics_adid_collection_enabled to false in your app's AndroidManifest.xml in the application tag. For example:
> 
>     <meta-data android:name="google_analytics_adid_collection_enabled" android:value="false" />

Source: https://firebase.google.com/support/guides/disable-analytics

See also https://stackoverflow.com/questions/39327141/disable-the-analytics-service-for-firebase-cloud-messaging

deanrobertcook:
Hi @baimafeima, thanks a lot, I'll look into it ASAP

baimafeima:
@deanrobertcook Thanks for looking into this. Wire wouldn't be the only privacy-focused app affected by this nasty Google move. But as many have requested that Wire be included in F-Droid instead of Google Play, eliminating FCM alltogether might remove one of the last blockers if I am not mistaken. I did some more research on this and it seems that Tutanota, a privacy-focused e-mail provider succeeded in eliminating FCM by developing their own [notification system](https://f-droid.org/en/2018/09/03/replacing-gcm-in-tutanota.html). The fact that Google includes tracking code and that this is proprietary - and your own open source policies - should be reason enough to completely eliminate this.

deanrobertcook:
@baimafeima apologies for the delay - after looking through the docs a bit, I don't think that there was anything meaningful that the tracking could have been collecting, but definitely as an added safety measure it makes sense to disable this tracking service. 

As for replacing the notification system, this represents a lot more work for us that we unfortunately can't afford right now. If you'd like to open another ticket that focuses on this separately, then please do! (this one will be closed when the PR is merged) 

Thanks again for bringing this up!

[...]

PR #1860, opened 2018-11-06
fix: disable FCM analytics

deanrobertcook:
It seems as though the core FCM library (which would probably be a transient
dependancy of the FCM notifications library) seems to also include the
analytics logic of FCM.

This commit explicity disables analytics, just in case.

fixes: https://github.com/wireapp/wire-android/issues/1835

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -231,6 +231,10 @@
                 android:resource="@xml/provider_paths"/>
         </provider>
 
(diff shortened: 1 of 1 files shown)


JPonte:
Isn't this one also needed?
```<meta-data android:name="google_analytics_adid_collection_enabled" android:value="false" />```

From https://firebase.google.com/support/guides/disable-analytics

JPonte:
Nevermind

PR #1881, opened 2018-11-22
fix: Correctly display buttons in conversation list in RTL

JPonte:
fixes: https://github.com/wireapp/android-project/issues/338
#### APK
[Download build #12109](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12109/artifact/build/artifact/wire-dev-PR1881-12109.apk)


CalumMcCall:
@JPonte what are the STR? When I switch to RTL I can't reproduce this bug.

JPonte:
Just opening the app after killing it and you should see both buttons on the right side. Make sure you have some archived conversations too.

PR #1922, opened 2018-12-21
feat: Allow log in with an SSO id

makingthematrix:
SSO login ("Company login") is now enabled on production.
If the user is logged in via SSO, it overrides email, phone, and password - wherever these three are requested, if SSO Id is present, they are not needed anymore.
In case of removing a client we still want the user to confirm that this is what they want, since a client removed by accident might be a bad mistake. In that case, `RemoveDeviceDialog` shows only "Are you sure?" and no edit box for the password.

SE: https://github.com/wireapp/wire-android-sync-engine/pull/460

### DIFF
--- a/app/src/main/scala/com/waz/zclient/MainActivity.scala
+++ b/app/src/main/scala/com/waz/zclient/MainActivity.scala
@@ -189,9 +189,11 @@ class MainActivity extends BaseActivity
               pendingPw    <- z.userPrefs(PendingPassword).apply()
               pendingEmail <- z.userPrefs(PendingEmail).apply()
               handle       <- z.users.selfUser.map(_.handle).head
+              ssoLogin     <- accountsService.activeAccount.map(_.exists(_.ssoId.isDefined)).head
             } yield {
               val (f, t) =
-                if (email.isDefined && pendingPw)                 (SetOrRequestPasswordFragment(email.get), SetOrRequestPasswordFragment.Tag)
+                if (ssoLogin)                                     (new MainPhoneFragment,                   MainPhoneFragment.Tag)
+                else if (email.isDefined && pendingPw)            (SetOrRequestPasswordFragment(email.get), SetOrRequestPasswordFragment.Tag)
                 else if (pendingEmail.isDefined)                  (VerifyEmailFragment(pendingEmail.get),   VerifyEmailFragment.Tag)
                 else if (email.isEmpty && isLogin && isNewClient
                   && phone.isDefined)                             (AddEmailFragment(),                      AddEmailFragment.Tag)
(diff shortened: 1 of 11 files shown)


marcoconti83:
@makingthematrix Do we have a localizable string for the popup?

marcoconti83:
@makingthematrix Perfect, thanks for clarifying!

red-avtovo:
Bug1:
Logged in user doesn't see Unique username takeover screen

red-avtovo:
Bug2:
User name is show like a part of email address, but not the real one

red-avtovo:
Bug3:
Add email field is shown (fortunately it is not possible to change/add email address)

red-avtovo:
Bug4:
Reset password button is visible

red-avtovo:
Bug 5:
Delete account button is visible

julesmummdry:
Bug 6:
Restoring backup for SSO user does not work. After pressing the button to choose the backup file, I see a modal saying "Something went wrong - Your history can not be restored".
For SSO users we do not have a email and pw that means restoring the backup should work without it.

makingthematrix:
The bug #6 is not connected to SSO. Its main cause is in TestingGallery which sends an invalid intent to the application. If you test this PR without TestingGallery, everything works.

However, we could have a more detailed information in the popup, instead "Something went wrong". 
I will create a separate ticket for it.

makingthematrix:
Fix for Bug 1 is there. The screen for setting the handle will appear if the handle is empty.

app/src/main/scala/com/waz/zclient/MainActivity.scala
```diff
@@ -205,21 +208,25 @@ class MainActivity extends BaseActivity
...
             }
           case Right(PasswordMissing) =>
             for {
               self         <- am.getSelf
               pendingEmail <- am.storage.userPrefs(PendingEmail).apply()
+              ssoLogin     <- accountsService.activeAccount.map(_.exists(_.ssoId.isDefined)).head
             } {
               val (f ,t) =
-                if (self.email.isDefined)        (SetOrRequestPasswordFragment(self.email.get, hasPassword = true), SetOrRequestPasswordFragment.Tag)
+                if (ssoLogin)                    (new MainPhoneFragment,                                            MainPhoneFragment.Tag)
```
JPonte: Will this ever get here? If we get a PasswordMissing error we don't have cookie/token right? Going to the MainPhoneFragment won't do anything. Not sure we can even have a PasswordMissing in SSO to begin with.

app/src/main/scala/com/waz/zclient/preferences/dialogs/RemoveDeviceDialog.scala
```diff
@@ -77,15 +79,21 @@ class RemoveDeviceDialog extends DialogFragment with FragmentHelper {
     _.onClick(inject[BrowserController].openForgotPasswordPage())
   }
 
+  private var isSSO = false
```
JPonte: Why not a lazy val or def?

app/src/main/scala/com/waz/zclient/MainActivity.scala
makingthematrix: True. This check is not needed here. I'll remove it. 
It should not affect the functionality at all.

app/src/main/scala/com/waz/zclient/preferences/pages/DeviceDetailsView.scala
```diff
@@ -233,23 +233,23 @@ case class DeviceDetailsViewController(view: DeviceDetailsView, clientId: Client
...
       limitReached <- am.clientState.head.map {
         case LimitReached => true
         case _ => false
       }
-      _ <- am.deleteClient(clientId, password).map { //TODO use password instead of str
+      _ <- am.deleteClient(clientId, password).map {
         case Right(_) =>
           for {
-            _ <- passwordController.setPassword(password)
+            _ <- password.fold(Future.successful({}))(passwordController.setPassword)
```
yeryomenkom: Just a style issue. Always prefer to use instance of Unit type explicitly instead of empty block.
So should be Future.successful(()) instead of Future.successful({})

PR #1931, opened 2019-01-02
fix: Catch exceptions linked to invalid date/time patterns in timestamps

makingthematrix:
Date patterns are localized, which means we don't have full control over them.
It might happen that the translator makes a typo and the pattern becomes invalid and crashes the app at runtime, when we try to display a timestamp. 
For this reason we need to safeguard date and time patterns. If there is a problem with it, the app should fall back to a default, hardcoded pattern which we know to be safe.
#### APK
[Download build #12209](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12209/artifact/build/artifact/wire-dev-PR1931-12209.apk)
[Download build #12210](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12210/artifact/build/artifact/wire-dev-PR1931-12210.apk)
[Download build #12211](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12211/artifact/build/artifact/wire-dev-PR1931-12211.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/utils/Time.scala
+++ b/app/src/main/scala/com/waz/zclient/utils/Time.scala
@@ -19,10 +19,12 @@ package com.waz.zclient.utils
 
 import android.content.Context
 import android.text.format.DateFormat
+import com.waz.ZLog.ImplicitTag.implicitLogTag
+import com.waz.log.ZLog2._
 import com.waz.zclient.R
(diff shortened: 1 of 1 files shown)


CalumMcCall:
Looks good to me, although maybe consider just merging the two try catch blocks as they are identical apart from the error messages. Maybe you could just remove the date/time words and just say they pattern is invalid?

PR #1935, opened 2019-01-07
fix: Tapping 'Send' repeatedly may cause the same message to be sent many times

makingthematrix:
fixes https://github.com/wireapp/android-project/issues/354

We can experience delays between when the user taps the 'Send' button and when the message is actually sent and, in consequence, the editbox is cleared. If the 'Send' button is tapped again before that, the message is being sent again.

This fix introduces a signal of messages being sent. If the 'Send' button is tapped twice, the app will
recognize the message is already being processed and ignore it. In the same time, it does not change anything from the user's perspective - another solution would be to clear the editbox immediately, but that would introduce a short time when the message is gone from the screen.
#### APK
[Download build #12212](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12212/artifact/build/artifact/wire-dev-PR1935-12212.apk)
[Download build #12216](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12216/artifact/build/artifact/wire-dev-PR1935-12216.apk)
[Download build #12222](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12222/artifact/build/artifact/wire-dev-PR1935-12222.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/cursor/CursorController.scala
+++ b/app/src/main/scala/com/waz/zclient/cursor/CursorController.scala
@@ -27,7 +27,7 @@ import com.google.android.gms.common.{ConnectionResult, GoogleApiAvailability}
 import com.waz.ZLog.ImplicitTag._
 import com.waz.api.NetworkMode
 import com.waz.content.{GlobalPreferences, UserPreferences}
-import com.waz.model.{ConvExpiry, Mention, MessageData}
+import com.waz.model._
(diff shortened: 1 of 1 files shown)


JPonte:
Since you now keep track of the message being sent in a signal, you can listen to that signal and block the send button in a conversation that contains an outgoing message.

red-avtovo:
found this issue on Master, so we should fix it here or merge this PR and open an issue

PR #1967, opened 2019-01-30
Add feature flag for "data permission" options in account

marcoconti83:

Need to add feature flags for some options in the data permission settings, namely:
- Send anonymous crash reports
- Allow marketing communication

We want to hide those options completely when the feature flag is removed.

# Changes

Removed the to individual options if the flags are false, and also removed the main "data permission" setting button if both flags are false, as there is no point in having an empty section.

# Testing
See the issue with testing in https://github.com/wireapp/wire-android/pull/1964
#### APK
[Download build #12292](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12292/artifact/build/artifact/wire-dev-PR1967-12292.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/preferences/pages/DataUsagePermissionsView.scala
+++ b/app/src/main/scala/com/waz/zclient/preferences/pages/DataUsagePermissionsView.scala
@@ -23,6 +23,7 @@ import android.os.Bundle
 import android.util.AttributeSet
 import android.view.View
 import android.widget.LinearLayout
+import com.waz.ZLog
 import com.waz.ZLog.ImplicitTag._
 import com.waz.service.AccountManager
 import com.waz.threading.Threading
@@ -33,6 +34,7 @@ import com.waz.zclient.tracking.GlobalTrackingController.analyticsPrefKey
 import com.waz.zclient.tracking.GlobalTrackingController
 import com.waz.zclient.utils.{BackStackKey, ContextUtils}
 import com.waz.zclient.{R, ViewHelper}
+import com.waz.zclient.BuildConfig
 
 class DataUsagePermissionsView(context: Context, attrs: AttributeSet, style: Int) extends LinearLayout(context, attrs, style) with ViewHelper {
   def this(context: Context, attrs: AttributeSet) = this(context, attrs, 0)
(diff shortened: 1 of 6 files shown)


app/src/main/scala/com/waz/zclient/preferences/pages/DataUsagePermissionsView.scala
```diff
@@ -46,42 +48,46 @@ class DataUsagePermissionsView(context: Context, attrs: AttributeSet, style: Int
...
     }
   }
 
   val receiveNewsAndOffersSwitch = returning(findById[SwitchPreference](R.id.preferences_receive_news_and_offers)) { v =>
 
-    def setReceivingNewsAndOffersSwitchEnabled(enabled: Boolean) = {
-      v.setEnabled(enabled)
-      v.setAlpha(if (enabled) 1.0f else 0.5f)
-    }
+    if (BuildConfig.ALLOW_MARKETING_COMMUNICATION) {
```
marcoconti83: Here I just added an `if`. This causes all indentation to change so all following lines look changed, but I did not modify the behavior, just indented.

app/src/main/scala/com/waz/zclient/preferences/pages/DataUsagePermissionsView.scala
```diff
@@ -46,42 +48,46 @@ class DataUsagePermissionsView(context: Context, attrs: AttributeSet, style: Int
...
-    def setAnalyticsSwitchEnabled(enabled: Boolean) = {
-      v.setEnabled(enabled)
-      v.setAlpha(if (enabled) 1.0f else 0.5f)
-    }
-
-    v.setPreference(analyticsPrefKey, global = true)
-    v.pref.flatMap(_.signal).onChanged {
-      case true  => tracking.optIn()
-      case false => tracking.optOut()
+    if (BuildConfig.SUBMIT_CRASH_REPORTS) {
```
marcoconti83: The `setAnalyticsSwitchEnabled` was not previously used at all so I removed it

app/src/main/scala/com/waz/zclient/preferences/pages/DataUsagePermissionsView.scala
```diff
@@ -46,42 +48,46 @@ class DataUsagePermissionsView(context: Context, attrs: AttributeSet, style: Int
...
-              ContextUtils.showGenericErrorDialog()
+      setReceivingNewsAndOffersSwitchEnabled(false)
+      am.head.flatMap(_.hasMarketingConsent).foreach { isSet =>
+        v.setChecked(isSet, disableListener = true)
+        setReceivingNewsAndOffersSwitchEnabled(true)
+
+        v.onCheckedChange.onUi { set =>
+          setReceivingNewsAndOffersSwitchEnabled(false)
+          am.head.flatMap(_.setMarketingConsent(Some(set))).foreach { resp =>
+            resp match {
```
yeryomenkom: Do not use pattern matching if you can avoid it.
In this case do 
if (resp.isLeft)

app/src/main/scala/com/waz/zclient/preferences/pages/DataUsagePermissionsView.scala
```diff
@@ -23,6 +23,7 @@ import android.os.Bundle
 import android.util.AttributeSet
 import android.view.View
 import android.widget.LinearLayout
+import com.waz.ZLog
```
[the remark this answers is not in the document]
marcoconti83: Fixed

app/src/main/scala/com/waz/zclient/preferences/pages/DataUsagePermissionsView.scala
marcoconti83: Fixed

PR #1978, opened 2019-02-01
Use base 64 string for certificate

marcoconti83:

I found the previous representation of the byte array for certificate a bit hacky. It was a collection of strings each representing a series of octet, each encoded as an hex number. 

Since this is how it was expressed, the code to decode it had to apply all those operations to get back to the real byte array.

I replaced it with a base 64 encoding, which we have out of the box.

# Changes

Re-encoded the original byte array value as base64, and decode it accordingly.

# Testing

Manually tested on device: network requests are still working.

#### APK
[Download build #12317](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12317/artifact/build/artifact/wire-dev-PR1978-12317.apk)
[Download build #12320](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12320/artifact/build/artifact/wire-dev-PR1978-12320.apk)


marcoconti83:
@pwnsdx have a look just in case.

@all I'm also open to discuss if this is better or not, if you think the previous representation was better, I'm fine sticking to the original one.

pwnsdx:
LGTM 👍

PR #1990, opened 2019-02-11
Readonly profiles

JPonte:
This adds a method to lock the properties on the account page that the user can't edit if their profile is readonly. 

The flag isn't implemented on SE yet so right now the readonly value is mocked.
#### APK
[Download build #12343](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12343/artifact/build/artifact/wire-dev-PR1990-12343.apk)
[Download build #12344](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12344/artifact/build/artifact/wire-dev-PR1990-12344.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/preferences/pages/AccountView.scala
+++ b/app/src/main/scala/com/waz/zclient/preferences/pages/AccountView.scala
@@ -192,7 +192,7 @@ class AccountViewController(view: AccountView)(implicit inj: Injector, ec: Event
   val selfPicture: Signal[ImageSource] = self.map(_.picture).collect{case Some(pic) => WireImage(pic)}
 
   //TODO: Replace with flag coming from self
-  private val accountIsLocked = Signal.const(true)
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/preferences/pages/AccountView.scala
```diff
@@ -181,6 +191,13 @@ class AccountViewController(view: AccountView)(implicit inj: Injector, ec: Event
 
   val selfPicture: Signal[ImageSource] = self.map(_.picture).collect{case Some(pic) => WireImage(pic)}
 
+  //TODO: Replace with flag coming from self
+  private val accountIsLocked = Signal.const(true)
```
marcoconti83: I would default to `false` for now, or all dev/internal builds will have this locked

marcoconti83:
Merging this one, so that @CalumMcCall can work with integrating this feature with SE.

PR #1991, opened 2019-02-11
Add flag to control logging on custom builds

marcoconti83:

On custom builds, logs were disabled. We want to enable them to debug.

# Changes
- Added a new flag to force logging to be enabled, regardless of the application ID. Since the application ID is personalized for custom applications, we can't hardcode any application ID to test against
- Removed this line as it was unused: https://github.com/wireapp/wire-android/blob/ec206db070964ba7e7f9cba95c1921c0e59abd48/app/src/main/scala/com/waz/zclient/utils/package.scala#L274

# Testing
Will be tested by @jspittka on a custom build.

#### APK
[Download build #12345](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12345/artifact/build/artifact/wire-dev-PR1991-12345.apk)
[Download build #12347](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12347/artifact/build/artifact/wire-dev-PR1991-12347.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/utils/package.scala
+++ b/app/src/main/scala/com/waz/zclient/utils/package.scala
@@ -271,8 +271,8 @@ package object utils {
   val InternalId     = "com.wire.internal"
   val ExperimentalId = "com.wire.x"
 
-  val IsProd: Boolean = Set(ProdId, CandidateId).exists(BuildConfig.APPLICATION_ID.contains)
-  val SafeLoggingEnabled: Boolean = Set(ProdId, CandidateId, InternalId).exists(BuildConfig.APPLICATION_ID.contains)
+  val SafeLoggingEnabled: Boolean = BuildConfig.FORCE_ENABLE_LOGGING ||
+    Set(ProdId, CandidateId, InternalId).contains(BuildConfig.APPLICATION_ID)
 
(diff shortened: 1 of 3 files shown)


app/src/main/scala/com/waz/zclient/utils/package.scala
```diff
@@ -271,8 +271,8 @@ package object utils {
   val InternalId     = "com.wire.internal"
   val ExperimentalId = "com.wire.x"
 
-  val IsProd: Boolean = Set(ProdId, CandidateId).exists(BuildConfig.APPLICATION_ID.contains)
-  val SafeLoggingEnabled: Boolean = Set(ProdId, CandidateId, InternalId).exists(BuildConfig.APPLICATION_ID.contains)
+  val SafeLoggingEnabled: Boolean = BuildConfig.FORCE_ENABLE_LOGGING ||
```
JPonte: Wouldn't it better to then add the correct`FORCE_ENABLE_LOGGING` to all the flavours in the gradle to avoid having to have the package ids hardcoded in this file?

marcoconti83: Good point, will do

marcoconti83:
build again

PR #1994, opened 2019-02-12
Re-enable logging and fix logging crash

marcoconti83:

Logs were disabled on all build but devs because logs would crash due to `java.lang.Object.toString` not being available.

# Changes

This commit reverts e5969077a4c251437f35eb0e11bb9c62c158e56c, effectivelly re-applying 3012f1e672600b187aea89de0e2103a26cc86c91.

The real fix is to make sure that the `toString` method is not stripped for any object by proguard.

# Testing

I build a experimental release build locally before and after the fix. Before the fix, I would see the exception in the logs at every start. After the fix, the exception is gone.
#### APK
[Download build #12353](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12353/artifact/build/artifact/wire-dev-PR1994-12353.apk)
[Download build #12354](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12354/artifact/build/artifact/wire-dev-PR1994-12354.apk)

### DIFF
--- a/app/proguard-rules.txt
+++ b/app/proguard-rules.txt
@@ -21,8 +21,9 @@
    *** get*();
 }
 
--keep class scala.collection.SeqLike { public java.lang.String toString(); }
+# Needed for logging
 -keepclassmembers class * { public java.lang.String toString(); }
+
 -keep class scala.concurrent.forkjoin.** { *; }
 -keep class org.webrtc.** { *; }
 -keep class com.waz.call.FlowManager { *; }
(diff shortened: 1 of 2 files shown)


app/proguard-rules.txt
```diff
@@ -22,6 +22,7 @@
 }
 
 -keep class scala.collection.SeqLike { public java.lang.String toString(); }
```
yeryomenkom: Maybe we now can remove this rule? Since now we keep toString unchanged for all classes?

marcoconti83: Good point

marcoconti83:
Some issues found in manual testing. Please don't merge yet.

marcoconti83:
Re-tested manually after last commit, it works. The previous report of issues was due to another build running in parallel on the same device and logging the exception.

PR #2000, opened 2019-02-15
Fix CreateConversationSettings UI elements not reacting to events

marcoconti83:
Many UI elements in the "create conversation" screen were not responsive to user event after switching to another fragment and coming back.

In particular, the "show/hide" arrow for settings would be visible only the first time the fragment is opened, and the list of users would not update after switching guests on/off if going back and forth between fragments.

# Changes

This was caused by calling `viewHolder.foreach` in the body of a `returning(... viewholder ...)`. With this code, the foreach would be called only once when the fragment is created. If the view is then destroyed and re-created, the new view inside the viewholder would not have been "visited" by the foreach.

The fix is to call the `viewholder.foreach` in the `onCreateView` method, which is called again every time the view is re-created (and the new individual views are already available in the view holder).

# Testing
Tested manually on device, before and after the fix. Steps:
- log in as team user
- open create conversation flow
- enter conversation name
- leave option as: allow guests
- click next (go to user selections)
- click back to previous screen
- issue 1: you can't expand/collapse the options anymore
- switch off "allow guest"
- click next
- issue 2: guests are still in the list

Both issues are fixed with this PR

#### APK
[Download build #12361](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12361/artifact/build/artifact/wire-dev-PR2000-12361.apk)
[Download build #12365](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12365/artifact/build/artifact/wire-dev-PR2000-12365.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/conversation/creation/CreateConversationSettingsFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/conversation/creation/CreateConversationSettingsFragment.scala
@@ -42,14 +42,10 @@ class CreateConversationSettingsFragment extends Fragment with FragmentHelper {
   private lazy val userAccountsController       = inject[UserAccountsController]
 
   private lazy val inputBox = view[InputBox](R.id.input_box)
-
-  private lazy val guestsToggle = returning(view[SwitchCompat](R.id.guest_toggle)) { vh =>
-    findById[ImageView](R.id.allow_guests_icon).setImageDrawable(GuestIconWithColor(getStyledColor(R.attr.wirePrimaryTextColor)))
-    createConversationController.teamOnly.currentValue.foreach(teamOnly => vh.foreach(_.setChecked(!teamOnly)))
-    vh.foreach(_.setOnCheckedChangeListener(new OnCheckedChangeListener {
-      override def onCheckedChanged(buttonView: CompoundButton, isChecked: Boolean): Unit = createConversationController.teamOnly ! !isChecked
-    }))
-  }
+  private lazy val guestsToggle = view[SwitchCompat](R.id.guest_toggle)
+  private lazy val convOptions = view[View](R.id.create_conv_options)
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/conversation/creation/CreateConversationSettingsFragment.scala
```diff
@@ -114,7 +95,9 @@ class CreateConversationSettingsFragment extends Fragment with FragmentHelper {
 
   override def onViewCreated(v: View, savedInstanceState: Bundle): Unit = {
 
-    callInfo
+    callInfo.foreach { view =>
+        userAccountsController.isTeam.onUi(vis => view.setVisible(vis))
```
JPonte: This should be either in the returning or `onCreate`.

app/src/main/scala/com/waz/zclient/conversation/creation/CreateConversationSettingsFragment.scala
```diff
@@ -128,11 +111,27 @@ class CreateConversationSettingsFragment extends Fragment with FragmentHelper {
...
+      view.setOnCheckedChangeListener(new OnCheckedChangeListener {
+        override def onCheckedChanged(buttonView: CompoundButton, isChecked: Boolean): Unit = createConversationController.teamOnly ! !isChecked
+      })
+    }
     readReceiptsToggle
 
-    convOptions
-    convOptionsArrow
+    convOptions.foreach { view =>
+      userAccountsController.isTeam.onUi(vis => view.setVisible(vis))
```
JPonte: Same here.

marcoconti83:
@JPonte pointed out in person that this works for `viewholder.foreach` **only when** the `foreach` is not nested in a signal. If a signal is involved, then that block should either be in the `returning`, or on `onCreate`.

I will update my PR.

PR #2010, opened 2019-02-20
skeleton of HeaderFooterAdapter

yeryomenkom:


app/src/main/scala/com/waz/zclient/views/HeaderFooterAdapter.java
```diff
@@ -0,0 +1,83 @@
...
+
+    protected View header;
+    protected View footer;
+
+    public HeaderFooterAdapter(RecyclerView.Adapter<T> wrapped) {
+        this.wrapped = wrapped;
+    }
+
+    public void setHeader(View view) {
+        header = view;
```
JPonte: This should trigger data set changed.

app/src/main/scala/com/waz/zclient/views/HeaderFooterAdapter.java
```diff
@@ -0,0 +1,83 @@
...
+    }
+
+    public void setFooter(View view) {
+        footer = view;
```
JPonte: So should this.

app/src/main/scala/com/waz/zclient/views/HeaderFooterAdapter.java
yeryomenkom: I am not sure. I would rather give a choice for user of this adapter.

JPonte: But it might lead to the user forgetting to call `notifyItemInserted(0)` and the data will be inconsistent.

JPonte: Or `removed` or `changed`. Because the user of the adapter might want to change or add different headers at some point but has no way of accessing the state.

yeryomenkom: Makes sense.

JPonte:
Jenkins has a style check for Java files. Either fix them or deactivate it.

PR #2014, opened 2019-02-21
Make backend picker listen to the preference

JPonte:
#### APK
[Download build #12387](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12387/artifact/build/artifact/wire-dev-PR2014-12387.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/utils/BackendPicker.java
+++ b/app/src/main/java/com/waz/zclient/utils/BackendPicker.java
@@ -58,7 +58,7 @@ public class BackendPicker {
         }
     }
 
-    private void showDialog(BackendConfig prod, Activity activity, final Callback<BackendConfig> callback) {
+    private void showDialog(final BackendConfig prod, Activity activity, final Callback<BackendConfig> callback) {
         final String[] backends = new String[] {
             Backend.StagingBackend().environment(),
             prod.environment()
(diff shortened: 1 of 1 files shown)


app/src/main/java/com/waz/zclient/utils/BackendPicker.java
```diff
@@ -75,14 +75,35 @@ public void onClick(DialogInterface dialog, int which) {
...
+        dialog.show();
+
+        SharedPreferences.OnSharedPreferenceChangeListener listener = new SharedPreferences.OnSharedPreferenceChangeListener() {
+            @Override
+            public void onSharedPreferenceChanged(SharedPreferences sharedPreferences, String key) {
+                if (key.equals(CUSTOM_BACKEND_PREFERENCE)) {
+                    BackendConfig be = getBackendConfig(prod);
+                    if (be != null) {
+                        callback.callback(be);
+                        prefs().unregisterOnSharedPreferenceChangeListener(this);
```
marcoconti83: Does this line cause the app to stop listening after one change? If yes, was this a QA requirements? Just curious why we do it.

JPonte: Unregistering the listeners is usually a common practice.
The `DefaultSharedPreferences` is a global service from android so it will outlive this component. Leaving a listener in the service could lead to a memory leak or null pointer.

PR #2020, opened 2019-02-27
fix: A partner should be treated as a guest when displaying the bottom bar

makingthematrix:
Fixes a bug when the user (who is a partner, not a guest) opens group conversations' details and from there goes to a single participant details. The only possible action for a partner on this screen should be "Open Conversation" which appers on the left button. The right button ("...") should be invisible.
Because of the bug it was visible, but when clicked, no options appeared.
#### APK
[Download build #12407](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12407/artifact/build/artifact/wire-dev-PR2020-12407.apk)


marcoconti83:
build again

PR #2021, opened 2019-02-27
fix: on start conversation action menu was empty

CalumMcCall:
- This was due to currentConv being an empty signal, so this commit
   changes things to use currentConvOpt instead. A long term fix is to
   use a separate Controller for the conversationListManager and the
   conversation details tab.
#### APK
[Download build #12409](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12409/artifact/build/artifact/wire-dev-PR2021-12409.apk)
[Download build #12411](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12411/artifact/build/artifact/wire-dev-PR2021-12411.apk)
[Download build #12412](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12412/artifact/build/artifact/wire-dev-PR2021-12412.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/participants/ConversationOptionsMenuController.scala
+++ b/app/src/main/scala/com/waz/zclient/participants/ConversationOptionsMenuController.scala
@@ -92,6 +92,7 @@ class ConversationOptionsMenuController(convId: ConvId, mode: Mode)(implicit inj
     isBot               <- otherUser.map(_.exists(_.isWireBot))
     removePerm          <- inject[UserAccountsController].hasRemoveConversationMemberPermission(convId)
     isGuest             <- participantsController.isCurrentUserGuest
+    currentConv         <- convController.currentConvOpt
     selectedParticipant <- participantsController.selectedParticipant
   } yield {
     import com.waz.api.User.ConnectionStatus._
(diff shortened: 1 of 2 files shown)


CalumMcCall:
build again

CalumMcCall:
> I guess the controller and the footer menu should be rewritten, preferably in Kotlin, as our first big task of rewriting parts of UI in this language.
> Until then, I'd suggest to simply take advantage of the fact that we know if we're opening the menu from the conversation list or not. If yes, then `isGuest` and `selectedParticipant` fields in calculating `optionItems` don't make any sense. We don't have to wait for anything coming from these two signals (`participantsController.isCurrentUserGuest` and `participantsController.selectedParticipant`). Instead we can set them to default values, like this:
> 
> ```
>     isGuest             <- if (!mode.inConversationList) participantsController.isCurrentUserGuest else Signal.const(false)
>     selectedParticipant <- if (!mode.inConversationList) participantsController.selectedParticipant else Signal.const(None)
> ```
> 
> and later, in the part which is used to add `RemoveMember` to the menu, we can specify that this should work if and only if we're opening the menu from the profile.
> So, instead of:
> 
> ```
>       case Mode.Normal(_) if isGroup && selectedParticipant.isDefined =>
> ```
> 
> we can simply write:
> 
> ```
>       case Mode.Normal(false) if isGroup && selectedParticipant.isDefined =>
> ```
> 
> and that should achieve this effect.
> 
> This way the `isCurrentUserGuest` signal definition can stay as it is. It's perfectly fine the way it is - it just shouldn't be used if we don't have a current user.

Thanks for the suggestion! I didn't realize mode contained that useful info! :)

PR #2022, opened 2019-02-27
fix: Ask the backend for the rich info if the user is from our team

makingthematrix:
#### APK
[Download build #12410](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12410/artifact/build/artifact/wire-dev-PR2022-12410.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/participants/fragments/SingleParticipantFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/participants/fragments/SingleParticipantFragment.scala
@@ -26,6 +26,7 @@ import android.support.v7.widget.{LinearLayoutManager, RecyclerView}
 import android.view.{LayoutInflater, View, ViewGroup}
 import android.widget.TextView
 import com.waz.ZLog.ImplicitTag._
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/participants/fragments/SingleParticipantFragment.scala
```diff
@@ -203,8 +204,9 @@ class SingleParticipantFragment extends FragmentHelper {
       (for {
         zms                <- inject[Signal[ZMessaging]].head
         user               <- participantsController.otherParticipant.head
-        _                  <- zms.users.syncIfNeeded(Set(user.id))
         isGuest            = !user.isWireBot && user.isGuest(zms.teamId)
+                             // if the user is from our team we ask the backend for the rich profile (but we don't wait for it)
```
CalumMcCall: Your comment doesn't seem to match the check below exactly. We check if we are in a team, not whether the other user is in the same team as us. Does that matter? Maybe it's worth rewording this comment? Or maybe I've just misunderstood :P

PR #2028, opened 2019-03-04
fix: Restore the Block button in the bottom menu in the conv list

makingthematrix:
### Issues
After the recent changes the "conversation actions" menu stopped displaying the Block button when opened by a long tap from the conversation list. The same menu opened from the participant's details displays the Block button correctly.

### Causes
The buttons displayed in the menu are controlled by a multitude of conditions. It's very easy to introduce a minor bug like that when we change something in the code. 

### Solutions

A condition was introduced by mistake that prevented the Block button form appearing. I removed it.

### Testing

Go to the conversation list. Tap long on a 1:1 conversation. The bottom menu will open. The Block button should be among the options.

#### APK
[Download build #12424](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12424/artifact/build/artifact/wire-dev-PR2028-12424.apk)


makingthematrix:
The signal was blocked on `isGuest` and `currentConv` in the for/yield above. It seems that removing this condition at this point, inside the yield part, doesn't block the flow.

CalumMcCall:
You're right, the bug doesn't happen. My bad!

PR #2038, opened 2019-03-13
Log migration: phase 1

johnxnguyen:
## What's new in this PR?

This PR completes the first phase of public logging in UI:

- Remove all usage of `ZLog` (implicit tags, log methods)
- Create `LogUI` object, which is allows access to safe logging methods
- Define `LogShow` instances for UI defined types
- Derive `LogTag` for all types that need it

### Notes

Phase 2 will involve using `LogsService` to enable/disable logs (via build configuration / user preferences). Until then, internal logging will be **disabled**.

Some other things:

- Don't be afraid of the huge commit history, I just made a commit for changes per file.
- `LogTag` used to be a type alias for `String`, now it is it's own type. The old `LogTag` was being used as tags for fragments, which was implicitly derived using macros (in SE). I've replaced these derived tags with the simple class name of the specific fragment type.

### Depends On
- [x] https://github.com/wireapp/wire-android-sync-engine/pull/505

### To Do
- [ ] Unpin SE once it's merged
#### APK
[Download build #12445](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12445/artifact/build/artifact/wire-dev-PR2038-12445.apk)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -22,10 +22,10 @@ ext {
     playServicesVersion = '15.0.1'
     audioVersion = System.getenv("AUDIO_VERSION") ?: '1.209.0@aar'
     stethoVersion = '1.5.0'
-    zMessagingDevVersion = "141.0.0-DEV"
+    zMessagingDevVersion = "141.0.0-SNAPSHOT"
     paging_version = "1.0.0"
     // Release version number must be like this X.0(.Y)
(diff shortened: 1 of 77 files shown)


johnxnguyen:
This PR should definitely be **SQUASHED**.

johnxnguyen:
@marcoconti83 can you please check that I bumped SE correctly?

marcoconti83:
> @marcoconti83 can you please check that I bumped SE correctly?

Yep, looks OK!

PR #2046, opened 2019-03-15
Fixes to reopening the SSO Login page and dialog

makingthematrix:
### fix: Reopening the SSO Login from an SSO link should reset the page

Until now bringing the app from the background tothe front simply displayed the last screen the user was on. With the introduction of deep links this will have to change. Just as with opening the app from a notification, the app will have to figure out what screen should be displayed.
If the user clicks on an SSO link while the app is in the background and already on the `SSOWebViewFragment`, the fragment should not just be displayed, but it should be reloaded.

fixes https://wearezeta.atlassian.net/browse/AN-6056

### fix: The SSO dialogs should not pile on each other

If the user went back and forth between the Welcome Page and subsequent login/registration pages,
the SSO dialog was sometimes created twice for every page, but when the user clicked `cancel`,
only one was really dismissed. The other didn't appear, but it was also not gone: when the user
went back and forth again and saw the dialog again, now two clicks on `cancel` were required
to dismiss it. Then three, then four, etc.
I fixed the creation of the second dialog, and also changed the way listeners are added and
removed from the dialog (they should be handled in `onStart` and `onStop`).

fixes https://wearezeta.atlassian.net/browse/AN-6086

#### APK
[Download build #12449](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12449/artifact/build/artifact/wire-dev-PR2046-12449.apk)
[Download build #12451](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12451/artifact/build/artifact/wire-dev-PR2046-12451.apk)
[Download build #12454](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12454/artifact/build/artifact/wire-dev-PR2046-12454.apk)
[Download build #12456](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12456/artifact/build/artifact/wire-dev-PR2046-12456.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/InputDialog.scala
+++ b/app/src/main/scala/com/waz/zclient/InputDialog.scala
@@ -23,9 +23,11 @@ import android.os.Bundle
 import android.support.annotation.StringRes
 import android.support.v4.app.DialogFragment
 import android.support.v7.app.AlertDialog
-import android.view.View
+import android.text.TextWatcher
 import android.view.View.OnAttachStateChangeListener
+import android.view.{LayoutInflater, View}
 import android.widget.{EditText, TextView}
+import com.waz.utils.returning
 import com.waz.zclient.utils.RichTextView
 
 object InputDialog {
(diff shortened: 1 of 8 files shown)


marcoconti83:
@makingthematrix there is a conflict now :(

PR #2048, opened 2019-03-18
feat: track latency of notifications in new table

CalumMcCall:
### Issues

Add first table for notification tracking. This is the first stage, where we store all notifications received over FCM into a table.
#### APK
[Download build #12455](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12455/artifact/build/artifact/wire-dev-PR2048-12455.apk)
[Download build #12458](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12458/artifact/build/artifact/wire-dev-PR2048-12458.apk)
[Download build #12461](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12461/artifact/build/artifact/wire-dev-PR2048-12461.apk)

### DIFF
--- a/app/src/main/scala/com/waz/services/fcm/FCMHandlerService.scala
+++ b/app/src/main/scala/com/waz/services/fcm/FCMHandlerService.scala
@@ -148,13 +148,14 @@ object FCMHandlerService {
               idle
             )).map(_ => {})
         }
-        _ <- nId.fold(Future.successful({})) { nId =>
-          fcmPushes.insert(
+        _ <- nId match {
+          case Some(n) => fcmPushes.insert(
             FCMNotification(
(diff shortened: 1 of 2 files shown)


app/src/main/scala/com/waz/services/fcm/FCMHandlerService.scala
```diff
@@ -147,6 +148,14 @@ object FCMHandlerService {
               idle
             )).map(_ => {})
         }
+        _ <- nId.fold(Future.successful({})) { nId =>
```
yeryomenkom: In such cases I would prefer to use explicit pattern matching on `nId`.
And do not forget to use `()` instead of `{}`

PR #2052, opened 2019-03-19
Addresses Issue #1946 to enable copying sent ephemeral messages

### LINKED ISSUE #1946: Unsent timed message cannot be edited or copied to clipboard
## The problem

A timed message that remains unsent to due to some error cannot be edited or copied to clipboard.

## Environment

This appears to be a version-independent feature by design, as discussed in Issue #1827 (Timed messages cannot be edited – but deleted only). However, I experienced this on stock Android 9.0 on a Moto X4 with Google Fi service in a recent version of Wired about two weeks ago, probably Wire 3.24.684.

## Details

Although inability to edit a timed message is implemented by design, an unsent message is not yet a timed message because there is not yet any recipient.

There are two reasons why editing and copy to clipboard should be permitted for an unsent message, and the current restrictions may be considered a bug or a misfeature.

First, if an unsent message is being manually resent after a delay, editing might be needed for accuracy if conditions have changed.

Second, if Wire is repeatedly failing, and the original message is long, the user should be permitted to copy-and-paste into an alternative messaging application, instead of being frustrated by being forced to retype the entire message into a different app.

If copy to clipboard of the unsent message is permitted, then editing is not essential. The user can merely copy-and-paste the unsent message into a new one, edit that before sending, then delete the unsent one. So I recommend simply allowing copy to clipboard for now (should be an easy change in the code somewhere), then addressing the editing question later as resources permit.

harishvishwakarma:
## What's new in this PR?
Ability to copy sent ephemeral messages
### Issues

Earlier user wasn't able to copy sent ephemeral messages to clipboard. This PR enables that.  Fixes #1946 

### Solutions

The solution adds a conditional check to see if the message is a sent ephemeral message and then allows copying.

### Testing

The app was run on Emulator(Pixel 2), S9+ and Xiaomi A2 to test.

### DIFF
--- a/app/src/main/scala/com/waz/zclient/messages/MessageBottomSheetDialog.scala
+++ b/app/src/main/scala/com/waz/zclient/messages/MessageBottomSheetDialog.scala
@@ -118,7 +118,7 @@ object MessageBottomSheetDialog {
     case object Copy extends MessageAction(R.id.message_bottom_menu_item_copy, R.string.glyph__copy, R.string.message_bottom_menu_action_copy) {
       override def enabled(msg: MessageData, zms: ZMessaging, p: Params): Signal[Boolean] =
         msg.msgType match {
-          case TEXT | TEXT_EMOJI_ONLY | RICH_MEDIA if !msg.isEphemeral => Signal.const(true)
+          case TEXT | TEXT_EMOJI_ONLY | RICH_MEDIA if zms.selfUserId == msg.userId || !msg.isEphemeral => Signal.const(true)
           case _ => Signal.const(false)
         }
     }


AndroidBob:

marcoconti83:
Hi @harishvishwakarma ,

The PR is quite simple and it looks great! 

We don't have a proper unit test coverage on the UI layer, so thanks for testing on the phone, for now that's the simpler way to test it.

I will coordinate with our QA department on changing the QA automation to reflect this change, and then we will merge it!

marcoconti83:
Hi @harishvishwakarma ,
I checked in with our QA team and we are ready to adapt our automation to consider this change. I'm merging it in.

Thanks for addressing this issue and contributing!

Issue #1986, opened 2019-02-06
Privacy concern about the last remained trackers


machiav3lli:
So, after spending some time "searxing", testing but not-building, that's actually the best what I could get, as the exodus privacy promotes 3 infamous trackers(firebase analytics, HockeyApp and Mixpanel) for the play store version's apk.
And yeah firebase analytics isn't most known for being privacy friendly, so my question/suggestion would be, is there any plan to release a version of Wire without this(and the other) trackers?! and if so, is the plan to have a f-droid release is still on the prioirty list?
that's me trying not to be redundant and ask about replacing FCM with SSE(or other alternative).
![screenshot_exodus_privacy_20190206-200143](https://user-images.githubusercontent.com/40302595/52370897-6a446f00-2a54-11e9-8ea0-f209bfb99db6.png)
thanks.

marcoconti83:
Hi @machiav3lli ,
thanks for the feedback. We removed Hockey and Mixpanel from our app - there might still be some mention to them in some variable name, but the libraries are not included anymore. 

Can you clarify in which public version you still see them?

Regarding Firebase, we still use Firebase for native pushes, but we don't use the analytics component.

machiav3lli:
thanks for your answer @marcoconti83 ,
the apk which had Hockey and Mixpanel was from ApkMirror, so maybe an old version from your play store's.
Regarding Firebase, may I know what kind of native pushes you mean, cause as I know, the cloud messaging doesn't need it even.
And again thanks for the great job you're doing, trying to satisfy our sense of insecurity on this matter.

marcoconti83:
Hi @machiav3lli ,
if downloading from the Google Play Store is not an option for you, you can always download from [our website](https://wire.com/en/download/).

Regarding Firebase, the app uses [Firebase Cloud Messaging](https://firebase.google.com/docs/cloud-messaging/) to deliver notifications to the phone. It's worth noting that the pushes that go through Firebase do not contain any sensitive information such as messages or contacts. 

The notification just informs the application that there is new data available, then the app will then react by fetching the actual data from the Wire server.

machiav3lli:
Hi again @marcoconti83 ,
The version I had with only firebase analytics is downloaded from the website (hoping to see the app sometime in the near future on Fdroid),
I have couple of apps which use the the google cloud messaging without having firebase implemented as a "tracker"(source: Exodus),
[...]

PR #2055, opened 2019-03-21
fix: remove analytics library from dependencies

marcoconti83:
## What's new in this PR?
Removing some warning about tracking libraries by excluding those libraries completely.

### Issues
Wire tests positive for trackers on the Exodus detection tool/website. See https://github.com/wireapp/wire-android/issues/1986

### Causes
The Firebase library (which we use for FCM) also includes some tracking libraries. Even if we don't use them, they are still included and this will trigger a warning.

### Solutions
I used the suggestion from https://github.com/wireapp/wire-android/issues/1986#issuecomment-465808522 and excluded those libraries manually from the Firebase import package.

### Testing

There are two concerns here:
1. Will this change remove the warning?
1. Will this change affect negatively the behavior of notifications for the user?

For the warning, I manually checked a local build on device with _ClassyShark3xodus_ as suggested in https://github.com/wireapp/wire-android/issues/1986#issuecomment-474808790

_ClassyShark3xodus_ reported trackers before the fix, and no trackers after the fix. ✅

For the notification behavior, this is not so straightforward to test as in general it's hard to test absence of bugs, and notification is a complex system. In agreement with @AnastasiiaWire we decided to let this build run on internal apps for a couple of weeks before releasing, and gather feedback from internal users. We will merge this PR only when the current release is done, then test it internally and if no problem is detected, it will go public with the following release.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -350,7 +350,10 @@ dependencies {
     implementation "com.google.android.gms:play-services-maps:$playServicesVersion"
     implementation "com.google.android.gms:play-services-location:$playServicesVersion"
     implementation "com.google.android.gms:play-services-gcm:$playServicesVersion"
-    implementation 'com.google.firebase:firebase-messaging:17.3.0'
+    implementation ('com.google.firebase:firebase-messaging:17.3.0') {
+        exclude group: 'com.google.firebase', module: 'firebase-analytics'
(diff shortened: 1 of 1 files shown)


marcoconti83:
FYI @raphaelrobert

marcoconti83:
build again

PR #2077, opened 2019-03-29
Add user deep link opening restrictions for partners.

yeryomenkom:
## What's new in this PR?

Adding restrictions for opening user deep links.

#### APK
[Download build #12521](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12521/artifact/build/artifact/wire-dev-PR2077-12521.apk)
[Download build #12525](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12525/artifact/build/artifact/wire-dev-PR2077-12525.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/deeplinks/DeepLinkService.scala
+++ b/app/src/main/scala/com/waz/zclient/deeplinks/DeepLinkService.scala
@@ -18,6 +18,7 @@
 package com.waz.zclient.deeplinks
 
 import android.content.Intent
+import com.waz.content.MembersStorage
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
 import com.waz.service.AccountManager.ClientRegistrationState.Registered
 import com.waz.service.{AccountManager, AccountsService, UserService}
@@ -40,6 +41,7 @@ class DeepLinkService(implicit injector: Injector) extends Injectable with Deriv
   private lazy val account             = inject[Signal[Option[AccountManager]]]
   private lazy val userService         = inject[Signal[UserService]]
   private lazy val convController      = inject[ConversationController]
+  private lazy val membersStorage      = inject[MembersStorage]
 
   def checkDeepLink(intent: Intent): Unit = Option(intent.getDataString).flatMap(DeepLinkParser.parseLink) match {
     case None =>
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/deeplinks/DeepLinkService.scala
```diff
@@ -99,6 +99,11 @@ class DeepLinkService(implicit injector: Injector) extends Injectable with Deriv
                 await { service.getSelfUser.zip(service.findUser(userId)) } match {
                   case (Some(self), Some(other)) if self.id == other.id =>
                     OpenDeepLink(token, UserTokenInfo(connected = false, currentTeamMember = true, self = true))
+                  case (Some(self), Some(other)) if self.isPartner(self.teamId) || other.isPartner(other.teamId) =>
+                    if (other.isConnected || self.createdBy.contains(self.id) || other.createdBy.contains(self.id))
```
marcoconti83: `isConnected` won't work here. You will never be connected if you are in the same team (a connection entity does not exists on the backend).

The condition that we want to check is that you are at least in one conversation together (it can be a group convo, a 1:1, ...)

The condition should be something like (pseudocode):
```
val hasConversation = conversationService.allConversation.find { _.participants.contain(other) } != null
if (hasConversation || self.createdBy.contains(other.id) || other.createdBy.contains(self.id) {
....
}
```

Notice that there is also a typo in the current if. It should say: `self.createdBy.contains(other.id)`, not `self.createdBy.contains(self.id)`.

app/src/main/scala/com/waz/zclient/deeplinks/DeepLinkService.scala
```diff
@@ -99,11 +101,17 @@ class DeepLinkService(implicit injector: Injector) extends Injectable with Deriv
                 await { service.getSelfUser.zip(service.findUser(userId)) } match {
                   case (Some(self), Some(other)) if self.id == other.id =>
                     OpenDeepLink(token, UserTokenInfo(connected = false, currentTeamMember = true, self = true))
-                  case (Some(self), Some(other)) if self.isPartner(self.teamId) || other.isPartner(other.teamId) =>
-                    if (other.isConnected || self.createdBy.contains(self.id) || other.createdBy.contains(self.id))
+                  case (Some(self), Some(other)) if self.isPartner(self.teamId) || other.isPartner(self.teamId) =>
+                    if (other.isInTeam(self.teamId)) {
```
marcoconti83: Thanks for the fix. I think this `if (other.isInTeam(self.teamId))` can go. It doesn't really matter the team, if you are in the same convo then you can see them. 

It will make the code cleaner.

yeryomenkom: The problem with your suggestion:
- if you are a partner and other user is a regular user, we will not open anything.

marcoconti83: That's actually intended. If you are a partner you will not see **anyone** unless you are in the same conversation. I forgot to specify "anyone" when we talked. So, partner will not see guest unless they are in the same conversation. This also covers the case when they are connected. Connected users always have a 1:1 conversation.

yeryomenkom: I see. Will be fixed in https://github.com/wireapp/wire-android/pull/2079

PR #2079, opened 2019-03-29
Fix user deep link opening for unconnected user

yeryomenkom:
## What's new in this PR?

Return pattern match case for not found user inside the DeepLinkService.
#### APK
[Download build #12526](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12526/artifact/build/artifact/wire-dev-PR2079-12526.apk)


marcoconti83:
Can you clarify how this fixes #2077? From what I can see, this PR does not restrict the ability of partners to see guests that they are not in a conversation with.

PR #2113, opened 2019-04-29
Fix/custom backend

johnxnguyen:
### Changes

- A single `BackendController` instance is now bound in `WireApplication`. There was no need to recreate it every time we needed it, now we can inject it.
- The backend selector dialog has been extracted out of `BackendController`. This is a consequence of binding the controller (the context it uses is not able to present dialogs), but I believe it makes for a cleaner controller.
- The external support urls (accounts, teams, website) are now persisted.
- `BrowserController` has been refactored to use these support urls.
- All places where we were opening urls manually is now delegated to `BrowserController`.
- The urls in `strings_no_translate.xml` have been tidied up and adjusted to support replacement of the host. These string resources are not intended to be used directly, but rather through the `BrowserController`.
#### APK
[Download build #12590](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12590/artifact/build/artifact/wire-dev-PR2113-12590.apk)
[Download build #12591](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12591/artifact/build/artifact/wire-dev-PR2113-12591.apk)
[Download build #12592](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12592/artifact/build/artifact/wire-dev-PR2113-12592.apk)

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -518,7 +518,6 @@
     <string name="pref_about_tos_title">Terms of Use</string>
     <string name="pref_about_privacy_policy_title">Privacy Policy</string>
     <string name="pref_about_licenses_title">License information</string>
-    <string name="pref_about_licenses_url">@string/url_third_party_licences</string>
     <string name="pref_about_version_title">Version %s</string>
     <string name="pref_about_copyright_title">\u00A9 Wire Swiss GmbH</string>
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -36,7 +36,7 @@ ext {
     customAvsName = 'avs'
     customAvsGroup = 'com.wearezeta.avs'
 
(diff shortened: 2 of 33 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -518,7 +518,6 @@
     <string name="pref_about_tos_title">Terms of Use</string>
     <string name="pref_about_privacy_policy_title">Privacy Policy</string>
     <string name="pref_about_licenses_title">License information</string>
-    <string name="pref_about_licenses_url">@string/url_third_party_licences</string>
```
marcoconti83: Why is this gone?

johnxnguyen: It just redirects to the no translation strings file. It's the only usage so I removed it and used the one from the no translations file directly.

PR #2114, opened 2019-04-30
fix: test module

johnxnguyen:
## What's new in this PR?

I would like to encourage the team to start adding tests to the UI project and this PR is a first step toward achieving that goal.

### Issues

The test module has been broken for some time. Admittedly, there are hardly any tests in the UI project, however it's highly unlikely this would change in the current state.

### Causes

There was some test classes still using Roboelectric, which was removed as a dependency. Other test classes were depending on objects from the sync engine that are long gone.

### Solutions

Discard the classes that are completely broken and can't be salvaged without a big effort. Fix any remaining tests where possible.
#### APK
[Download build #12593](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12593/artifact/build/artifact/wire-dev-PR2114-12593.apk)


marcoconti83:
👍

PR #2136, opened 2019-05-16
fix: Notifications are ignored if the app was killed before.

makingthematrix:
If a message arrives when the app is killed or in the background, it is first handled by `FCMHandlerService`. The service needs to access `GlobalModule` and `AccountsService` before it can call `PushService` in order to sync the app. (We don't use the received message directly as it might be out of sync - we perform a sync instead). But if the app is killed there is no `GlobalModule` (nor `AccountService`) present and the service silently stops without calling `PushService`.

In this PR the service first ensures that the app is initialized properly. If the app simply works in the background, no initialization is needed and the code will work as before. If the app is killed, we will look for the stored global config and initialize the app according to it. Only if the config is missing, we will not initialize the app and the service won't work. To avoid this, the first launch of the app after installing it should be done through `LaunchActivity`: there we choose the backend and we save our decision in the stored backend config. After that, even if the app is killed, we will have the information about the chosen backend and the service will be able to initialize the app. (It didn't work before because if we had our Prod backend as default, it was not saved in the stored backend config, and the service didn't know that Prod was our backend).
#### APK
[Download build #12620](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12620/artifact/build/artifact/wire-dev-PR2136-12620.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/LaunchActivity.scala
+++ b/app/src/main/scala/com/waz/zclient/LaunchActivity.scala
@@ -45,7 +45,12 @@ class LaunchActivity extends AppCompatActivity with ActivityHelper with DerivedL
     }
 
     if (backendController.shouldShowBackendSelector) showDialog(callback)
-    else callback(backendController.getStoredBackendConfig.getOrElse(Backend.ProdBackend))
+    else backendController.getStoredBackendConfig match {
+      case Some(be) => callback(be)
+      case None =>
+        backendController.setStoredBackendConfig(Backend.ProdBackend)
+        callback(Backend.ProdBackend)
+    }
   }
 
   /// Presents a dialog to select backend.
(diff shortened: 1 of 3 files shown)


app/src/main/scala/com/waz/zclient/LaunchActivity.scala
```diff
@@ -45,7 +45,12 @@ class LaunchActivity extends AppCompatActivity with ActivityHelper with DerivedL
     }
 
     if (backendController.shouldShowBackendSelector) showDialog(callback)
-    else callback(backendController.getStoredBackendConfig.getOrElse(Backend.ProdBackend))
+    else backendController.getStoredBackendConfig match {
+      case Some(be) => callback(be)
+      case None =>
+        backendController.setStoredBackendConfig(Backend.ProdBackend)
+        callback(Backend.ProdBackend)
+    }
```
johnxnguyen: Ahhh..... This is why it didn't work in production. It's satisfying to finally know, good find!

2048-Bit:
When do we think this will make it into the main release. Wire is effectively unusable on Android because of this issue for my entire org.

2048-Bit:
@makingthematrix do you have a dev build you need tested for this issue?  If so, send me the APK. I tried to build the dev branch, but it's failing. :/

PR #2178, opened 2019-06-14
Improvement: zmessaging version string suffix

johnxnguyen:
## What's new in this PR?

A minor improvement to https://github.com/wireapp/wire-android/pull/2170. Instead of just checking for a `-PR` suffix, let's check if there is any suffix, and only append `-DEV` if none exists. This makes it simpler to not only specify PR builds, but also SNAPSHOT builds.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -444,7 +444,7 @@ def getApiKey(String property) {
 }
 
 static def buildSuffixForVersion(String versionString) {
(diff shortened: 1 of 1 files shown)


johnxnguyen:
@marcoconti83 true, but I think it's also nice to have more persistent option when you work over days/sessions

PR #2187, opened 2019-06-18
fix: Chatheads in participants lists display placeholders instead of avatars

makingthematrix:
## What's new in this PR?

This is a fix for a bug in `ChatHeadView` connected to the new assets.

### Issues

With new assets introduced, `ChatHeadView` was rewritten. The old version cleared the displayed image every time it was asked to load user data, because it might happened that the avatar was updated. In the new version, however, it only made the avatar disappear and be replaced by a placeholder.

### Solutions

I removed the line clearing the avatar on `setUserData`.

### Testing

Go to the group participants view. Wait a few seconds. The bug caused the avatars of participants to be replaced with the initial of their usernames. This should not happen now.

## Notes
The new version should handle avatar updates differently, but it doesn't. I will work on that in another PR.
 I also made some slight refactoring: A few placeholders were created for each avatar, even though only one was necessary.

### DIFF
--- a/app/src/main/scala/com/waz/zclient/common/views/ChatHeadView.scala
+++ b/app/src/main/scala/com/waz/zclient/common/views/ChatHeadView.scala
@@ -68,9 +68,9 @@ class ChatHeadView(val context: Context, val attrs: AttributeSet, val defStyleAt
   val attributes: Attributes = parseAttributes(attrs)
 
   private val options = for {
-    z <- zms
+    z         <- zms
     Some(uId) <- userId
-    user <- z.usersStorage.signal(uId)
+    user      <- z.usersStorage.signal(uId)
   } yield optionsForUser(user, z.teamId.exists(user.teamId.contains(_)), attributes)
 
   options.onUi(setInfo)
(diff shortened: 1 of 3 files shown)


app/src/main/scala/com/waz/zclient/common/views/ChatHeadView.scala
```diff
@@ -232,6 +225,7 @@ object ChatHeadView {
     }
 
     override def draw(canvas: Canvas): Unit = {
+      verbose(l"draw placeholder")
```
CalumMcCall: What is the purpose of this new log line?

app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
```diff
@@ -38,7 +39,8 @@ import com.waz.zclient.views.AvailabilityView
 import com.waz.zclient.{R, ViewHelper}
 import org.threeten.bp.Instant
 
-class SingleUserRowView(context: Context, attrs: AttributeSet, style: Int) extends RelativeLayout(context, attrs, style) with ViewHelper with ThemedView {
+class SingleUserRowView(context: Context, attrs: AttributeSet, style: Int)
+  extends RelativeLayout(context, attrs, style) with ViewHelper with ThemedView with DerivedLogTag {
```
CalumMcCall: Why is this class implementing `DerivedLogTag`?

makingthematrix: removed

app/src/main/scala/com/waz/zclient/common/views/ChatHeadView.scala
makingthematrix: removed

PR #2196, opened 2019-06-25
fix: The avatar of the blocked person should be circular

makingthematrix:
## What's new in this PR?

A small change to `BlockedUserProfileFragment`. The initial issue was that the avatar of a blocked person should have the "blocked" icon on top. In the end we decided not to do it - iOS also displays the icon only on small chatheads, but not on the avatar in the profile. But in the meantime we found out the avatar appears as rectangular, not circular. That was changed. Also, now, if we want to display the "blocked" icon, it's just one additional line of code.

### DIFF
--- a/app/src/main/scala/com/waz/zclient/pages/main/connect/BlockedUserProfileFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/pages/main/connect/BlockedUserProfileFragment.scala
@@ -18,10 +18,13 @@
 
 package com.waz.zclient.pages.main.connect
 
+import android.graphics.Bitmap
 import android.os.Bundle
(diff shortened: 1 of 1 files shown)


makingthematrix:
build again

makingthematrix:
buld again

marcoconti83:
I checked the build failure. It fails at the step where it tries reports to this PR that the build succeeded :)

marcoconti83:
(Please ignore my previous message, it was incorrect.)

marcoconti83:
build again

marcoconti83:
I figured out the issue with the build machine, a proper build should complete successfully soon

PR #2206, opened 2019-07-03
fix: Timer and a blinking dot for quick recording

makingthematrix:
## What's new in this PR?

A minor fix for the quick recording feature.

### Issues

The red dot was not blinking while quick recording a message (only on the first recording - it worked for subsequent ones).
The timer next to the red dot was missing.

### Solutions

The animator for the red dot was initialized only *after* the recording. The timer was there, but displayed an empty string if the message was not played (and it was not *played* when it as being recorded).

### Testing

Try to record an audio message using the quick recording feature and look at the left side of the recording panel.

### DIFF
--- a/app/src/main/scala/com/waz/zclient/common/controllers/MediaRecorderController.scala
+++ b/app/src/main/scala/com/waz/zclient/common/controllers/MediaRecorderController.scala
@@ -92,7 +92,7 @@ class MediaRecorderControllerImpl(context: Context) extends MediaRecorderControl
 
   override def cancelRecording(): Unit = stopRecording()
 
-  override def isRecording: Boolean = false //TODO: ???
+  override def isRecording: Boolean = recorder.isDefined
 
   override def getFile: File = file
 }
--- a/app/src/main/scala/com/waz/zclient/conversation/toolbar/AudioMessageRecordingView.scala
+++ b/app/src/main/scala/com/waz/zclient/conversation/toolbar/AudioMessageRecordingView.scala
@@ -76,7 +76,7 @@ class AudioMessageRecordingView (val context: Context, val attrs: AttributeSet, 
 
   private val playbackControls = Signal[PlaybackControls]()
 
-  val actionUpMinY = getDimenPx(R.dimen.audio_message_recording__slide_control__height) - 2 *
+  private val actionUpMinY = getDimenPx(R.dimen.audio_message_recording__slide_control__height) - 2 *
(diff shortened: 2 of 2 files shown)


app/src/main/scala/com/waz/zclient/common/controllers/MediaRecorderController.scala
```diff
@@ -92,7 +92,7 @@ class MediaRecorderControllerImpl(context: Context) extends MediaRecorderControl
 
   override def cancelRecording(): Unit = stopRecording()
 
-  override def isRecording: Boolean = false //TODO: ???
+  override def isRecording: Boolean = startRecordingOffset.isDefined
```
johnxnguyen: Perhaps it would be better to decouple `isRecording` from the presence of `startRecordingOffset` to avoid some potential false positive situations.

I would suggest to check for the existence of the `recorder` instead.

app/src/main/scala/com/waz/zclient/conversation/toolbar/AudioMessageRecordingView.scala
```diff
@@ -278,21 +282,13 @@ class AudioMessageRecordingView (val context: Context, val attrs: AttributeSet,
...
-  private def startRecordingIndicator() = {
-    if (recordingIndicatorDotAnimator.isEmpty) {
-      recordingIndicatorDotAnimator = Some(returning(ObjectAnimator.ofFloat(recordingIndicatorDotView, View.ALPHA, 0f)) { an =>
-        an.setRepeatCount(ValueAnimator.INFINITE)
-        an.setRepeatMode(ValueAnimator.REVERSE)
-        an.setDuration(RecordingIndicatorHiddenInterval)
-        an.setStartDelay(RecordingIndicatorVisibleInterval)
-      })
+  private lazy val recordingIndicator =
+    returning(ObjectAnimator.ofFloat(recordingIndicatorDotView, View.ALPHA, 0f)) { an =>
```
johnxnguyen: This should probably be moved up to the top with all the other properties.

app/src/main/scala/com/waz/zclient/common/controllers/MediaRecorderController.scala
makingthematrix: ok

app/src/main/scala/com/waz/zclient/conversation/toolbar/AudioMessageRecordingView.scala
makingthematrix: ok

PR #2207, opened 2019-07-04
fix: Quick recorder exception

makingthematrix:
Standard recording was rewritten in Kotlin recently. Quick recording is in Scala, and it was failing on some of the devices. This fix makes the quick recording use the same Kotlin code as the standard recording.
It also provides the code which connects Scala and Kotlin for future use.

### DIFF
--- a/app/src/main/scala/com/waz/zclient/common/controllers/MediaRecorderController.scala
+++ b/app/src/main/scala/com/waz/zclient/common/controllers/MediaRecorderController.scala
@@ -19,85 +19,92 @@ package com.waz.zclient.common.controllers
 
 import java.io.File
 
-import android.media.MediaRecorder
 import android.content.Context
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
-import com.waz.utils.returning
-import com.waz.zclient.utils.media.{AudioEncoder, AudioSource, OutputFormat}
+import com.waz.zclient.{Injectable, Injector, KotlinServices}
 import com.waz.zclient.log.LogUI._
+import io.reactivex.android.schedulers.AndroidSchedulers
+import io.reactivex.disposables.Disposable
+import io.reactivex.internal.functions.Functions
+import com.waz.zclient.utils.ScalaToKotlin._
 
+import scala.util.{Failure, Try}
 
 trait MediaRecorderController {
-  def startRecording(): Unit
-  def stopRecording(): Boolean
+  def startRecording(onFinish: File => Unit = _ => {}): Unit
+  def stopRecording(): Try[Unit]
   def cancelRecording(): Unit
   def isRecording: Boolean
 
-  def getFile: File
   def duration: Option[Long]
 }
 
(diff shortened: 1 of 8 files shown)


app/src/main/scala/com/waz/zclient/common/controllers/MediaRecorderController.scala
```diff
@@ -59,36 +58,42 @@ class MediaRecorderControllerImpl(context: Context) extends MediaRecorderControl
...
 
     try {
       rec.prepare()
+      rec.start()
     } catch {
       case e: Throwable =>
-        verbose(l"Failed to prepare recorder ${showString(e.getMessage)}") //TODO: Abort?
+        verbose(l"Failed to prepare or start recorder: ${showString(e.getMessage)}")
+    } finally {
+      recorder = Some(rec)
```
johnxnguyen: Shouldn't we store the recorder in the `try` block after starting the recording?

app/src/main/scala/com/waz/zclient/common/controllers/MediaRecorderController.scala
```diff
@@ -26,11 +26,10 @@ import com.waz.utils.returning
 import com.waz.zclient.utils.media.{AudioEncoder, AudioSource, OutputFormat}
 import com.waz.zclient.log.LogUI._
 
-import scala.util.Try
 
 trait MediaRecorderController {
   def startRecording(): Unit
-  def stopRecording(): Unit
+  def stopRecording(): Boolean
```
johnxnguyen: It would be more expressive if this returned a result / either.

app/src/main/scala/com/waz/zclient/common/controllers/MediaRecorderController.scala
```diff
@@ -19,80 +19,87 @@ package com.waz.zclient.common.controllers
...
+        onFinish(file)
+        recordingDisposable = None
+      })
+        .observeOn(AndroidSchedulers.mainThread())
+        .subscribe(
+          Functions.emptyConsumer(),
+          { t: Throwable =>
+            error(l"Recording failed", t)
+            recordingDisposable = None
+            startRecordingOffset = None
```
johnxnguyen: Shouldn't we clean up the mp4 file here?

app/src/main/scala/com/waz/zclient/common/controllers/MediaRecorderController.scala
```diff
@@ -19,80 +19,87 @@ package com.waz.zclient.common.controllers
...
   }
 
-  override def cancelRecording(): Unit = stopRecording()
-
-  override def isRecording: Boolean = recorder.isDefined
+  override def cancelRecording(): Unit = {
+    verbose(l"cancelRecording")
+    recordingDisposable.foreach(_.dispose())
+    startRecordingOffset = None
+  }
```
johnxnguyen: Shouldn't we also clean up the mp4 file here?

PR #2212, opened 2019-07-10
fix: Don't fail on retrieving asset details

makingthematrix:
This is the UI part of the fix.
SE: https://github.com/wireapp/wire-android-sync-engine/pull/572
fixes: https://wearezeta.atlassian.net/browse/AN-6298

When we send an asset message, first we try to get some details about it. The data in the `AssetDetails` entity depends on the asset's type: it's different for video, audio, image, or an unrecognized file. In the old version, we tried to get the data based on the Mime type which
was provided. When the mime was wrong, we were not able to get the data and the process failed.
Now, the process tries to find out what the mime is by itself. Initially we guess it based on the extension of the filename, but if we fail to retrieve the data for that mime type, we either assume the asset is an unrecognized file, or - in the case of video - we check first if the asset isn't an audio file (MP4 is an extension for both video or audio). Only if we fail to retrieve audio data, we assume it's an
unrecognized file. In all cases, we don't crash the process.

The reason of the original bug was that we use MP4 as a format for audio recordings. It works well when the recording is sent directly, but when sharing to another conversation, the information about the mime type is lost, and on the receiving end we guess (wrongly) that it is video MP4.

### DIFF
--- a/app/src/main/scala/com/waz/zclient/assets2/AssetDetailsServiceImpl.scala
+++ b/app/src/main/scala/com/waz/zclient/assets2/AssetDetailsServiceImpl.scala
@@ -36,7 +36,7 @@ import com.waz.zclient.log.LogUI._
 import org.threeten.bp
 
 import scala.concurrent.duration._
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.ExecutionContext
 import scala.math.round
 import scala.util.{Failure, Success, Try}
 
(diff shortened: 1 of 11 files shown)


johnxnguyen:
> ...when sharing to another conversation, the information about the mime type is lost...

Why does it get lost? Is there a way to ensure it doesn't?

makingthematrix:
> > ...when sharing to another conversation, the information about the mime type is lost...
> 
> Why does it get lost? Is there a way to ensure it doesn't?

We share the asset only as an URI, without the explicit information about the mime. The only hint for the mime type is the extension in the filename.

app/src/main/scala/com/waz/zclient/assets2/AssetDetailsServiceImpl.scala
```diff
@@ -46,67 +46,89 @@ class AssetDetailsServiceImpl(uriHelper: UriHelper)
...
   import AssetDetailsServiceImpl._
 
-  override def extract(mime: Mime, content: PreparedContent): Future[AssetDetails] = {
-    if (Mime.Image.supported.contains(mime)) extractForImage(content)
-    else if (Mime.Video.supported.contains(mime)) extractForVideo(asSource(content))
-    else if (Mime.Audio.supported.contains(mime)) extractForAudio(asSource(content))
-    else Future.successful(BlobDetails)
-  }
+  override def extract(content: PreparedContent): (AssetDetails, Mime) =
+    content.getMime(uriHelper) match {
```
johnxnguyen: There's a fair bit of duplication in this method, could you try to refactor it out?

makingthematrix: ok, done

makingthematrix:
> Unit tests would be nice, at least one asserting the correct mime is extracted for an mp4 audio file .

I fixed existing android unit tests which test `AssetDetailsService`, but they still fail because the cursor in `AndroidUriHelper` is null. I suspect this has something to do with the location of resources. Tests which copy the test files to the external cache work fine.

PR #2213, opened 2019-07-10
PoC to enforce device password policy

CalumMcCall:
### Issues

This PR is a proof-of-concept to implement the requirements in [AN-6292](https://wearezeta.atlassian.net/browse/AN-6292). It uses the Android Device Administration API to require users to set a device/lock screen password of a given strength before they are allowed to login to the Wire app. 

`SecurityPolicyService` is used by the Android system to notify the App that we have been granted device admin privileges. Then we set the password policy, or any other admin policy.

The policy is set in `SecurityPolicyService` and is currently a minimum password length of 12 alphabetic characters.

### Steps to use

This PoC does not support enforcing password policies on Apps with accounts already logged in, although this would be a straightforward addition(probably to `startFirstFragment` in `MainActivity`).

1. Wipe any existing data or log out of Wire
2. Enter email+pass
3. Click login
4. The app will display a dialog and afterwards fire and Intent which will prompt the user to enable Admin privileges for the Wire application.
5. Once enabled, Wire will set and enforce a password policy of a minimum 12 character alphabetic password. Try logging in again, if the pre-existing password doesn't comply with the policy, the user will again be prompted by a dialog to set a compliant password, otherwise, login will continue as normal.

### Further work

 - Override the `onDisabled` method in `SecurityPolicyService` to detect when we lose privileges and log the user out when this happens. This is necessary as the user can remove our admin privileges at any time.
 - Check `isAdminActive` and if we don't have admin privileges then log the user out until they comply with the policy. This is necessary for users who are already logged in.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -211,6 +211,17 @@
             </intent-filter>
         </receiver>
 
+        <receiver android:name="com.waz.services.SecurityPolicyService"
+            android:label="@string/security_policy_title"
+            android:description="@string/security_policy_desc"
+            android:permission="android.permission.BIND_DEVICE_ADMIN">
+            <meta-data android:name="android.app.device_admin"
+                android:resource="@xml/device_admin" />
+            <intent-filter>
+                <action android:name="android.app.action.DEVICE_ADMIN_ENABLED" />
+            </intent-filter>
+        </receiver>
+
         <provider
             android:name="com.waz.content.WireContentProvider"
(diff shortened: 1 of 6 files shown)


johnxnguyen:
This is the password policy of the entire phone and not just for the Wire account, right?

CalumMcCall:
@johnxnguyen Yeah, it's the policy for the password used to protect the lock screen.

marcoconti83:
Closing as obsolete

PR #2223, opened 2019-07-23
feat: Security checklist

makingthematrix:
Initial implementation of a "controller" checking security concerns and preventing the app
from being opened if something seems wrong. Currently lives in Developer Settings.


app/src/main/scala/com/waz/zclient/SecurityCheckList.scala
```diff
@@ -0,0 +1,97 @@
...
+
+  val checkIfRooted: Check = { _ =>
+    Future {
+      lazy val releaseTagsExist = getSystemProperty("ro.build.tags").contains("release-keys")
+      lazy val otacertsExist = new File("/etc/security/otacerts.zip").exists()
+      lazy val canRunSu = runCommand("su")
+
+      verbose(l"checkIfRooted, releaseTagsExist: $releaseTagsExist, otacertsExist: $otacertsExist, canRunSu: $canRunSu")
+
+      releaseTagsExist && otacertsExist && !canRunSu // true means not rooted
```
johnxnguyen: I'd suggest negating the condition so `true` means rooted (then removing the redundant comment). This would align with the name `checkIfRooted`.

makingthematrix: done

app/src/main/scala/com/waz/zclient/security/RootDetectionCheck.scala
```diff
@@ -0,0 +1,78 @@
...
+        .asInstanceOf[String]
+    ).toOption
+
+  private def runCommand(command: String): Boolean =
+    Try {
+      verbose(l"runCommand($command)")
+      Runtime.getRuntime.exec(command)
+    }.isSuccess
+
+  private def noteThatPhoneIsRooted(): Unit = {
```
makingthematrix: `RootDetectedFlag` should be defined in SE in `GlobalPreferences`. Look for example how `GlobalPreferences.CheckedForPlayServices` is used in `WireApplication.checkForPlayServices`.

app/src/main/scala/com/waz/zclient/security/SecurityChecklist.scala
```diff
@@ -0,0 +1,61 @@
...
+import com.waz.threading.Threading.Implicits.Background
+import com.waz.zclient.security.SecurityChecklist._
+import com.waz.zclient.log.LogUI._
+
+import scala.concurrent.Future
+
+class SecurityChecklist(list: List[(Check, List[Action])]) extends DerivedLogTag {
+
+  def run(): Future[Boolean] = {
+    if (list.nonEmpty) info(l"Running security checks")
```
makingthematrix: If we check `list.nonEmpty` here, we may as well not call `runChecks` at all on if it's empty.

app/src/main/scala/com/waz/zclient/security/PreviouslyRootedCheck.scala
```diff
@@ -0,0 +1,38 @@
...
+ */
+package com.waz.zclient.security
+
+import android.content.Context
+import android.preference.PreferenceManager
+import com.waz.zclient.security.RootDetectionCheck.RootDetectedFlag
+
+import scala.concurrent.Future
+
+class PreviouslyRootedCheck(implicit context: Context) extends SecurityChecklist.Check {
```
makingthematrix: It can be merged with `RootDetectionCheck`

johnxnguyen: Done

app/src/main/scala/com/waz/zclient/security/SecurityChecklist.scala
johnxnguyen: Done

app/src/main/scala/com/waz/zclient/security/RootDetectionCheck.scala
johnxnguyen: Done

Madis0:
Why does Wire check for root? Is this feature added for all users or only for teams?

marcoconti83:
Hi @Madis0 ,
This will check will only be enabled for specific customers that ask for this feature in a custom build. The rest of the users will not be affected by this and can still use Wire on rooted phones.

PR #2246, opened 2019-08-08
feat: add button content description for tests

CalumMcCall:
### Issues

Addresses [AN-6313](https://wearezeta.atlassian.net/browse/AN-6313). This PR adds a text description that can be read by automated tools. Determining the action button's state was otherwise unreliable, and was hampering testing.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1146,6 +1146,12 @@
     <string name="app_entry_sso_dialog_message">Enter the company SSO access code.</string>
     <string name="app_entry_sso_input_hint"></string>
 
+    <string name="action_button_state_play">play</string>
+    <string name="action_button_state_pause">pause</string>
+    <string name="action_button_state_close">close</string>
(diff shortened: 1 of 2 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1146,6 +1146,12 @@
     <string name="app_entry_sso_dialog_message">Enter the company SSO access code.</string>
     <string name="app_entry_sso_input_hint"></string>
 
+    <string name="action_button_state_play">play</string>
+    <string name="action_button_state_pause">pause</string>
+    <string name="action_button_state_close">close</string>
+    <string name="action_button_state_redo">redo</string>
+    <string name="action_button_state_none">redo</string>
```
makingthematrix: Shouldn't it be "none", not "redo"?

marcoconti83:
```suggestion
    <string name="action_button_state_none">none</string>
```

PR #2259, opened 2019-08-20
Feature: app lock

johnxnguyen:
## What's new in this PR?

This PR includes changes relating to the App Lock. This is a feature where, in order to open the app, the user must authenticate using the device screen lock mechanism (pin, pattern, password etc). The user will only be prompted to unlock the app again if the app has been in the background for more than 10 seconds. 

This feature can be enabled in the options screen in the app settings, and it applies to the whole app (not for specific accounts).

For custom white label versions, this feature can be forced, and the timeout duration can be specified.

### DIFF
--- a/app/src/main/scala/com/waz/zclient/security/AppLockActivity.scala
+++ b/app/src/main/scala/com/waz/zclient/security/AppLockActivity.scala
@@ -76,7 +76,7 @@ class AppLockActivity extends AppCompatActivity with DerivedLogTag {
 
 object AppLockActivity extends DerivedLogTag {
 
-  val ConfirmDeviceCredentialsRequestCode = 1
+  final val ConfirmDeviceCredentialsRequestCode = 1
 
   private var timeEnteredBackground: Option[Instant] = Some(Instant.EPOCH)
 
--- a/app/src/main/scala/com/waz/zclient/security/AppLockActivity.scala
+++ b/app/src/main/scala/com/waz/zclient/security/AppLockActivity.scala
@@ -1,0 +1,87 @@
+/**
+ * Wire
+ * Copyright (C) 2019 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+package com.waz.zclient.security
+
+import java.util.Date
+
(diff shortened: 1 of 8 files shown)


app/src/main/scala/com/waz/zclient/security/AppLockActivity.scala
```diff
@@ -0,0 +1,92 @@
...
+
+  private var timeEnteredBackground: Option[Date] = Some(new Date(0))
+
+  def updateBackgroundEntryTimer(): Unit = {
+    timeEnteredBackground = Some(new Date())
+  }
+
+  def isAppLockExpired: Boolean = {
+    val now = new Date()
+    val secondsSinceEnteredBackground = (now.getTime - timeEnteredBackground.getOrElse(now).getTime) / 1000
```
gizemb: the calculation can be moved to a Util class.

app/src/main/scala/com/waz/zclient/security/AppLockActivity.scala
```diff
@@ -0,0 +1,92 @@
...
+      info(l"authentication successful")
+      timeEnteredBackground = None
+      finish()
+    }
+  }
+}
+
+object AppLockActivity extends DerivedLogTag {
+
+  val ConfirmDeviceCredentialsRequestCode = 1
```
gizemb: can be final

app/src/main/scala/com/waz/zclient/security/SecureActivity.scala
```diff
@@ -17,34 +17,62 @@
...
+      _ <- securityChecklist.run()
+      shouldShowAppLock <- shouldShowAppLock
+    } yield {
+      if (shouldShowAppLock) showAppLock()
+    }
+  }
+
+  override protected def onPause(): Unit = {
+    super.onPause()
+    AppLockActivity.updateBackgroundEntryTimer()
```
gizemb: since we're doing the check on onStart, I think we can make this update on onStop

app/src/main/scala/com/waz/zclient/security/AppLockActivity.scala
johnxnguyen: I changed from `Date` to `Instant` which offers a cleaner way to measure differences between two times.

app/src/main/scala/com/waz/zclient/security/AppLockActivity.scala
johnxnguyen: Done

app/src/main/scala/com/waz/zclient/security/SecureActivity.scala
johnxnguyen: I put this in `onPause` because I wanted to mark the time as soon as the activity was no longer visible. In reality though, is there much of a time difference between when `onPause` and `onStop` are called?

app/src/main/scala/com/waz/zclient/security/AppLockActivity.scala
gizemb: I meant that AppLockActivity doesn't need to know about date calculations and this logic could be moved elsewhere. But since we have a deadline soon and this looks cleaner now, I'll approve it.

johnxnguyen: Ah right, yes it's true. I'll make a note to revisit this 👍

PR #2265, opened 2019-08-23
Feature:  device password policy

johnxnguyen:
## What's new in this PR?

The PR introduces the device admin password policy. The base implementation was written by @CalumMcCall in https://github.com/wireapp/wire-android/pull/2213. I have made some adjustments and connected it to the `SecurityCheckList` which is run each time the app comes to the foreground.

There are two separate checks with corresponding actions. Firstly, we check if Wire is a device admin. If it is not, the user must activate this in the settings. We display a non cancellable alert dialog with a single action that will take the user to the settings perform activation.

Secondly, we use the security policy to check if the device password meets strength requirements.  If it does not, we will display a non cancellable alert dialog with a single action that will take the user to the security settings to change the password.

## Todo

The actual password policy still needs to be defined. This would be set in `SecurityPolicyService`.

The translatable strings need final copy. I've just made up some copy for now. We should also check to see if we can get a string describing the password requirements from the `DevicePolicyManager` API, so we don't have to describe the requirements ourselves. I saw that when setting the device password, there is already a label describing the requirements.


app/src/main/scala/com/waz/zclient/security/SecureActivity.scala
```diff
@@ -33,14 +39,15 @@ class SecureActivity extends AppCompatActivity with ActivityHelper with DerivedL
...
   private lazy val globalPreferences = inject[GlobalPreferences]
 
   override def onStart(): Unit = {
     super.onStart()
 
     for {
-      _ <- securityChecklist.run()
-      shouldShowAppLock <- shouldShowAppLock
+      allChecksPassed <- securityChecklist.run()
+      shouldShowAppLock <- shouldShowAppLock if allChecksPassed
```
johnxnguyen: This change means that the app lock won't be displayed if a check fails. I don't like this because it assumes that when a check fails, the app is blocked and unusable in some way. I've added it here to prevent the app lock from showing over blocking dialogs. I would like to address this in a later PR to see if there is a more robust way of doing this.

PR #2276, opened 2019-09-02
fix: Make sure the AppLockActivity is not created from LaunchActivity

makingthematrix:
In order to ask for the Android password/PIN we have to start `AppLockActivity`. To do that, we need to provide a reference to the parent activity. Right now we do this by choosing whichever activity triggered the request to ask for the password/PIN. At the start of the application, that's `LaunchActivity`. `LaunchActivity` triggers `AppLockActivity`, but it also chooses to start `MainActivity` or `AppEntryActivity`, and then closes itself.
Depending on the device, the order in which these activities are started may differ. It may lead to the situation when Main- or AppEntryActivity goes on top of `AppLockActivity` and then freezes, because it's waiting for user authentication, which never comes, because `AppEntryActivity` is not on top.

The fix simply ignores `LaunchActivity`. We know for sure that just after `LaunchActivity` another activity will be started, and we can start `AppLockActivity` from that one, ensuring that it will always be on top.

### DIFF
--- a/app/src/main/scala/com/waz/zclient/security/SecurityLifecycleCallback.scala
+++ b/app/src/main/scala/com/waz/zclient/security/SecurityLifecycleCallback.scala
@@ -20,22 +20,30 @@ package com.waz.zclient.security
 import android.app.{Activity, Application}
 import android.os.Bundle
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
-import com.waz.zclient.{Injectable, Injector}
+import com.waz.zclient.{Injectable, Injector, LaunchActivity}
 import com.waz.zclient.log.LogUI._
 
 class SecurityLifecycleCallback(implicit injector: Injector) extends Application.ActivityLifecycleCallbacks with Injectable with DerivedLogTag {
   private var activitiesStarted = 0
 
   override def onActivityStarted(activity: Activity): Unit = synchronized {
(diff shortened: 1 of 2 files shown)


makingthematrix:
build again

app/src/main/scala/com/waz/zclient/security/SecurityLifecycleCallback.scala
```diff
@@ -20,22 +20,34 @@ package com.waz.zclient.security
...
 
   override def onActivityStarted(activity: Activity): Unit = synchronized {
-    activitiesStarted += 1
-    verbose(l"onActivityStarted, activities started now: $activitiesStarted")
-    if (activitiesStarted == 1) inject[SecurityPolicyChecker].run(activity)
+    lazy val activityName = activity.getClass.getSimpleName
+    activity match {
+      case _: LaunchActivity =>
+      case _ if startedActivities.contains(activityName) =>
+        error(l"The activity $activityName is started again before being stopped")
```
gizemb: this shouldn't happen in a normally functioning app and we need to be aware of such a thing rather than silently logging it. so I think we should switch back to counting the number of activities again.

makingthematrix: ok, done

PR #2277, opened 2019-09-03
testing gallery bugfixes

Gardosen:
- implemented the usage of file provider to make the testing gallery compatible with Android 7.0+

### DIFF
--- a/testing_gallery/src/main/java/com/wire/testinggallery/SharingCommandReceiver.java
+++ b/testing_gallery/src/main/java/com/wire/testinggallery/SharingCommandReceiver.java
@@ -25,6 +25,7 @@ import android.content.pm.PackageManager;
 import android.content.res.Resources;
 import android.net.Uri;
 import android.support.annotation.NonNull;
+import android.support.v4.content.FileProvider;
 
 import java.io.File;
 import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


testing_gallery/src/main/java/com/wire/testinggallery/SharingCommandReceiver.java
```diff
@@ -108,7 +109,8 @@ private Intent createFilesShareIntent(String requestedPackage, List<File> files)
     @NonNull
     private Intent createFileShareIntent(String requestedPackage, File file) {
         Intent shareIntent = createShareIntent(requestedPackage);
-        shareIntent.putExtra(Intent.EXTRA_STREAM, Uri.fromFile(file));
+        Uri fileUri = FileProvider.getUriForFile(MainActivity.context, MainActivity.context.getString(R.string.file_provider_authority), file);
```
gizemb: we already have the context in the method that calls this method. We can simply pass that context as a parameter to that method. That way, we can get rid of the extra code in MainActivity. It is always better to use the specific context the component is running, rather than calling the Application context.

Gardosen: changed it, and passed the content from the parental method to createFileShareIntent

PR #2296, opened 2019-09-17
Disable FlowManager on emulator

marcoconti83:
This is a temporary measure to allow QA to use emulator. Needed because AVS does not implement a specific function for x86 architecture, therefore starting the app on the emulator results in an error.

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/call/FlowManagerService.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/call/FlowManagerService.scala
@@ -62,19 +62,17 @@ class DefaultFlowManagerService(context:      Context,
     doWithFlowManager(_.networkChanged())
   }
 
-  lazy val flowManager: Option[FlowManager] = Try {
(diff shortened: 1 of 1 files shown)


wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/call/FlowManagerService.scala
```diff
@@ -62,9 +63,17 @@ class DefaultFlowManagerService(context:      Context,
   }
 
   lazy val flowManager: Option[FlowManager] = Try {
-    val fm = new FlowManager(context, null, if (globalPrefs.getFromPref(AutoAnswerCallPrefKey)) avsAudioTestFlag else 0)
-    fm.addListener(flowListener)
-    fm
+    FlowManagerService.isEmulator match {
```
johnxnguyen: Returning `null` will assign `Some(null)` to `flowManager`. What we need is something like

```
flowManager = if (isEmulator) None else  Try { .... }
```

marcoconti83:
build again

PR #2297, opened 2019-09-18
Refactor custom build configuration loading

marcoconti83:
## What's new in this PR?

Main changes:
- Read the `GITHUB_API_TOKEN` environment variable instead of the generic `API_TOKEN`
- Print out the build configuration (whether it's custom or not) so that it's easy to verify what was built
- Merge the cusom configuration JSON (if any) on top of the default one. This change allow us to put in the custom configurations only the difference compared to the base configuration, so that we don't need to copy every new key there
- The APK name includes the application bundle

### Issues

This PR aims at make custom configuration loading cleaner, so that it's easier to debug on the build machine when something goes wrong

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -36,11 +36,11 @@ if (project.file('user.gradle').exists()) {
 }
 
 task copyCustomResources(type: Copy) {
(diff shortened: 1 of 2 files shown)


marcoconti83:
build again

marcoconti83:
build again

PR #2318, opened 2019-09-27
[Fix] Test processing of asset message pairs

johnxnguyen:
## What's new in this PR?

This PR includes a unit test for the fix introduce in https://github.com/wireapp/wire-android/pull/2313. Specifically, it is testing that if we process the original and upload asset message events together in the same batch, then we won't encounter a null pointer exception and we will successfully merge the assets.
#### APK
[Download build #86](http://10.10.124.11:8080/job/Pull%20Request%20Builder/86/artifact/build/artifact/wire-dev-PR2318-86.apk)

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/test/scala/com/waz/service/MessageEventProcessorSpec.scala
+++ b/wire-android-sync-engine/zmessaging/src/test/scala/com/waz/service/MessageEventProcessorSpec.scala
@@ -23,18 +23,20 @@ import com.waz.content._
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
 import com.waz.model.ConversationData.ConversationType
 import com.waz.model.GenericContent.Text
+import com.waz.model.GenericContent.Asset
 import com.waz.model._
+import com.waz.model.nano.Messages
 import com.waz.model.otr.UserClients
 import com.waz.service.assets2.{AssetService, DownloadAssetStorage}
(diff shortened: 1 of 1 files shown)


wire-android-sync-engine/zmessaging/src/test/scala/com/waz/service/MessageEventProcessorSpec.scala
```diff
@@ -179,6 +181,61 @@ class MessageEventProcessorSpec extends AndroidFreeSpec with Inside with Derived
...
+      (storage.updateOrCreateAll _).expects(*).onCall { updaters: Map[MessageId, Option[MessageData] => MessageData] =>
+        Future.successful(updaters.values.map(_.apply(None)).toSet)
+      }
+
+      // We will save both assets
+      (assets.save _).expects(*).twice().returns(Future.successful(()))
+
+      // When
+      val processor = getProcessor
+      inside(result(processor.processEvents(conv, isGroup = false, Seq(originalEvent, uploadEvent))).head) {
```
marcoconti83: What do you think about adding another test where the order is inverted?

johnxnguyen: To test that it fails?

marcoconti83: Discussed in 1:1, addressing our of order messages will be a separate ticket

PR #2327, opened 2019-10-02
Chore: Remove some commented out, non maintained tests

marcoconti83:
## What's new in this PR?

Remove some commented out code. These were tests, I tried to un-comment it but it does not compile anymore. I doubt we will go back to analyze this commented code any time soon, and I think having large chunks of commented code is bad practice.

The downside is that by removing this code we forget that we were meant to test that class. My long term strategy is to add code coverage, then let the coverage report tell us which classes are not covered.

#### APK
[Download build #105](http://10.10.124.11:8080/job/Pull%20Request%20Builder/105/artifact/build/artifact/wire-dev-PR2327-105.apk)


marcoconti83:
> It would be good to make a Jira ticket with a link to this PR to help us not forget to retest these components.

I would prefer to rely on some code coverage report instead in the future. The coverage should tell us which classes/packages are not covered. 

The reason I am skeptical about creating a ticket is that one of tests I removed was one of the ~5 test suites in a package that has ~20 files that are meant to be tested. I'm pulling number from memory, but that's the ballpark figure. The coverage there is already pretty horrible, so I think there is little value in remembering that we need to cover one of those and not the 15 others. I suspect there are other sections of the app in a similar state.

PR #2347, opened 2019-10-09
Fix/folders property payload

marcoconti83:
## What's new in this PR?

Fix cross-platform compatibility with webapp and iOS implementation for updates to the folders.

### Issues

There is a remote user property called `labels`. The Android application was assuming the **value** of that `label` property to be:
```
[
   {folder1},
   {folder2}
]
```

The iOS and webapp were assuming the **value** of that property to be:
```
{
   "labels": [
     {folder1},
     {folder2}
   ]
}
```

The webapp and iOS assumption comes from:

- Best practices of not having JSON array as top-level item (see [JSON Hijacking](https://haacked.com/archive/2009/06/25/json-hijacking.aspx/))
- Future-proofing for versioning (e.g. ability to introduce a "labels_v2" key later)

### Solutions

I updated the Android implementation to expect the same payload as iOS and webapp.

### Testing

Unit tests have been updated, and they cover the serialization extensively. Additionally, I tested the Android implementation on device against the web implementation on device, and they synchronize properly.

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/model/Event.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/model/Event.scala
@@ -27,6 +27,7 @@ import com.waz.model.UserData.ConnectionStatus
 import com.waz.model.otr.{Client, ClientId}
 import com.waz.service.PropertyKey
 import com.waz.service.conversation.RemoteFolderData
+import com.waz.service.conversation.RemoteFolderData._
 import com.waz.sync.client.ConversationsClient.ConversationResponse
 import com.waz.sync.client.OtrClient
 import com.waz.utils.JsonDecoder._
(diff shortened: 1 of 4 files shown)


makingthematrix:
I have some comments about making the code more Scala-idiomatic, but we can merge this PR and make changes in another.

1. We have now many layers of case classes describing folders: `FolderData`, `RemoteFolderData`, `IntermediateFolderData`, `FoldersPropertyRemotePayload`, and `FoldersEvent`. I think we can at least make `IntermediateFolderData` a private case class used only in `FoldersPropertyRemotePayload` (also, can we rename `FoldersPropertyRemotePayload` to e.g. `FoldersProperty`?) and never expose it to other parts of the code. `FoldersProperty` should be able to map the json to `Seq[RemoteFolderData]` by itself.
2. There's no need to use `new` when creating case classes, and overloading constructors is usually done with the `apply` method in the companion object (although, I don't think we need to do that).
3. Cosmetic: a case class without any methods or internal fields doesn't need empty curly brackets `{}` after the declaration, and by convention we declare an empty sequence with `Seq.empty[T]`, not `Seq[T]()` - the latter suggests that every empty sequence is a separate object while it only returns a reference to the same empty sequence (welcome to the Scala's functional programming world :) ).

Issue #193, opened 2016-09-30
help extract camera into standalone project 


madibalive:
Hi , currently in the process painstakingly extraction the camera module into a standalone, i realized the library and files are spread across , am in rewriting parts that have tight coupling 
Has anyone tried this and or can direct me in the right direction , any help would be appreciated ??

deanrobertcook:
Hi @madibalive, the best way would be to extract `GlobalCameraController.scala`, `CameraPreviewTextureView.scala` and `CameraPreviewObserver.java` into a separate module/project. The trickiest part would be to get the dependency injection working, but for a quick prototype you could just have a singleton `GlobalCameraController` created in your app, and when you construct a `CameraPreviewTextureView` just find that singleton and use that instead. 

The `PermissionsController` is only necessary if you want to handle Android M permissions, so you can probably remove it and just execute the blocks of code in its methods directly.

The other thing you would need as well is the `CancellableFuture` class from our [SE](https://github.com/wireapp/wire-android-sync-engine) project, but that has only taken me about 10 minutes to do in the past, since it's pretty isolated from everything.

Once you have all the dependencies sorted out, it's just a matter of placing an instance of `CameraPreviewTextureView` somewhere into your view hierarchy and adding a listener to it.

I hope that provides some high-level guidance. There are of course things I'll have forgotten about, but let me know how it goes. This is something I'd also be interested in doing myself, if I had the time :)

aftabsikander:
@deanrobertcook  I am looking into this myself in process of making a standalone project for camera and video feature. A quick question why **CameraPreviewTextureView** and other camera related classes are written in **scala** any special reason.

aftabsikander:
@madibalive Were you able to extract camera module into a standalone module. If do can you please share link.

marcoconti83:
Hi all,
it has been 3 years since the last comment on this thread. I will close the issue as I assume it's not relevant anymore. Feel free to reopen if you have more questions!

PR #2349, opened 2019-10-10
[Chore] Add content descriptions for rows

johnxnguyen:
### Issues

QA needs to be able to identify the different types of rows on the conversation folders list. Specifically, they need to be able to identify what kind of item a row represents (incoming conversation, normal conversation, section header) as well as some state information (is the header expanded or collapsed, or which section a row belongs to).

### Solutions

Add content descriptions for each item type and set the content description on the row views when binding the items.
#### APK
[Download build #193](http://10.10.124.11:8080/job/Pull%20Request%20Builder/193/artifact/build/artifact/wire-dev-PR2349-193.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/conversationlist/adapters/ConversationListAdapter.scala
+++ b/app/src/main/scala/com/waz/zclient/conversationlist/adapters/ConversationListAdapter.scala
@@ -72,7 +72,7 @@ abstract class ConversationListAdapter
   override def getItemId(position: Int): Long = items(position) match {
     case Item.IncomingRequests(first, _) => first.str.hashCode
     case Item.Header(id, _, _)           => id.str.hashCode
-    case Item.Conversation(data)         => data.id.str.hashCode
+    case Item.Conversation(data, _)      => data.id.str.hashCode
   }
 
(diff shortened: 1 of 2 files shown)


app/src/main/scala/com/waz/zclient/conversationlist/adapters/ConversationListAdapter.scala
```diff
@@ -209,18 +228,22 @@ object ConversationListAdapter {
 
     override def areItemsTheSame(oldItemPosition: Int, newItemPosition: Int): Boolean = {
       (oldList(oldItemPosition), newList(newItemPosition)) match {
-        case (lhs: Header,       rhs: Header)           => lhs.id == rhs.id
-        case (lhs: Conversation, rhs: Conversation)     => lhs.data.id == rhs.data.id
-        case (_: IncomingRequests, _: IncomingRequests) => true
-        case _                                          => false
+        case (lhs: Header, rhs: Header) =>
```
marcoconti83: We need to agree on a style. @johnxnguyen and @makingthematrix , you risk changing each other code back and forth because you use a different style

johnxnguyen: I put these on new lines because otherwise the case lines would be too long. In which case, it doesn't make sense to align the `=>`'s.

PR #224, opened 2016-10-07
[AN-4532] Ephemeral message stuff

hakonbo:
#### APK
[Download build #7693](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7693/artifact/build/artifact/wire-dev-PR224-7693.apk)
[Download build #7694](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7694/artifact/build/artifact/wire-dev-PR224-7694.apk)
[Download build #7695](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7695/artifact/build/artifact/wire-dev-PR224-7695.apk)

PR #223, opened 2016-10-07
Simplify account details info @ Raphael's request

infotexture:
#### APK
[Download build #7682](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7682/artifact/build/artifact/wire-dev-PR223-7682.apk)

PR #2361, opened 2019-10-11
Fix: Store folder collapse state per user

marcoconti83:
### Issues

The folder collapse state was previously the same across all accounts

### Causes

The fragment survives the switching of accounts, and the preference was not reloaded. The preference was loaded once at creation of the fragment.

### Solutions

Instead of loading the property once, register for changes on a signal instead.

### Testing

Manually tested on emulator with two accounts.
#### APK
[Download build #223](http://10.10.124.11:8080/job/Pull%20Request%20Builder/223/artifact/build/artifact/wire-dev-PR2361-223.apk)
[Download build #224](http://10.10.124.11:8080/job/Pull%20Request%20Builder/224/artifact/build/artifact/wire-dev-PR2361-224.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/conversationlist/ConversationListFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/conversationlist/ConversationListFragment.scala
@@ -88,23 +88,17 @@ abstract class ConversationListFragment extends BaseFragment[ConversationListFra
     }
   }
 
-  protected def beforeListCreation(): Future[Unit] = Future.successful(())
-
   override def onCreateView(inflater: LayoutInflater, container: ViewGroup, savedInstanceState: Bundle) =
     inflater.inflate(layoutId, container, false)
 
   override def onViewCreated(view: View, savedInstanceState: Bundle) = {
     super.onViewCreated(view, savedInstanceState)
-    for {
-      _ <- beforeListCreation()
-    } yield {
-      conversationListView.foreach { lv =>
-        lv.setLayoutManager(new LinearLayoutManager(getContext))
-        lv.setAdapter(adapter)
-        lv.setAllowSwipeAway(true)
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/conversationlist/ConversationListFragment.scala
```diff
@@ -356,26 +336,36 @@ class ConversationFolderListFragment extends NormalConversationFragment {
         groups    <- convListController.groupConvsWithoutFolder
         oneToOnes <- convListController.oneToOneConvsWithoutFolder
         custom    <- convListController.customFolderConversations
-      } yield (incoming, favorites, groups, oneToOnes, custom)
+        prefs     <- userPreferences
+        states    <- prefs(ConversationFoldersUiState).signal
```
johnxnguyen: Can't you use the `foldersUiState` signal defined above?

app/src/main/scala/com/waz/zclient/conversationlist/ConversationListFragment.scala
```diff
@@ -323,30 +323,10 @@ class ConversationFolderListFragment extends NormalConversationFragment {
   import UserPreferences.ConversationFoldersUiState
 
   private lazy val userPreferences = inject[Signal[UserPreferences]]
-  private var foldersUiState: FoldersUiState = Map.empty
-
-  override protected def beforeListCreation(): Future[Unit] = loadFoldersUiState()
```
gizemb: if this is not used, we can delete the overridden method too. I think nothing else uses it

johnxnguyen: You can delete this from the base class now, as well as it's invocation.

Issue #231, opened 2016-10-12
Forwarding video loses audio


someoneEsle:
I've already reported this issue over two months ago, but since I ran into it again today...
Steps to reproduce (Android 7, Nexus 5X):
- Open a conversation, share a file, select a video*
- Once the video has been sent, long press on it, forward, select another wire conversation
- The new contact you sent the video to will see the video correctly but audio will be missing

Let me know if you need anything else.

*by the way, sharing videos from gallery fails some 99% of the time (I also reported this in the past..), leaving me with the only choice of sharing from within Wire

freditbn:
my issue is very similar:

No matter how i try to forward videos videos from wire, sound is lost when forwarded. The videos play silently.

The sound also disappears when videos not recorded on wire are forwarded within the app.

The sound is lost on the sender and recipient device. No matter what video player is launched to play the forwarded video.

when the video is saved on the device from within wire, the sound plays when the video is played. However when the downloaded video is sent from the gallery or an explorer to a wire contact. the sound is also lost.

Sent videos from gallery that have never been in a wire conversation before seem to keep sound.

Environment

Wire version that exhibits the issue: 2.41.359
Last Wire version that did not exhibit the issue (if applicable): n/a
Android OS version used to run Wire: 7.0
Is this a custom firmware or a stock one: stock LG, V10f-AME-XX, rooted
Mobile phone model/manufacturer: LG V20, H990DS
Mobile network type (EDGE/LTE/Wi-Fi/Offline): LTE/WiFi

makingthematrix:
@matt3055 Is your bug still happening? From the description I assume it's different from the original one (and that one was fixed afaik).

PR #227, opened 2016-10-09
SE 81.1.300/81.1232

hakonbo:
#### APK
[Download build #7685](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7685/artifact/build/artifact/wire-dev-PR227-7685.apk)
[Download build #7687](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7687/artifact/build/artifact/wire-dev-PR227-7687.apk)

PR #2364, opened 2019-10-13
[Feature] Expand on move to folder

johnxnguyen:
## What's new in this PR?

This PR contains changes to expand the folder when a conversation is added to favorites or moved to a custom folder. Other changes include

- The logic for the folder states has been extracted into `FolderStateController`.
- Folders are only expanded / collapsed if not already so.

## Notes

I'm not particularly happy with the implementation, namely that the `ConversationListController` now holds the reference to the `FolderStateController`. The reasons why I move the responsibility from the `ConversationFolderListFragment`:

1. The datasource of the `ConversationFolderListAdapter` is a large amalgamation of signals, therefore it only reacts to changes but doesn't have any context about those changes. Instead of trying to figure out from the datasource if a conversation was favorited or moved to a custom folder, it made more sense to update the folder states directly.
2. This however needs to be done before the datasource binds to the adapter, which means we need to update the folder states before we add the conversation to the new folder.
3. To ensure this happens synchronously, I found I needed to do this in the `ConversationListController` methods.
4. Finally, since the `ConversationListController` knows nothing about the fragment, it needs to take over responsibility of the folder states.

Furthermore, the specs say we should also scroll to the conversation after expanding the folder. I couldn't find any simple and /or stable way to do this. I'm open to ideas!

#### APK
[Download build #227](http://10.10.124.11:8080/job/Pull%20Request%20Builder/227/artifact/build/artifact/wire-dev-PR2364-227.apk)
[Download build #231](http://10.10.124.11:8080/job/Pull%20Request%20Builder/231/artifact/build/artifact/wire-dev-PR2364-231.apk)


app/src/main/scala/com/waz/zclient/conversationlist/FolderStateController.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  def store(states: FolderUiStates): Future[Unit] = for {
+    prefs <- userPreferences.head
+    _     <- prefs(ConversationFoldersUiState).update(states)
+  } yield {}
+}
+
+object FolderStateController {
+
+  type FolderUiStates = Map[Uid, Boolean]
```
makingthematrix: It looks like everywhere we could simply use `FolderId`s and `Boolean`s, sometimes as tuples, sometimes as separate parameters. I don't see any advantage in creating a new case class and a type alias, and in using `Uid` instead of `FolderId`.

app/src/main/scala/com/waz/zclient/conversationlist/FolderStateController.scala
```diff
@@ -0,0 +1,67 @@
...
+  } yield {}
+
+  def prune(folderIds: Set[Uid]): Future[Unit] = for {
+    state               <- folderUiStates.head
+    knownStates          = state.keySet
+    unusedFolderStates   = knownStates -- folderIds
+    _                   <- store(state -- unusedFolderStates)
+  } yield {}
+
+  def store(states: FolderUiStates): Future[Unit] = for {
```
makingthematrix: This can be a private method.

app/src/main/scala/com/waz/zclient/conversationlist/FolderStateController.scala
johnxnguyen: It's mostly just to give a name to types to improve readability. The header items need an id, and I thought it was appropriate not to strictly couple the header item with a folder. For this reason I gave the header ids the type of `Uid`. I'll change them.

PR #2368, opened 2019-10-14
fix: Sort custom folders alphabetically

makingthematrix:


johnxnguyen:
The list of folders is already sorted by the `ConversationLIstController` in the signal `customFolderConversations`. However, there it is sorted by the conversation `name`. Perhaps it's better to fix it there?

johnxnguyen:
Actually, that it's sorted by `name` is irrelevant. I'm just wondering why the sort there doesn't work.

PR #252, opened 2016-10-19
Ephemeral system notifications

hakonbo:
#### APK
[Download build #7754](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7754/artifact/build/artifact/wire-dev-PR252-7754.apk)
[Download build #7756](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7756/artifact/build/artifact/wire-dev-PR252-7756.apk)
[Download build #7758](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7758/artifact/build/artifact/wire-dev-PR252-7758.apk)

Issue #244, opened 2016-10-16
Missing a "send" key when in landscape


mimf:
I don't know if it was always like this or just since the last update. I think it's since the last update, because I never noticed and it's certainly annoying.

When I'm typing a message in landscape mode, there's no "send" key (the arrow-looking button). It's there in portrait mode. So in order to send a message, I type in landscape and then change to portrait, to send.

bjoernherbig:
Thanks for reporting this. We will check. What device are you on?

mimf:
Android 5.0.2

bjoernherbig:
right. i meant the tablet type you are using.

mimf:
It's not a tablet, it's a smartphone. An LG smartphone.

vova-lohika:
We do not support landscape mode for smartphones. Only tablets.

mimf:
Oh, I'm using an app to change orientation for the apps that somewhy do not support it. My bad. I guess I'll create a feature request issue then.

PR #2375, opened 2019-10-17
fix: If obfuscation the "Recents" list is impossible, remove Wire from the list

makingthematrix:
**Please don't merge before we decide what Android versions should use what solution**

Implements an alternative solution to obfuscation of the "Recents" list for Android 5.1 where the obfuscation doesn't work:
Instead the app is completely removed from the list of "Recents" apps. It can be still accessed simply by clicking the icon on the main screen of the device.

I rewrote the old `SecurityLifecycleCallback` into `ActivityLifecycle`. Now it exposes a signal and classes which want to react to going to the background may subscribe to the signal (that's the process reversal - `SecurityLifecycleCallback` called `SecurityPolicyChecker` directly). In result, `ActivityLifecycle` serves now a similar role to `UiLifeCycle`, but it doesn't rely on `BaseActivity` calling `acquireUi` and `releaseUi`. I think in another PR we should replace `UiLifeCycle`
with it.

#### APK
[Download build #244](http://10.10.124.11:8080/job/Pull%20Request%20Builder/244/artifact/build/artifact/wire-dev-PR2375-244.apk)
[Download build #252](http://10.10.124.11:8080/job/Pull%20Request%20Builder/252/artifact/build/artifact/wire-dev-PR2375-252.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/WireApplication.scala
+++ b/app/src/main/scala/com/waz/zclient/WireApplication.scala
@@ -89,7 +89,7 @@ import com.waz.zclient.pages.main.conversationpager.controller.ISlidingPaneContr
 import com.waz.zclient.pages.main.pickuser.controller.IPickUserController
 import com.waz.zclient.participants.ParticipantsController
 import com.waz.zclient.preferences.PreferencesController
-import com.waz.zclient.security.{SecurityLifecycleCallback, SecurityPolicyChecker}
+import com.waz.zclient.security.{ActivityLifecycle, SecurityPolicyChecker}
 import com.waz.zclient.tracking.{CrashController, GlobalTrackingController, UiTrackingController}
 import com.waz.zclient.utils.{AndroidBase64Delegate, BackStackNavigator, BackendController, ExternalFileSharing, LocalThumbnailCache, UiStorage}
 import com.waz.zclient.views.DraftMap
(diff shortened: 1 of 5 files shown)


app/src/main/scala/com/waz/zclient/security/ActivityLifecycle.scala
```diff
@@ -20,22 +20,24 @@ package com.waz.zclient.security
 import android.app.{Activity, Application}
 import android.os.Bundle
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
-import com.waz.zclient.{Injectable, Injector, LaunchActivity}
+import com.waz.utils.events._
 import com.waz.zclient.log.LogUI._
+import com.waz.zclient.{Injectable, Injector, LaunchActivity}
 
-class SecurityLifecycleCallback(implicit injector: Injector)
+class ActivityLifecycle(implicit injector: Injector)
```
gizemb: ActivityLifecycle**Callback**, **Helper** etc. would be better I think

app/src/main/scala/com/waz/zclient/WireApplication.scala
```diff
@@ -445,6 +447,11 @@ class WireApplication extends MultiDexApplication with WireContext with Injectab
       inject[MessageNotificationsController],
       assets2Module)
 
+    val securityCallback = inject[ActivityLifecycle]
```
gizemb: val activityLifecycleCallback = inject[ActivityLifecycleCallback]
this variable is not specific to security callbacks anymore.

PR #251, opened 2016-10-19
Various ephemeral design tweaks

hakonbo:
#### APK
[Download build #7752](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7752/artifact/build/artifact/wire-dev-PR251-7752.apk)
[Download build #7753](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7753/artifact/build/artifact/wire-dev-PR251-7753.apk)
[Download build #7755](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7755/artifact/build/artifact/wire-dev-PR251-7755.apk)
[Download build #7760](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7760/artifact/build/artifact/wire-dev-PR251-7760.apk)

Issue #248, opened 2016-10-19
Impossible to know who is a contact with pseudo


ElucGeek:
I have a new contact in my Wire contact list, it's pseudo is random (like abcde) and his picture is a landscape. How can I know who is he from my address book ? Can I see his email or phone number ? There should be an indication telling me from where I got the contact (i.ie manually added, linked from my address book, discovered from my email, discovered from my phone number).

bjoernherbig:
acknowleded. we are planning improvements to this soon. no timeline to share yet though.

h-2:
I have stated in this also in #317 

But just to give more explanation:
It is not only about people that I don't know, but it is also confusing if I actually know the poeple, but they didn't enter their full name or a nickname. E.g. I am told that Bob uses Wire, but I have 5 different Bobs in my address book.
That is precisely why integration with the address book is good, if I give someone the name "Bob (friendly guy from Debian)" in my address book than it should show up as "Bob (friendly guy from Debian)" in Wire and not "Robert Johnson" which I have no clue about. 

I would also propose this for user pictures, at least in the cases where there is a user picture in my address book. This is a huge recognition factor on my device, because the image is used in my e-mail program, my sms-program, Kontalk, Signal, Conversations...

IpsmLorem:
Linking to the adress book should be optional. User should be able to choose local name and picture if they do not want to risk contact book leakage and still have the ability to have a clean/usable contact list.

cf. https://github.com/wireapp/wire-ios/issues/208

someoneEsle:
Although the last update brings usernames and allows details fetched from the address book to be displayed (when tapping someone's name within a conversation), I would still say that Wire needs to go one step further and *always* prioritise the names fetched from the address book -- both in the conversations list and in the people list.
This is especially important when forwarding something, for example. Right now, having the address book's details hidden behind a few taps after opening a conversation does not help telling who is who while forwarding.

gilshwartz:
I would suggest to simply add a nick property to a contact, which by default is the same as the contact's name. However, a user can locally (for his or her own personal use) change the nick and the contacts list will use the nick for display, rather than the contact's name.

n3niu:
[...]

Issue #246, opened 2016-10-17
[Feature Request] Allow read/write messages, while having a call


boldt:
This is a FR: 
Unfortunately, it is not possible to read/write messages, while having a call.

xtotdam:
Hey! You say this is duplicate of #487 which is duplicate of #246 which is duplicate of #487 which is duplicate of #246 which is duplicate of #487.... I don't think I get it :)

someoneEsle:
Hey everyone! We just pushed a beta to the play store that adds this functionality. Let us know if you find any bugs!

boldt:
Nice to see, that my 2.5 years old issue got fixed. I'm looking forward to test it.

Issue #245, opened 2016-10-16
Different Emoticons on different devices


boldt:
Hello,

we have three devides:

* Sony Xperia Z5, Android 6.0.1, Wire 2.18.288
* Huawai Mate 8, Android 6.0 weith EMUI 4.0, Wire 2.18.288
* Amazon Fire, Android 5 (Fire OS 5.3.1.0), Wire 2.18.288

Emoticons on the Huaway (Screenshot) and Amazon Fire device:

![wire_1476653991580](https://cloud.githubusercontent.com/assets/750793/19421170/c00d3c20-93fa-11e6-848d-16aca13da713.png)

Emoticons on the Sony device:

![screenshot_20161016-230806](https://cloud.githubusercontent.com/assets/750793/19421238/4e2bcc46-93fc-11e6-9a04-6f59a7116306.png)

As you can see, the Sony has more Emoticons than the Huawai. 
Sending a not existing from the Sony to the Huaway, the Huaway just produces a crossed box.

bjoernherbig:
We use the system emoticons, so there can be differences in what the keyboard shows.

boldt:
Are you sure, this issue should be closed?

I mean, the as a sender, I dont know which emoticons exist on the receiving device. Unfortunatley, I produce quite a lot of "crossed Xes" by using my icons.

bjoernherbig:
we are using standard android fonts. if the receiving font does not provide support for certain emoticons, an empty box with a cross will be shown. this is in line with what you see e.g. in text messages.

PR #2376, opened 2019-10-18
[Fix] Stuck app

johnxnguyen:
### Issues

It could happen that after coming back online and opening the app, Wire would get stuck in the loading process. The result was typically visible in three ways:

1. The shield splash screen was shown indefinitely,
2. a black screen was shown indefinitely, or
3. the conversation list was shown, but some avatars were missing and tapping on a conversation did nothing (and the app was generally unusable).

Terminating the app and restarting would not resolve the problem.

### Causes

The exact cause is not known, but I have been able to reproduce this bug reliably and have located it to a very specific part of the app. The situation is as follows:

1. After logging in and being able to receive and send messages, terminate the app.
2. Enter airplane mode (so that we can accumulate notifications)
3. Send several messages to this device from another user with a profile picture.
4. Exit airplane mode
5. Launch the app.

When the app launches we'll process events from the notification stream to sync the app with the backend. Eventually, we will attempt to display user notifications in the UI. Since the sender has a profile picture, we will also attempt to display their avatar in the user notification.

We will ask `Glide` to fetch the image data for us, using a blocking request on a background thread:

```scala
Threading.Background {
  Option(WireGlide(cxt)
    .asBitmap()
    .load(picture)
    .apply(new RequestOptions().circleCrop())
    .submit(128, 128)
    .get()). // This method blocks!
}.future
```

`Glide` then will make the request to our backend via `ImageAssetFetcher`. The result is returned as a `CancellableFuture[InputStream]`, containing the image data. We then wait **_indefinitely_** for this future to complete, before passing the input stream into a callback. This also happens on a background thread:

```scala
Try { Await.result(data, Duration.Inf) } match {
  case Failure(err) =>
    verbose(l"Asset loading failed $request, $err")
    callback.onLoadFailed(new RuntimeException(s"Fetcher. Asset loading failed: $err"))
  case Success(is) =>
    verbose(l"Asset loaded $request")
    callback.onDataReady(is)
}
```

I'm not entirely sure how these two blocking calls affect each other, as I'm not certain that they're causing a deadlock. I'm also confused about why this issue seems to only occur after a cold start after accumulating notifications. It could be that these blocking calls are also interfering, or being interfered by other activity related to backend syncing.

### Solutions

I came across a few solutions to this problem. First, replacing an infinite `Await` when fetching the image data with a timeout resulted in the app being stuck as long as the timeout, then proceeded to load as normal.

Second, removing the avatar from the user notification altogether resulted in the app not being stuck at all.

Both of these solutions (also combined) aren't optimal, sync they only assist in reducing the symptoms. They therefore are not the solutions I ended up with.

After refactoring the code related to the user notification, I found that the bug was fixed, even if we had an infinite `Await`. The refactoring was only structural, I simplified a messy nested `for` comprehension to make the code readable, resulting in a single `for` comprehension. I'm not sure why, but for some reason it fixes the bug, and we still get to keep the avatar in the user notification.

As a precaution, I replaced the infinite `Await` for the image data with non blocking callbacks which timeout after 30 seconds. While this has no effect on the user notifications, it will prevent blocking situations if the download gets stuck for some reason.

### Testing

Manually tested multiple times.

## Notes

The issue described above may also be a cause of the "stuck in sync" bug, and I expect to see this bug much less, (hopefully not at all).

Also, the logic for determine which picture to fetch (from a group of notifications) has been simplified. It was inefficiently searching all users in the notifications and only displaying an image if there was only one image found. This expense contributes to the slowness of the app launch if there are many accumulated notifications. The simplification is to merely show the avatar, if there is only a single notification.

#### APK
[Download build #245](http://10.10.124.11:8080/job/Pull%20Request%20Builder/245/artifact/build/artifact/wire-dev-PR2376-245.apk)
[Download build #246](http://10.10.124.11:8080/job/Pull%20Request%20Builder/246/artifact/build/artifact/wire-dev-PR2376-246.apk)
[Download build #248](http://10.10.124.11:8080/job/Pull%20Request%20Builder/248/artifact/build/artifact/wire-dev-PR2376-248.apk)
[Download build #249](http://10.10.124.11:8080/job/Pull%20Request%20Builder/249/artifact/build/artifact/wire-dev-PR2376-249.apk)
[Download build #251](http://10.10.124.11:8080/job/Pull%20Request%20Builder/251/artifact/build/artifact/wire-dev-PR2376-251.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/notifications/controllers/MessageNotificationsController.scala
+++ b/app/src/main/scala/com/waz/zclient/notifications/controllers/MessageNotificationsController.scala
@@ -380,10 +380,10 @@ class MessageNotificationsController(bundleEnabled: Boolean = Build.VERSION.SDK_
   private def getPictureForNotifications(userId: UserId, nots: Seq[NotificationData]): Future[Option[Bitmap]] =
     if (nots.size == 1 && !nots.exists(_.ephemeral)) {
       val result = for {
-        Some(userStorage) <- inject[AccountToUsersStorage].apply(userId)
-        user <- userStorage.get(nots.head.user)
-        Some(picture) = user.flatMap(_.picture)
-        bitmap <- loadPicture(picture)
+        Some(storage) <- inject[AccountToUsersStorage].apply(userId)
+        user          <- storage.get(nots.head.user)
+        picture        = user.flatMap(_.picture)
+        bitmap        <- picture.fold(Future.successful(Option.empty[Bitmap]))(loadPicture)
       } yield bitmap
 
       result.recoverWith {
--- a/app/src/main/scala/com/waz/zclient/notifications/controllers/MessageNotificationsController.scala
+++ b/app/src/main/scala/com/waz/zclient/notifications/controllers/MessageNotificationsController.scala
@@ -30,6 +30,7 @@ import com.waz.api.NotificationsHandler.NotificationType
 import com.waz.api.NotificationsHandler.NotificationType._
 import com.waz.content.{UserPreferences, _}
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
+import com.waz.model.UserData.Picture
 import com.waz.model._
 import com.waz.service.push.NotificationUiController
 import com.waz.service.{AccountsService, UiLifeCycle}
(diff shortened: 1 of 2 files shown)


marcoconti83:
Nice write-up @johnxnguyen !

app/src/main/scala/com/waz/zclient/notifications/controllers/MessageNotificationsController.scala
```diff
@@ -377,36 +378,36 @@ class MessageNotificationsController(bundleEnabled: Boolean = Build.VERSION.SDK_
...
-              }
-            } yield
-              bmp.map(Bitmap.fromAndroid)
-
-          } yield bitmap
-        case _ => Future.successful(None)
+      val result = for {
+        Some(userStorage) <- inject[AccountToUsersStorage].apply(userId)
+        userIds = nots.map(_.user).toSet
+        users <- userStorage.getAll(userIds).map(_.flatten)
```
makingthematrix: By the way, since we need only one picture, you might as well use `find` instead of `getAll`. It will return either the first user with id in the `userIds` or None.

johnxnguyen:
build again

PR #2378, opened 2019-10-18
fix: Don't try to sync history in response to web socket disconnecting

makingthematrix:
Right now we sync history after each change of web socket connection, even on disconnection,
which does nothing and ends in an exception being thrown. It's not a crash, but it's unnecessary,
and the fix is very simple.

#### APK
[Download build #250](http://10.10.124.11:8080/job/Pull%20Request%20Builder/250/artifact/build/artifact/wire-dev-PR2378-250.apk)


makingthematrix:
> If we don't sync history on web socket re-connection, we will lose messages. We have to sync history before we start processing new events from the web socket. More details [here](https://github.com/wearezeta/documentation/blob/master/topics/core/client-side-synchronization.md) but I'm happy to discuss this 1:1.

This is about disconnection, not reconnection. The current code tries syncing even if the changes is from a working network to offline.

marcoconti83:
Thanks for clarifying @makingthematrix !

PR #264, opened 2016-10-21
[AN-4228] Improved invite button

JPonte:
Ticket: https://wearezeta.atlassian.net/browse/AN-4228
#### APK
[Download build #7782](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7782/artifact/build/artifact/wire-dev-PR264-7782.apk)
[Download build #7784](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7784/artifact/build/artifact/wire-dev-PR264-7784.apk)
[Download build #7788](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7788/artifact/build/artifact/wire-dev-PR264-7788.apk)

PR #2382, opened 2019-10-22
[Feature] AVS 5.3

johnxnguyen:
## What's new in this PR?

This PR contains the changes required to bump AVS from 5.0 to 5.3. For a list of API changes, see [here](https://github.com/wearezeta/avs/tree/master/docs/api).

### Changes

With AVS 5.3 a user is now able to join a call from multiple devices. The API now gives us client ids via the callback handlers for the video state changes and group member changes. 

Up until now, we only paired a video state with a user id. However now the user id alone is not enough to uniquely identify a call participant. For this reason, a `Participant` is now defined by a `UserId` and `ClientId`. 

The `Participant` is used as a key in the `CallInfo` property `videoReceiveStates`. This change has rippling effects throughout the calling code, but mostly just to gain access to the `userId` within.

### Testing

- A unit test for decoding the json returned from AVS
- Manual test calls between device and emulator.

## Notes

I was a bit unsure about how to handle the call participant list, since we don't know the client ids of the participants there. However, after inspecting the iOS implementation of AVS 5.3, it appears that they do not show devices as individual participants, but rather users (as usual). I decided to follow this as a guide.

##  Bonus

Calling now works on the emulator!

#### APK
[Download build #264](http://10.10.124.11:8080/job/Pull%20Request%20Builder/264/artifact/build/artifact/wire-dev-PR2382-264.apk)
[Download build #265](http://10.10.124.11:8080/job/Pull%20Request%20Builder/265/artifact/build/artifact/wire-dev-PR2382-265.apk)


johnxnguyen:
build again

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/call/Calling.scala
```diff
@@ -25,6 +25,11 @@ import scala.concurrent.Promise
 
 object Calling {
 
+  // The wrapped `wcall` instance.
+  type Handle = Uint32_t
+
+  val WCALL_ENV_DEFAULT: Int = 0
```
marcoconti83: What is this variable doing? Can you add a comment?

johnxnguyen: Done.

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
```diff
@@ -185,7 +189,7 @@ class CallingFragment extends FragmentHelper {
           case _ => true
         }.sortWith {
           case (_:SelfVideoView, _) => true
-          case (v1, v2)             => v1.userId.str.hashCode > v2.userId.str.hashCode
+          case (v1, v2)             => v1.hashCode > v2.hashCode
```
gizemb: the hashCode computation here is changed. is that something that we want?

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
```diff
@@ -209,11 +213,11 @@ class CallingFragment extends FragmentHelper {
         }
 
         val viewsToRemove = viewMap.filter {
-          case (uid, selfView) if uid == selfId => !gridViews.contains(selfView)
-          case (uId, _)                         => !videoUsers.contains(uId)
+          case (participant, selfView) if participant == selfParticipant => !gridViews.contains(selfView)
```
gizemb: here again, if clientId's are different, it will fallback to other cases. is this intentional?

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
johnxnguyen: It seems like the sorting of other participant video views was arbitrary, in which case it still is.

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
johnxnguyen: I think yes, because if you join the call from another device, then that "participant" is considered to be just like any other participant (even if it is still you).

johnxnguyen:
build again

johnxnguyen:
build again

PR #2391, opened 2019-10-29
[Fix] Call participants not considering devices

johnxnguyen:
### Issues

If a user participates in a call with more than one device, they only appear as a single row in the call participants list.

### Causes

The participants list was being derived from the `CallInfo.others` property, which is a `Map[UserId, LocalInstant]`. Participants are now defined by `UserId` and `ClientId` pairs.

### Solutions

- Change the map to `Map[Participant, LocalInstant]`
- rename the property to `otherParticipants
- update all usages
- remove redundant usages

### Testing

Manually tested. AVS will check this thoroughly in their test suites.

## Notes

There is a potential bug on the AVS side where you do not see your other devices in the call, which is being investigated now.

Regardless, when your other devices do appear in the list, they are appended with "(YOU)". This is to align with iOS.


johnxnguyen:
build again

Issue #308, opened 2016-11-14
feedback loop persisted after call ended


Be-ing:
I connected with my friend through Wire when we were in the same room with some quiet background music playing. I called her Motorola Droid Maxx with Wire to test Wire's call functionality. Our phones were close to each other, within a few inches. With the latency of the call, each phone picked up the other playing a sound that originated from the music, creating a continuous loop of beeping, like a delay effect. Strangely, her phone kept making that noise after the call was ended. I had to force quit Wire on her phone to get the beeping to stop. I tried to reproduce the issue, but Wire was not picking up any sounds on her phone after I started Wire again. I restarted her phone and tried again. The continuous loop of beeping occurred as expected, but I could not replicate the issue of her phone continuing to make that sound after ending the call.

jspittka:
Running 2 phones close to each other will result in a feedback loop similar to what often is experienced on stage when a microphone picks up the speaker signal. We have some safeguards in the software to suppress the annoying feedback sound, and in combination with the latency, this may sound as you described it. Still this is not a typical use-case and the resulting sound is expected. What is not expected is that the call seems to have continued on one of your devices after the call was hung up. The only explanation that I can see is that unfortunately something went wrong on that device and either the device and/or Wire stalled. Recording and playout loops run on a very high priority and therefore kept running while everything else became unresponsive. After restarting the device the audio system was reset and worked normally again. That would also explain why you were not able to reproduce the issue later. We apologize for the inconvenience and hope this will not happen again as we constantly improve our software.

Be-ing:
>while everything else became unresponsive

Everything else did not become unresponsive. Neither the Wire UI nor the Android UI froze.

mykola-mokhnach:
@Be-ing I'm closing the report for now. Feel free to reopen it if you observe the thing works wrong.

PR #303, opened 2016-11-10
[CM-1122] Fix out of sync accent colors

JPonte:
Since the client receives the id of the colour and that id is the index
of the array, the removal of the yellow changed the keys of the local
colour map.
Ticket: https://wearezeta.atlassian.net/browse/CM-1122
#### APK
[Download build #7878](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7878/artifact/build/artifact/wire-dev-PR303-7878.apk)
[Download build #7879](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7879/artifact/build/artifact/wire-dev-PR303-7879.apk)
[Download build #7880](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7880/artifact/build/artifact/wire-dev-PR303-7880.apk)

PR #301, opened 2016-11-10
[AN-4668] Adds "open" option for files to message actions

shumengye:
#### Description
Adds "open" option to menu actions for file assets. 

#### Ticket
https://wearezeta.atlassian.net/browse/AN-4668
#### APK
[Download build #7870](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7870/artifact/build/artifact/wire-dev-PR301-7870.apk)
[Download build #7871](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7871/artifact/build/artifact/wire-dev-PR301-7871.apk)
[Download build #7872](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7872/artifact/build/artifact/wire-dev-PR301-7872.apk)
[Download build #7876](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7876/artifact/build/artifact/wire-dev-PR301-7876.apk)
[Download build #7877](http://192.168.10.18:8080/job/Pull%20Request%20Builder/7877/artifact/build/artifact/wire-dev-PR301-7877.apk)

Issue #297, opened 2016-11-08
[interface] wastes lots of space


h-2:
I know that design issues are kind of subjective and that wire is not aiming for material design, but I really think that it should be able to display more than 1.5 lines of dialog on one screen :frowning: 

![screenshot_20161108-081300](https://cloud.githubusercontent.com/assets/2996863/20113517/ff4f062c-a5f0-11e6-8675-f7098e64ed59.png)

I have shown where a lot of space is wasted. The most important thing is the additional time notifier in the middle, which seems to always be inserted after the conversation was interrupted for a while.

Some more humble suggestions:
  * move attachments, gifs and ping buttons behind one button to save the extra line
  * move name above or below avatar to have more space for text
  * remove redundancy between name in the top and avatar in the chat

Also I think it would vastly improve the usability, if the interface (input and topbar) would be more distinct from scrollview / chat area.

Anyways, all criticism is constructive. Thanks for the app!

bjoernherbig:
thanks. 
we take it as constructive feedback and food for thoughts. we're constantly discussing improvements to the ui so your feedback is appreciated. 
i'll close this for now to keep the issues lean, but we're taking your feedback into the discussion.

ElucGeek:
I feel exactly the same as @h-2 and was about to open an issue for this. In one to one conversation you can just see the last post with a couple of line, then it's hidden. When replaying in a group chat with many activity it's hard to follow while typing, you have to stop typing, close the keyboard, scroll, read, and come back. The more space you can save vertical, would be the better IMHO.
Is there any update about this ?

PR #2395, opened 2019-10-31
feat: Use Wire password and a dialog instead of the Android functionality

makingthematrix:
Instead of relying on a partially broken Android functionality for asking for password when coming out of the background, we will now use the Wire password and a dialog inside the app. The PR introduces a new property for counting failed attempts - in personal builds it's set to 0 and not used, but in builds like EY it's used to wipe the user's data after N failed attempts. Also, the biometric prompt can be used (if available) to speed up authentication.

I rewrote a big part of the security policy checker.
#### APK
[Download build #294](http://10.10.124.11:8080/job/Pull%20Request%20Builder/294/artifact/build/artifact/wire-dev-PR2395-294.apk)
[Download build #295](http://10.10.124.11:8080/job/Pull%20Request%20Builder/295/artifact/build/artifact/wire-dev-PR2395-295.apk)
[Download build #297](http://10.10.124.11:8080/job/Pull%20Request%20Builder/297/artifact/build/artifact/wire-dev-PR2395-297.apk)
[Download build #298](http://10.10.124.11:8080/job/Pull%20Request%20Builder/298/artifact/build/artifact/wire-dev-PR2395-298.apk)
[Download build #301](http://10.10.124.11:8080/job/Pull%20Request%20Builder/301/artifact/build/artifact/wire-dev-PR2395-301.apk)
[Download build #303](http://10.10.124.11:8080/job/Pull%20Request%20Builder/303/artifact/build/artifact/wire-dev-PR2395-303.apk)
[Download build #305](http://10.10.124.11:8080/job/Pull%20Request%20Builder/305/artifact/build/artifact/wire-dev-PR2395-305.apk)
[Download build #308](http://10.10.124.11:8080/job/Pull%20Request%20Builder/308/artifact/build/artifact/wire-dev-PR2395-308.apk)


app/src/main/res/values/strings.xml
```diff
@@ -108,6 +108,15 @@
     <string name="crashes_and_analytics_request_no">No</string>
     <string name="crashes_and_analytics_request_agree">I Agree</string>
 
+    <!-- Request Password dialog -->
+    <string name="request_password_title">Provide account password</string>
```
johnxnguyen: Why don't we use existing copy?

```
<string name="app_lock_locked_title">Wire is locked due to inactivity</string>
```

We also have

```
<string name="app_lock_locked_message">Please enter your device credentials.</string>
```

which could be adjusted to

```
<string name="app_lock_locked_message">Please enter your account password.</string>
```

app/src/main/res/values/strings.xml
```diff
@@ -108,6 +108,15 @@
...
+    <string name="request_password_ok">OK</string>
+    <string name="request_password_cancel">Cancel</string>
+    <string name="request_password_biometric_title">Biometric Prompt</string>
+    <string name="request_password_biometric_description">Use the biometric authentication or click cancel and provide the password</string>
```
johnxnguyen: Also these should use the existing copy, but modified for biometrics. Are we able to know what kind of biometric authentication it is?

makingthematrix: We don't know that so I just proposed a generic text here.

johnxnguyen: Then I would suggest the title be the same as the for the password prompt (`app_lock_locked_title`), and the description be `Unlock using biometric authentication.`

There's no need to say "click cancel and...", since it is the only other option and is most certainly what the user would do if they don't want to use biometrics.

app/src/main/scala/com/waz/zclient/conversationpager/SecondPageFragment.scala
```diff
@@ -47,13 +47,15 @@ class SecondPageFragment extends FragmentHelper
   private lazy val userAccountsController = inject[UserAccountsController]
   private lazy val conversationController = inject[ConversationController]
 
-  override def setUserVisibleHint(isVisibleToUser: Boolean): Unit = {
+  // TODO: The method is deprecated.  If it's necessary to inform Android that SecondPageFragment is invisible,
+  // try ViewPager.addOnPageChangeListener: https://stackoverflow.com/a/25822814/2975925
+/*  override def setUserVisibleHint(isVisibleToUser: Boolean): Unit = {
```
johnxnguyen: Is this being addressed in an other PR?

app/src/main/scala/com/waz/zclient/security/SecurityPolicyChecker.scala
```diff
@@ -94,94 +77,98 @@ class SecurityPolicyChecker(implicit injector: Injector, ec: EventContext) exten
...
 
+  private def wipeOnCookieInvalid(accountManager: AccountManager)(implicit context: Context) =
     if (BuildConfig.WIPE_ON_COOKIE_INVALID) {
       verbose(l"check WIPE_ON_COOKIE_INVALID")
+      val check = new CookieValidationCheck(accountManager.auth)
+      val actions = List(new WipeDataAction(Some(accountManager.userId)))
+
+      val a = Option('a)
+      val b = Option('b)
+      val c = a.zip(b)
```
johnxnguyen: What does this do?

app/src/main/res/values/strings.xml
makingthematrix: I want it to be a more generic dialog which we can use in different places.
But yes, instead of adding new generic strings, I can take them from the `apply` method, provided by the code which uses the dialog.

app/src/main/scala/com/waz/zclient/conversationpager/SecondPageFragment.scala
makingthematrix: It will have to be, but so far I in testing I didn't find what was the problem it supposed to be fixing. I don't want to forget about it, but I don't want to address it right now.

app/src/main/scala/com/waz/zclient/security/SecurityPolicyChecker.scala
makingthematrix: removed

app/src/main/res/values/strings.xml
johnxnguyen: So I understand `request_password_title` isn't being used anymore? If so then it should be deleted.

app/src/main/res/values/strings.xml
```diff
@@ -1391,7 +1400,7 @@
     <string name="read_receipts_remotely_changed_message">You can change this option in your account settings.</string>
     
     <string name="app_lock_locked_title">Wire is locked due to inactivity</string>
-    <string name="app_lock_locked_message">Please enter your device credentials.</string>
+    <string name="app_lock_locked_message">Please enter your account password</string>
```
johnxnguyen: This needs a full stop at the end because it's a message and not a title.

app/src/main/res/values/strings.xml
johnxnguyen: Are you intending to change these strings?

makingthematrix: done

app/src/main/res/values/strings.xml
makingthematrix: deleted

app/src/main/scala/com/waz/zclient/BaseActivity.scala
```diff
@@ -71,71 +70,35 @@ class BaseActivity extends AppCompatActivity
...
-    if (dpm.isActivePasswordSufficient) {
-      verbose(l"current password is sufficient")
-      inject[SecurityPolicyChecker].authenticateIfNeeded(activity)
-    }
-    else {
-      verbose(l"current password is insufficient")
-      startActivity(new Intent(DevicePolicyManager.ACTION_SET_NEW_PASSWORD))
-    }
+    if (BuildConfig.BLOCK_ON_PASSWORD_POLICY)
+      SecurityPolicyService.checkAdminEnabled(dpm, secPolicy, ContextUtils.getString(R.string.security_policy_description)(this))(this)
```
gizemb: I think there's a chance that onCreate will be called after onActivityResult -if the system resources are low. If that's true, we'll be going through this SecurityPolicyService checks twice. Can you check this scenario by enabling "Don't keep activities" setting?

app/src/main/scala/com/waz/zclient/conversationpager/SecondPageFragment.scala
gizemb: Can we open an issue on Jira?

app/src/main/scala/com/waz/zclient/preferences/dialogs/RequestPasswordDialog.scala
```diff
@@ -18,86 +18,185 @@
...
             true
           case _ => false
         }
     })
   }
+
+  private def sendPassword(password: String): Unit = {
+    onPassword ! Password(password)
+    KeyboardUtils.closeKeyboardIfShown(getActivity)
+    dismiss()
```
gizemb: can this be the reason your dialog is being dismissed immediately?

app/src/main/scala/com/waz/zclient/preferences/dialogs/RequestPasswordDialog.scala
```diff
@@ -18,86 +18,185 @@
...
+
+  private lazy val callback: BiometricPrompt.AuthenticationCallback = new BiometricPrompt.AuthenticationCallback {
+    override def onAuthenticationError(errorCode: Int, errString: CharSequence): Unit = {
+      super.onAuthenticationError(errorCode, errString)
+      verbose(l"onAuthenticationError, code: $errorCode, str: $errString")
+      // the cancellation code is 13; I couldn't find a constant for it
+      onBiometric ! (if (errorCode == 13) BiometricAnswerCancelled else BiometricAnswerError(errString.toString))
+      prompt.cancelAuthentication()
+    }
+
```
gizemb: I think it would be more readable if all the fields would be at the top of the class and then the methods started.

app/src/main/scala/com/waz/zclient/preferences/dialogs/RequestPasswordDialog.scala
```diff
@@ -18,86 +18,185 @@
...
+    .build
+
+  private lazy val executor: ExecutorWrapper = ExecutorWrapper(Threading.Ui)
+
+  private lazy val callback: BiometricPrompt.AuthenticationCallback = new BiometricPrompt.AuthenticationCallback {
+    override def onAuthenticationError(errorCode: Int, errString: CharSequence): Unit = {
+      super.onAuthenticationError(errorCode, errString)
+      verbose(l"onAuthenticationError, code: $errorCode, str: $errString")
+      // the cancellation code is 13; I couldn't find a constant for it
+      onBiometric ! (if (errorCode == 13) BiometricAnswerCancelled else BiometricAnswerError(errString.toString))
```
gizemb: I think it's BiometricConstants.ERROR_NEGATIVE_BUTTON but that class is restricted to api so we can't see. 
We can create a constant here ourselves. It's better than commenting I think :)

app/src/main/scala/com/waz/zclient/preferences/dialogs/RequestPasswordDialog.scala
```diff
@@ -18,86 +18,185 @@
...
+      dialog.setArguments(returning(new Bundle()) { b =>
+        b.putString(TitleArg, title)
+        message.foreach(b.putString(MessageArg, _))
+        biometricDesc.foreach(b.putString(BiometricDescriptionArg, _))
         error.foreach(b.putString(ErrorArg, _))
+        b.putBoolean(IsCancellable, isCancellable)
+        b.putBoolean(UseBiometric, onBiometric.isDefined)
       })
+      dialog.onPassword.onUi(onPassword)
+      onBiometric.foreach(f => dialog.onBiometric.onUi(f))
```
gizemb: I think it would be better to have these on some method of the Dialog class, and to use this method only as an instance creator.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/UserService.scala
```diff
@@ -282,6 +284,17 @@ class UserServiceImpl(selfUserId:        UserId,
     } yield resp
   }
 
+  // A hacky solution for now: we pretend to change the password to itself and interpret
+  // the error message "password must differ" as the correct one and all others as failures.
+  // Note that it's not possible to get `Right` as the response from BE.
```
gizemb: why did we need that? maybe we can change the implementation that returns the error, rather than changing the reaction to it in here.

app/config/checkstyle/suppressions.xml
```diff
@@ -33,6 +33,7 @@
     <suppress checks="." files="TouchImageView\.java" />
     <suppress checks="." files="CursorDragHelper\.java" />
     <suppress checks="." files="AppEntryError\.java" />
+    <suppress checks="." files="SecondPageFragment\.scala" />
```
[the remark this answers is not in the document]
makingthematrix: I commented out the deprecated code and I'll open a JIRA ticket for it. So, this line is not necessary.

app/src/main/scala/com/waz/zclient/BaseActivity.scala
```diff
@@ -71,71 +70,35 @@ class BaseActivity extends AppCompatActivity
...
-    dpm.setPasswordMinimumLowerCase(secPolicy, 1)
-    if (dpm.isActivePasswordSufficient) {
-      verbose(l"current password is sufficient")
-      inject[SecurityPolicyChecker].authenticateIfNeeded(activity)
-    }
-    else {
-      verbose(l"current password is insufficient")
-      startActivity(new Intent(DevicePolicyManager.ACTION_SET_NEW_PASSWORD))
-    }
+    if (BuildConfig.BLOCK_ON_PASSWORD_POLICY)
```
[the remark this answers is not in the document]
makingthematrix: "Block on password policy" means enforcing certain rules on the Android password. This depends now only on the build type.

app/src/main/scala/com/waz/zclient/preferences/dialogs/RequestPasswordDialog.scala
```diff
@@ -18,86 +18,185 @@
...
+      dialog.setCancelable(isCancellable)
     }
+    context.asInstanceOf[BaseActivity].getSupportFragmentManager
+      .beginTransaction
+      .add(fragment, RequestPasswordDialog.Tag)
+      .addToBackStack(RequestPasswordDialog.Tag)
+      .commit
```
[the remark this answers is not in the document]
makingthematrix: After the today's refactoring there should no such risk anymore: The dialog can be closed only as a reaction to typing in the password, using the biometric, or cancelling, and then it's dismissed, which means that it's also removed from the backstack. And since this is a blocking dialog, as long as it is open, no other instance of it can be opened.

app/src/main/scala/com/waz/zclient/preferences/dialogs/RequestPasswordDialog.scala
makingthematrix: Yes. I rewrote that part of the code.

app/src/main/scala/com/waz/zclient/preferences/dialogs/RequestPasswordDialog.scala
makingthematrix: Ok.

app/src/main/scala/com/waz/zclient/preferences/dialogs/RequestPasswordDialog.scala
makingthematrix: Changed to ERROR_NEGATIVE_BUTTON. Thanks!

app/src/main/scala/com/waz/zclient/preferences/dialogs/RequestPasswordDialog.scala
makingthematrix: I rewrote it to have only one function passed as a parameter which handles the user's actions and that function is now obligatory.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/UserService.scala
makingthematrix: I talked to BE about it. They agreed a new endpoint is needed - one that could be used simply to check the password. Right now, though, it doesn't exist. Every time we send a password typed by the user, it is in order to perform an action. So, one possible solution is to re-use the functionality for setting the password which is already in `UserService`. We can changed it when (if) BE creates a new endpoint.

PR #2398, opened 2019-11-01
Reorder sync handler updates

CalumMcCall:
### Issues

In `syncConversations`, we fetch a page of results, and then fetch the next page of results before updating the conversations with the results we just got. This means we could potentially only update the conversations all at once at the end. This might not scale well with older phones and lots of conversations.

### Solutions

This PR reorders the operations, so we update the current batch of conversations before moving onto the next one. This helps keep the number of futures we have scheduled at one time down.
#### APK
[Download build #309](http://10.10.124.11:8080/job/Pull%20Request%20Builder/309/artifact/build/artifact/wire-dev-PR2398-309.apk)


AndroidBob:

marcoconti83:
build again

PR #2436, opened 2019-11-21
Do not propagate status update when team size threshold is reached.

android10:
### Description

When user in big team (with lots of connections) changes status, it will send out a message to all connections, which could be too big and sending will fail (backend rejects etc.)

To avoid such issue, we decided in first phase not to send out status message for members in big teams, with more than 400 users.

When your team reaches the limit, you also stop displaying status of other users in your team.

## Jira Ticket
 - [AN-6512](https://wearezeta.atlassian.net/browse/AN-6512)

#### APK
[Download build #387](http://10.10.124.11:8080/job/Pull%20Request%20Builder/387/artifact/build/artifact/wire-dev-PR2436-387.apk)


marcoconti83:
build again

PR #2440, opened 2019-11-22
Big Teams hide user status: UI Part

android10:
### Disclaimer

This functionality will be rewritten at a later stage, therefore not too much thought has been put in the architecture.

### Notes

When user in big team (with lots of connections) changes status, it will send out a message to all connections, which could be too big and sending will fail (backend rejects etc.)

To avoid such issue, we decided in first phase not to send out status message for members in big teams, with more than 400 users.

When your team reaches the limit, you also stop displaying status of other users in your team.

### Description of this PR

As a continuation of #2436. This PR aims to cover the UI part, which is about hiding the status of every user if a team size is bigger than a threshold defined in the configuration. 

## Jira Ticket
 - [AN-6512](https://wearezeta.atlassian.net/browse/AN-6512)
#### APK
[Download build #401](http://10.10.124.11:8080/job/Pull%20Request%20Builder/401/artifact/build/artifact/wire-dev-PR2440-401.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
+++ b/app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
@@ -100,7 +100,10 @@ class SingleUserRowView(context: Context, attrs: AttributeSet, style: Int)
   def setUserData(userData: UserData, teamId: Option[TeamId], createSubtitle: (UserData) => String = SingleUserRowView.defaultSubtitle): Unit = {
     chathead.loadUser(userData.id)
     setTitle(userData.getDisplayName)
+
+    //TODO: Only display if we do not reach the Threshold
     if (teamId.isDefined) setAvailability(userData.availability)
+
     setVerified(userData.isVerified)
     setSubtitle(createSubtitle(userData))
     setIsGuest(userData.isGuest(teamId) && !userData.isWireBot)
--- a/app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
+++ b/app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
@@ -98,7 +98,8 @@ class SingleUserRowView(context: Context, attrs: AttributeSet, style: Int)
     videoIndicator.setVisibility(if (user.isVideoEnabled) View.VISIBLE else View.GONE)
   }
 
-  def setUserData(userData: UserData, teamId: Option[TeamId],
+  def setUserData(userData: UserData,
+                  teamId: Option[TeamId],
                   hideStatus: Boolean,
                   createSubtitle: (UserData) => String = SingleUserRowView.defaultSubtitle): Unit = {
     chathead.loadUser(userData.id)
--- a/app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
+++ b/app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
@@ -24,6 +24,7 @@ import android.view.View.OnClickListener
 import android.view.{Gravity, View, ViewGroup}
 import android.widget.{CompoundButton, ImageView, LinearLayout, RelativeLayout}
 import androidx.appcompat.widget.AppCompatCheckBox
+import com.waz.log.BasicLogging.LogTag.DerivedLogTag
 import com.waz.model.{Availability, IntegrationData, TeamId, UserData}
 import com.waz.utils.events.{EventStream, SourceStream}
 import com.waz.utils.returning
@@ -39,7 +40,7 @@ import com.waz.zclient.{R, ViewHelper}
 import org.threeten.bp.Instant
 
 class SingleUserRowView(context: Context, attrs: AttributeSet, style: Int)
-  extends RelativeLayout(context, attrs, style) with ViewHelper with ThemedView {
+  extends RelativeLayout(context, attrs, style) with ViewHelper with ThemedView with DerivedLogTag {
   def this(context: Context, attrs: AttributeSet) = this(context, attrs, 0)
   def this(context: Context) = this(context, null, 0)
 
@@ -102,7 +103,7 @@ class SingleUserRowView(context: Context, attrs: AttributeSet, style: Int)
                   createSubtitle: (UserData) => String = SingleUserRowView.defaultSubtitle): Unit = {
     chathead.loadUser(userData.id)
     setTitle(userData.getDisplayName)
-    if (teamId.isDefined && !hideStatus) setAvailability(userData.availability)
+    setAvailability(if (teamId.isDefined && !hideStatus) userData.availability else Availability.None)
     setVerified(userData.isVerified)
     setSubtitle(createSubtitle(userData))
     setIsGuest(userData.isGuest(teamId) && !userData.isWireBot)
--- a/app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
+++ b/app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
@@ -97,13 +97,12 @@ class SingleUserRowView(context: Context, attrs: AttributeSet, style: Int)
     videoIndicator.setVisibility(if (user.isVideoEnabled) View.VISIBLE else View.GONE)
   }
 
-  def setUserData(userData: UserData, teamId: Option[TeamId], createSubtitle: (UserData) => String = SingleUserRowView.defaultSubtitle): Unit = {
+  def setUserData(userData: UserData, teamId: Option[TeamId],
+                  hideStatus: Boolean,
+                  createSubtitle: (UserData) => String = SingleUserRowView.defaultSubtitle): Unit = {
     chathead.loadUser(userData.id)
     setTitle(userData.getDisplayName)
-
-    //TODO: Only display if we do not reach the Threshold
-    if (teamId.isDefined) setAvailability(userData.availability)
-
+    if (teamId.isDefined && !hideStatus) setAvailability(userData.availability)
     setVerified(userData.isVerified)
     setSubtitle(createSubtitle(userData))
     setIsGuest(userData.isGuest(teamId) && !userData.isWireBot)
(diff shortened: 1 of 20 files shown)


app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
```diff
@@ -97,10 +98,12 @@ class SingleUserRowView(context: Context, attrs: AttributeSet, style: Int)
     videoIndicator.setVisibility(if (user.isVideoEnabled) View.VISIBLE else View.GONE)
   }
 
-  def setUserData(userData: UserData, teamId: Option[TeamId], createSubtitle: (UserData) => String = SingleUserRowView.defaultSubtitle): Unit = {
+  def setUserData(userData: UserData, teamId: Option[TeamId],
```
makingthematrix: Cosmetic: If you use many lines for a method declaration, please put each argument in its own line.

app/src/main/scala/com/waz/zclient/conversation/creation/AddParticipantsFragment.scala
```diff
@@ -249,15 +256,17 @@ case class AddParticipantsAdapter(usersSelected: SourceSignal[Set[UserId]],
   val onSelectionChanged = EventStream[(Either[UserId, (ProviderId, IntegrationId)], Boolean)]()
 
   (for {
-    teamId        <- inject[Signal[Option[TeamId]]]
+    teamId        <- teamId
     res           <- searchResults
     usersSelected <- usersSelected
     servsSelected <- servicesSelected
+    hideStatus <- Signal.future(TeamSize.shouldHideStatus(Signal.const(teamId), usersStorage))
```
makingthematrix: If you pass a signal as an argument (which I think is a bit ugly for non-private methods...) then you can simply use the `teamId` signal here. You don't have to get the value of the `teamId` signal and then wrap it into `Signal.const`:
```
for {
...
teamId <- teamId
...
hideStatus <- Signal.future(TeamSize.shouldHideStatus(teamSize, usersStorage))
} yield { ... }
```

app/src/main/scala/com/waz/zclient/conversationlist/adapters/ConversationListAdapter.scala
```diff
@@ -55,9 +64,14 @@ abstract class ConversationListAdapter
     * @param newItems the new data source.
     */
   protected def updateList(newItems: List[Item]): Unit = {
-    DiffUtil.calculateDiff(new DiffCallback(items.toList, newItems), false).dispatchUpdatesTo(this)
-    items.clear()
-    items.appendAll(newItems)
+    TeamSize.shouldHideStatus(teamId, usersStorage).onSuccess {
```
makingthematrix: `onSuccess` is deprecated. Please use `foreach` instead.
And `case` is not necessary in `case hide =>`, since `hide` covers all possible options (`case` is usually used when you want to divide the reasoning into a few branches, depending what's inside the argument). `hide =>` is enough.
Also, curly braces are not needed here.

app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListRow.scala
```diff
@@ -241,6 +244,9 @@ class NormalConversationListRow(context: Context, attrs: AttributeSet, style: In
     avatar.clearImages()
     avatar.setAlpha(getResourceFloat(R.dimen.conversation_avatar_alpha_active))
     conversationId.publish(Some(conversationData.id), Threading.Ui)
+    (if (hideStatus) Future.successful(Availability.None) else controller.availability(conversationData.id).head).onComplete { av =>
```
makingthematrix: This is a bit difficult to read...
1. I don't think you need to return a future here. The method returns `Unit`. You can write `if (!hideStatus) ...` and focus only on the part after the current `else`.
2. Please use `foreach` instead of `onComplete`.

app/src/main/scala/com/waz/zclient/cursor/MentionCandidatesAdapter.scala
```diff
@@ -22,7 +22,7 @@ import android.view.View.OnClickListener
 import android.view.{LayoutInflater, View, ViewGroup}
 import com.waz.model.{TeamId, UserData}
 import com.waz.utils.events.{EventStream, SourceStream}
-import com.waz.zclient.R
+import com.waz.zclient.{R}
```
makingthematrix: `import com.waz.zclient.R` should have stayed like this.

app/src/main/scala/com/waz/zclient/participants/ParticipantsAdapter.scala
```diff
@@ -147,6 +149,13 @@ class ParticipantsAdapter(userIds: Signal[Seq[UserId]],
 
   private val conv = convController.currentConv
 
+  private var hideUserStatus = false
+  TeamSize.shouldHideStatus(team, usersStorage).onSuccess {
```
makingthematrix: `foreach` instead of `onSuccess`.
`hide =>` instead of `case (hide) =>`
`{}` can be deleted, it does nothing
You can also import an implicit `Threading.Ui` instead of stating it explicitly after each future:
```
import Threading.Implicits.Ui
```
(but please put that line **inside** the class, not in the global import statements at the top of the file.

app/src/main/scala/com/waz/zclient/usersearch/SearchUIAdapter.scala
```diff
@@ -108,6 +113,11 @@ class SearchUIAdapter(adapterCallback: SearchUIAdapter.Callback)(implicit inject
       updateMergedResults()
   }
 
+  private var hideUserStatus = false
+  TeamSize.shouldHideStatus(Signal.const(team.map(_.id)), usersStorage).onSuccess {
```
makingthematrix: same as above

app/src/main/scala/com/waz/zclient/views/ConversationFragment.scala
```diff
@@ -202,6 +205,10 @@ class ConversationFragment extends FragmentHelper {
 
     convController.currentConvName.onUi { updateTitle }
 
+    TeamSize.shouldHideStatus(accountsController.teamId, usersStorage).onSuccess {
```
makingthematrix: same as above

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/TeamSize.scala
```diff
@@ -0,0 +1,65 @@
...
+import com.waz.content._
+import com.waz.log.BasicLogging.LogTag.DerivedLogTag
+import com.waz.model._
+import com.waz.threading.Threading
+import com.waz.utils.events.{EventContext, Signal}
+
+import scala.concurrent.Future
+
+
+trait TeamSize {
```
makingthematrix: Can we call it `TeamSizeHack`?

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/TeamSize.scala
```diff
@@ -0,0 +1,65 @@
...
+      case _ => Future.successful({})
+    }
+}
+
+object TeamSize extends DerivedLogTag {
+
+  val statusPropagationThreshold = 400
+  import Threading.Implicits.Background
+
+  def membersCount(teamId: Option[TeamId], usersStorage: UsersStorage): Future[Option[Int]] = {
```
makingthematrix: Can `membersCount` be private?

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/TeamSize.scala
```diff
@@ -0,0 +1,65 @@
...
+    val empty: Future[Option[Int]] = Future.successful(None)
+    teamId.fold(empty)({ id => usersStorage.getByTeam(Set(id)).map(users => Some(users.size))})
+  }
+
+  def isAboveStatusPropagationThreshold(teamId: Option[TeamId], usersStorage: UsersStorage): Future[Boolean] =
+    membersCount(teamId, usersStorage).map { maybeSize => maybeSize.fold(false)(_ >= statusPropagationThreshold) }
```
makingthematrix: So, we want to return `false` if `maybeSize` is `None` and otherwise we want to check if its values is greater than or equal to `statusPropagationThreshold`. There is a method for that: `exists`: `maybeSize.exists(_ >= statusPropagationThreshold)`. You can also change `maybeSize => maybeSize.exists(...)` to `_.exists(...)` but I admit it may make the code a bit unreadable.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/TeamSize.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  def shouldHideStatus(teamId: Signal[Option[TeamId]], usersStorage: Signal[UsersStorage]): Future[Boolean] =
+    for {
+      teamId <- teamId.head
```
makingthematrix: Please indent to align `<-` with each other.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/TeamSize.scala
marcoconti83: I would prefer not to. What is the advantage of describing this as a hack in the name? Eventually, either we remove it (which we can't unless we want to remove the functionality) or we rewrite it to be more elegant.

app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListRow.scala
```diff
@@ -241,6 +238,14 @@ class NormalConversationListRow(context: Context, attrs: AttributeSet, style: In
     avatar.clearImages()
     avatar.setAlpha(getResourceFloat(R.dimen.conversation_avatar_alpha_active))
     conversationId.publish(Some(conversationData.id), Threading.Ui)
+    if (hideStatus) {
+      AvailabilityView.hideAvailabilityIcon(title)
+    } else {
+      controller.availability(conversationData.id).head.map({ s => Some(s) }).foreach {
```
makingthematrix: This `map` here should be deleted. `foreach { status => ... }` is enough.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/TeamSize.scala
makingthematrix: `TeamSize` suggests... well, I'm not sure. If I didn't know what it is, it would be confusing for me that we have a trait called like that which is not really team size itself, but some functionality used only for displaying the availability icon.

makingthematrix: `TeamSizeThreshold`?

Issue #431, opened 2016-12-31
Typing indicator not very reliable (?)


someoneEsle:
Unfortunately I don't have concrete evidence for what I'm about to report, but I hope I won't sound like I'm starting a conspiracy theory - and hopefully other people can comment with their own experience.
I feel like the typing indicator isn't very.. truthful. Like, it has happened many times that while wire says that "X is typing", X is not typing at all. In fact they are not only not looking at the phone, but in some cases they are not even connected to the internet. This issue seems to occur when you are having a "rather active" conversation with someone: at some point you'll be staring at the screen waiting for your friend to send this loooong message, just to find out in a few seconds that they are not typing at all.
Useful information:
- I'm on Android 7.1, Nexus 5X, I've been experiencing this issue since I downloaded wire (months ago)
- my friends are using their phones, not the web app or desktop clients. I haven't noticed if this happens more often or only with iOS/Android users -- I feel like it happens with both but I'll keep an eye on it.

mykola-mokhnach:
Is this still the case?

someoneEsle:
Yes, still experience this on a daily basis

someoneEsle:
This should have been fixed in the latest iOS release. I'll close for now, reopen if it happens again

PR #2446, opened 2019-11-26
Fix availability not updating in conversation list

marcoconti83:
## What's new in this PR?

Fix availability not updating in the conversation list on last candidate

### Issues

The signal to update the availability was removed, and the availability was set only when the method `setConversation` was called.

The signal has been restored, and combined with a signal that informs the view when the availability should stop being displayed because the threshold has been reached.

### Testing

Tested manually in teams above and below the threshold

#### APK
[Download build #431](http://10.10.124.11:8080/job/Pull%20Request%20Builder/431/artifact/build/artifact/wire-dev-PR2446-431.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListRow.scala
+++ b/app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListRow.scala
@@ -34,7 +34,7 @@ import com.waz.service.call.CallInfo
 import com.waz.service.call.CallInfo.CallState.SelfCalling
 import com.waz.threading.Threading
 import com.waz.utils._
-import com.waz.utils.events.Signal
+import com.waz.utils.events.{Signal, SourceSignal}
 import com.waz.zclient.calling.CallingActivity
 import com.waz.zclient.calling.controllers.{CallController, CallStartController}
 import com.waz.zclient.common.controllers.UserAccountsController
@@ -54,7 +54,6 @@ import com.waz.zclient.utils.ContextUtils._
 import com.waz.zclient.utils.{ConversationSignal, StringUtils, UiStorage, UserSetSignal, UserSignal, ViewUtils}
 import com.waz.zclient.views.AvailabilityView
 import com.waz.zclient.{R, ViewHelper}
-
 import scala.collection.Set
 
 trait ConversationListRow extends View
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListRow.scala
```diff
@@ -96,6 +95,8 @@ class NormalConversationListRow(context: Context, attrs: AttributeSet, style: In
   val menuIndicatorView = ViewUtils.getView(this, R.id.conversation_menu_indicator).asInstanceOf[MenuIndicatorView]
 
   var conversationData = Option.empty[ConversationData]
+  val hideAvailability: SourceSignal[Boolean] = Signal(false)
```
makingthematrix: this can be `private` and you don't have to declare the type explicitly.

app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListRow.scala
```diff
@@ -174,12 +175,26 @@ class NormalConversationListRow(context: Context, attrs: AttributeSet, style: In
     }
   }
 
+  // Conversation title
```
makingthematrix: Since now setting the name is separated from setting the availability, and `conversationName` is a signal used only once, here, you can merge them. I can show you how 1-to-1.

app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListRow.scala
```diff
@@ -174,12 +175,26 @@ class NormalConversationListRow(context: Context, attrs: AttributeSet, style: In
...
     title.setText(name)
   }
 
+  // User availability (for 1:1)
+  (for {
+    Some(id)  <- conversationId
+    av            <- controller.availability(id)
+    hide          <- hideAvailability
+  } yield(av, hide)).onUi { case (av, hide) =>
+    if (hide) Availability.None else av match {
```
makingthematrix: This does nothing for `hide == 0` - it only returns `Availability.None`, but does not hide the icon.
Instead of if/else here a more Scala-idiomatic solution is to have separate cases:
```
case (av, true)             => AvailabilityView.hideAvailabilityIcon(title)
case (Availability.None, _) => AvailabilityView.hideAvailabilityIcon(title)
case (av, _)                => AvailabilityView.displayLeftOfText(title, av, title.getCurrentTextColor, pushDown = true)
```

Issue #484, opened 2017-01-16
Better Ping/Call Behavior


0xdade:
I've noted this on twitter previously, but thought opening an actual ticket would be more useful to everyone.

We have a group chat of about 50 users that is regularly active. Unfortunately, on many phones the ping button is very close to the home button. Similarly, the call button, while generally far enough away from everything, is still easy to mistakenly hit.

My friend @aagallag built an [XPosed module](https://github.com/b0tchsec/Make-Wire-Great-Again) to disable outgoing calls and pings, but we've discussed better methods as well. The preferred solution, I think, is to provide a popup confirmation dialog when sending a ping or starting a call. 

This means that when accidentally hitting either button, it doesn't automatically bug the rest of the chat room, but doesn't remove the ability to send pings or calls entirely from large chats.

tnguyen14:
I've having an issue with constantly hitting the Ping button by mistake as well.

On the iOS app, the ping button is part of the More menu (the one behind the 3 dots), which makes its use a bit more intentional. Can this be done on Android?

The idea of the confirmation dialog could work quite well for the call button. It might be a bit inconsistent for ping though.

Either way, having ping less easily accessible might be better.

bjoernherbig:
Thanks for the feedback. We have some changes planned that will make it easier to choose the order. No timeline to share yet though.

Gardosen:
i am closing this ticket now as it is over 3 years old and there was no response between now and then.

Kind regards
Marco from the Android Team

PR #2449, opened 2019-11-27
Feature: Authorization headers for Retrofit requests

gizemb:
### Issues

Adds Authorization header and token refresh mechanism for Retrofit requests.

### Testing

Since Settings page blocks changing accounts for now, could not test on multiple accounts.

#### APK
[Download build #444](http://10.10.124.11:8080/job/Pull%20Request%20Builder/444/artifact/build/artifact/wire-dev-PR2449-444.apk)
[Download build #457](http://10.10.124.11:8080/job/Pull%20Request%20Builder/457/artifact/build/artifact/wire-dev-PR2449-457.apk)
[Download build #484](http://10.10.124.11:8080/job/Pull%20Request%20Builder/484/artifact/build/artifact/wire-dev-PR2449-484.apk)
[Download build #496](http://10.10.124.11:8080/job/Pull%20Request%20Builder/496/artifact/build/artifact/wire-dev-PR2449-496.apk)


app/src/main/scala/com/waz/zclient/utils/AuthTokenObserver.scala
```diff
@@ -0,0 +1,41 @@
...
+import com.waz.threading.Threading
+import com.waz.utils.events.{EventContext, Signal}
+import com.waz.zclient.user.data.source.remote.AuthHeaderInterceptor
+import com.waz.zclient.{Injectable, Injector}
+
+class AuthTokenObserver(implicit injector: Injector, ec: EventContext) extends Injectable with DerivedLogTag {
+
+  private lazy val accountManager = inject[Signal[AccountManager]]
+
+  private val authToken = accountManager.flatMap(_.accessToken).onChanged
```
makingthematrix: This is an event stream, so it will be updated for the first time only after the first change. It won't give you the initial value. Is this ok?

gizemb: removed `onChanged`

app/src/main/scala/com/waz/zclient/utils/AuthTokenObserver.scala
```diff
@@ -0,0 +1,51 @@
...
+class AuthTokenObserver(implicit injector: Injector, ec: EventContext)
+    extends Injectable
+    with DerivedLogTag {
+
+  private lazy val accountManager = inject[Signal[AccountManager]]
+
+  private val authToken = accountManager.flatMap(_.accessToken)
+
+  //TODO: might observe accessToken and tokenType separately
+  authToken.on(Threading.Background) {
```
makingthematrix: Cosmetic: You can replace `.on(Threading.Background)`, and `Threading.Background` in the futures below, with `import Threading.Implicits.Background` put inside the class.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/AccountManager.scala
```diff
@@ -338,6 +338,14 @@ class AccountManager(val userId:   UserId,
     }
 
   }
+
+  def refreshToken(): Future[Boolean] = for {
```
makingthematrix: Cosmetic: It's the same as:
```
auth.refreshToken().map {
      case Left(_) => false
      case Right(_) => true
}
```

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/client/AuthenticationManager.scala
```diff
@@ -99,22 +106,30 @@ class AuthenticationManager(id: UserId,
...
-      case token => cookie.flatMap { cookie =>
-        debug(l"Non existent or potentially expired token: $token, will attempt to refresh with cookie: $cookie")
-        dispatchRequest(client.access(cookie, token)) {
-          case Left(resp @ ErrorResponse(ResponseCode.Forbidden | ResponseCode.Unauthorized, message, label)) =>
-            verbose(l"access request failed (label: ${showString(label)}, message: ${showString(message)}), will try login request. currToken: $token, cookie: $cookie, access resp: $resp")
-            tracking.exception(new RuntimeException(s"Access request failed: msg: $message, label: $label, cookie expired at: ${cookie.expiry} (is valid: ${cookie.isValid}), currToken expired at: ${token.map(_.expiresAt)} (is valid: ${token.exists(_.isValid)})"), null)
-            logout(reason = InvalidCookie).map(_ => Left(resp))
-        }
-      }
+      case token => refreshToken(token)
```
makingthematrix: This `token` here is always `None`. It's better to make it explicit.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/client/AuthenticationManager.scala
```diff
@@ -241,22 +256,26 @@ class AuthenticationManager2(id: UserId, accStorage: AccountStorage2, client: Lo
...
-      case token => cookie.flatMap { cookie =>
-        debug(l"Non existent or potentially expired token: $token, will attempt to refresh with cookie: $cookie")
-        dispatchRequest(client.access(cookie, token)) {
-          case Left(resp @ ErrorResponse(ResponseCode.Forbidden | ResponseCode.Unauthorized, message, label)) =>
-            verbose(l"access request failed (label: ${showString(label)}, message: ${showString(message)}, will try login request. currToken: $token, cookie: $cookie, access resp: $resp")
-            tracking.exception(new RuntimeException(s"Access request failed: msg: $message, label: $label, cookie expired at: ${cookie.expiry} (is valid: ${cookie.isValid}), currToken expired at: ${token.map(_.expiresAt)} (is valid: ${token.exists(_.isValid)})"), null)
-            wipeCredentials().map(_ => Left(resp))
-        }
-      }
+      case token => refreshToken(token)
```
makingthematrix: Same here. Please change to `case None => refreshToken(None)`.

gizemb: it might also be Some, but expired I guess?

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/AccountManager.scala
gizemb: wow! done :)

app/src/main/scala/com/waz/zclient/utils/AuthTokenObserver.scala
gizemb: I couldn't make the first one go away but I added for the ones below

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/client/AuthenticationManager.scala
gizemb: it can fall back to this case if it is Some but expired

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/client/AuthenticationManager.scala
gizemb: same

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
```diff
@@ -47,12 +47,6 @@ open class Network {
     fun userApi(): UserApi {
         return retrofit.create(UserApi::class.java)
     }
-
-    companion object {
```
android10: why the change from a class to an object? Do we want for this to behave as a singleton? Or we want to use its static methods?

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
```diff
@@ -1,20 +1,20 @@
...
 
-open class Network {
 
-    internal var retrofit: Retrofit
+object Network {
 
-    init {
-        retrofit = createNetworkClient(BASE_URL)
-    }
+    private const val BASE_URL = "https://staging-nginz-https.zinfra.io"
```
android10: I see this constant is used only here (and it is private), you can either inline it or move it inside the ```createNetworkClient()``` function and change the line:

```kotlin
private val retrofit: Retrofit by lazy { createNetworkClient() }
```

app/src/main/kotlin/com/waz/zclient/user/data/source/remote/AuthHeaderInterceptor.kt
```diff
@@ -0,0 +1,51 @@
...
+
+import android.annotation.SuppressLint
+import io.reactivex.subjects.PublishSubject
+import okhttp3.Interceptor
+import okhttp3.Response
+import java.util.concurrent.atomic.AtomicBoolean
+
+object AuthHeaderInterceptor : Interceptor {
+
+    private const val ERROR_CODE_UNAUTHORIZED = 401
```
android10: Maybe extract this to NetworkError enum or sealed class for safety? Sealed class are a good way to wrap an control errors and also an alternative as unsafe Strings.

app/src/main/kotlin/com/waz/zclient/user/data/source/remote/AuthHeaderInterceptor.kt
```diff
@@ -0,0 +1,51 @@
...
+    override fun intercept(chain: Interceptor.Chain): Response {
+        val response = chain.proceed(getRequest(chain))
+        if (!retryRequired(response)) {
+            return response
+        }
+
+        if (retryInProgress.compareAndSet(false, true)) {
+            waitForRetry.onNext(true)
+        }
+        waitForRetry.blockingFirst()
```
android10: As you pointed out, this is a bit hacky but if it is the only way we have for now to communicate with scala, let's keep it this way. Let's also make sure we document it for refactoring it.

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
gizemb: yes, we might right now, but in the future it'll be either staging or prod url, so we're gonna need some url parameter in createNetworkClient method. That's why I kept that as it is. Can change it though?

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
gizemb: I think it's better to create the whole retrofit set up etc. once, and keep it throughout the app lifecycle since we might be sending requests etc all the time. Does that make sense to you?

app/src/main/kotlin/com/waz/zclient/user/data/source/remote/AuthHeaderInterceptor.kt
gizemb: We're not throwing errors here right now so I don't see the point in wrapping it. This is just for http error code checking. Changed the code in here a little bit so let's continue this discussion on the new implementation.

app/src/main/kotlin/com/waz/zclient/user/data/source/remote/AuthHeaderInterceptor.kt
gizemb: [Here](https://github.com/wireapp/android-project/issues/430) it is.

gizemb:
build again

app/build.gradle
```diff
@@ -426,6 +429,10 @@ dependencies {
     testImplementation("$rootProject.deps.scalaTest") {
         exclude module: 'scala-library'
     }
+    testImplementation "com.squareup.okhttp3:mockwebserver:$rootProject.ext.versions.okHttp"
+    testImplementation 'org.robolectric:robolectric:4.3'
```
android10: For consistency on what you did above, do not hard code this value.

app/build.gradle
```diff
@@ -426,6 +429,10 @@ dependencies {
...
+    testImplementation 'org.robolectric:shadows-multidex:4.3'
```
android10: Same here

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
android10: Fair enough in this case. But mostly I'm more on the side of not thinking too much in the future (we can change it later) because things change so fast and we tend to overthink because of potential functionality that "could" be implemented.

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
android10: Absolutely it makes sense to have this as a Singleton Instance because it might be expensive to create and beinf reused all over the place. 

Singletons are good as soon as you are not saving state (which may lead to unexpected behavior due to inconsistent state)  

My point was mostly that this is going to be responsibility of our dependency injector, which we current not have. But as you know, this would be a class with an annotated ```@Singleton``` constructor.

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
```diff
@@ -1,20 +1,28 @@
...
+
+    private val retrofit: Retrofit by lazy { createNetworkClient(BASE_URL) }
+
+    @JvmStatic
+    val authRetryDelegate: AuthRetryDelegate by lazy { AuthRetryDelegateImpl() }
+
+    @JvmStatic
+    val authHeaderInterceptor by lazy { AuthHeaderInterceptor(authRetryDelegate)}
 
     private fun createNetworkClient(baseUrl: String): Retrofit {
```
android10: this can be renamed to ```createClient()``` since it is also part of the ```Network``` object. Just for readability.

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
```diff
@@ -1,20 +1,28 @@
...
+object Network {
 
-    init {
-        retrofit = createNetworkClient(BASE_URL)
-    }
+    private const val BASE_URL = "https://staging-nginz-https.zinfra.io"
+
+    private val retrofit: Retrofit by lazy { createNetworkClient(BASE_URL) }
+
+    @JvmStatic
```
android10: Any special reason for making this a static field? I also see only one usage so it can be private?

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
```diff
@@ -1,20 +1,28 @@
...
+    val authRetryDelegate: AuthRetryDelegate by lazy { AuthRetryDelegateImpl() }
+
+    @JvmStatic
+    val authHeaderInterceptor by lazy { AuthHeaderInterceptor(authRetryDelegate)}
```
android10: Same here for the ```@JvmStatic``` and and the visibility of this ```val```

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
```diff
@@ -44,15 +52,9 @@ open class Network {
...
-        return retrofit.create(UserApi::class.java)
-    }
+    fun userApi(): UserApi = createApi(UserApi::class.java)
 
-    companion object {
-        private const val BASE_URL = "https://staging-nginz-https.zinfra.io"
-        //Hardcoded just for testing
-        private const val API_TOKEN = "h5bEcw4ab4gnRWycVvn96YTkaN2P6lBpxYGjw-XQVYtsIPP1HI3W_kjD02OUydDjt_KZ9kp_OVdbjlrRK8VAAQ==.v=1.k=1.d=1575460548.t=a.l=.u=aa4e0112-bc8c-493e-8677-9fde2edf3567.c=15566912256171294211"
-    }
+    fun <T> createApi(apiClass: Class<T>): T = retrofit.create(apiClass)
```
android10: There is also only one usage for this: so maybe you can inline it and remove the generic function. It would be something like this:
```kotlin
fun userApi(): UserApi = retrofit.create(UserApi::class.java)
```

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
```diff
@@ -23,10 +31,10 @@ open class Network {
...
                 val newRequest = chain.request().newBuilder()
                     .addHeader("Content-Type", "application/json")
-                    .addHeader(
-                        "Authorization", "Bearer $API_TOKEN").build()
+                    .build()
                 chain.proceed(newRequest)
             }
+            .addInterceptor(authHeaderInterceptor)
             .addInterceptor(HttpLoggingInterceptor().apply {
                 level = if (BuildConfig.DEBUG) {
```
android10: Maybe you can change this code to more kotlin style. Something we should definitely keep in mind. 
```kotlin
level = when (BuildConfig.DEBUG) {
    true -> HttpLoggingInterceptor.Level.BODY
    false -> HttpLoggingInterceptor.Level.NONE 
}
```

More readable and concise right? :)

app/src/main/kotlin/com/waz/zclient/user/data/source/remote/AuthHeaderInterceptor.kt
```diff
@@ -0,0 +1,33 @@
+package com.waz.zclient.user.data.source.remote
+
+import android.annotation.SuppressLint
+import androidx.annotation.VisibleForTesting
+import okhttp3.Interceptor
+import okhttp3.Response
+
+@VisibleForTesting
```
android10: If we are making something visible only for the sake of testing maybe we have a problem. This is just a code smell. I would encapsulate the value where it belongs and in the case of tests just hard code it. The thing is if someone modifies the constant or get rid of the value the test will fail anyway.

android10: This constant belongs to ```AuthHeaderInterceptor```

app/src/main/kotlin/com/waz/zclient/user/data/source/remote/AuthHeaderInterceptor.kt
```diff
@@ -0,0 +1,33 @@
...
+const val AUTH_HEADER_TEXT = "Authorization"
+class AuthHeaderInterceptor(private val authRetryDelegate: AuthRetryDelegate) : Interceptor {
+
+    var token: String? = null
```
android10: This is very prone to Errors and maybe a good idea is to encapsulate this. Something like this:
```data class AuthToken(val token: String, val tokenType) 
```
Then you can pass this class to the ```AuthHeaderInterceptor``` as a collaborator. 

Both fields are related to each and it seems the work together. You can also mock this class for testing purpose.

app/src/main/kotlin/com/waz/zclient/user/data/source/remote/AuthHeaderInterceptor.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    var tokenType: String? = null
+
+    @SuppressLint("CheckResult")
```
android10: I did some refactor and this should look something like this:
```kotlin
class AuthHeaderInterceptor(private val authRetryDelegate: AuthRetryDelegate,
                            private val authToken: AuthToken): Interceptor {

    @SuppressLint("CheckResult")
    override fun intercept(chain: Interceptor.Chain): Response {
        val response = chain.proceed(getRequest(chain))
        if (!authRetryDelegate.retryRequired(response)) {
            return response
        }
        authRetryDelegate.startRetryProcess()
        authRetryDelegate.blockingWaitForRetryResult()

        return chain.proceed(getRequest(chain))
    }

    private fun getRequest(chain: Interceptor.Chain) = chain.request().newBuilder().apply {
        addHeader("Authorization", "${authToken.tokenType} ${authToken.token}")
        
    }.build()
}

data class AuthToken(val token: String = "", val tokenType: String = "")
```

Check how concise it is.

android10: We need to make sure that the code works but the idea is to avoid nulls as much as we can. Instead pass empty strings. Little by little we will add extension functions because for example in this default empty value: "", I do not like the string literals at all, so soon I would like to have something like this:

```kotlin
String.empty
```

android10:
Closing. This is deprecated. Job done on #2468 and #2484

Issue #504, opened 2017-01-19
Show footer for last message


someoneEsle:
Current behavior is that the footer is shown for the last *outgoing* message. However I feel like it would be better to show a footer for the *last message*, regardless of whether it is incoming or outgoing. Was this an intentional decision? If so, what is the reason behind it?

DOFfactory:
+1

someoneEsle:
There is a "+1" emoji to show your interest. Do not post meaningless comments please, it only adds noise (and emails) to the conversations. Also, using the +1 emoji is more useful as the developer can sort the issues by the number of +1s they received. Thanks!

PR #503, opened 2017-01-19
[AN-4890] Fix weblink preview issue

JPonte:
If there was text surrounding the link, the preview would replace the
link in the middle of the text. Now it adds it only at the end of the
message view.
Ticket: https://wearezeta.atlassian.net/projects/AN/issues/AN-4890
#### APK
[Download build #8308](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8308/artifact/build/artifact/wire-dev-PR503-8308.apk)
[Download build #8313](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8313/artifact/build/artifact/wire-dev-PR503-8313.apk)
[Download build #8314](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8314/artifact/build/artifact/wire-dev-PR503-8314.apk)

PR #502, opened 2017-01-19
Ids for collection items

JPonte:
#### APK
[Download build #8306](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8306/artifact/build/artifact/wire-dev-PR502-8306.apk)
[Download build #8307](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8307/artifact/build/artifact/wire-dev-PR502-8307.apk)

Issue #485, opened 2017-01-16
Finding users by username not working


0xdade:
## Unexpected behavior
For instance, I was trying to add @\alan earlier (without the slash) by username, at his request. When I went to do so, I ended up having to type his full name before he showed up in the list of people to add. Screenshot below.

![@alan](https://0xda.de/img/wire-alan.jpg)

## Expected behavior
When using someone's @ handle to add them, they should show up at the top of the list, perhaps in their own designated section to differentiate between the intended user and the "suggested" users.

mykola-mokhnach:
Can you please try it in the most recent version?

makingthematrix:
Fixed https://github.com/wireapp/wire-android/pull/2414 (or even earlier).

PR #2451, opened 2019-11-29
Devices screen: co-routines [WIP] 

BradleyWire:
### Issues
Kotlin migration of the device screens 

### Solutions

MVVM, Coroutines, LiveData 

### Testing

Click on settings > Devices and it should display a list of devices for now 

#### APK
[Download build #453](http://10.10.124.11:8080/job/Pull%20Request%20Builder/453/artifact/build/artifact/wire-dev-PR2451-453.apk)
[Download build #455](http://10.10.124.11:8080/job/Pull%20Request%20Builder/455/artifact/build/artifact/wire-dev-PR2451-455.apk)
[Download build #458](http://10.10.124.11:8080/job/Pull%20Request%20Builder/458/artifact/build/artifact/wire-dev-PR2451-458.apk)
[Download build #459](http://10.10.124.11:8080/job/Pull%20Request%20Builder/459/artifact/build/artifact/wire-dev-PR2451-459.apk)
[Download build #460](http://10.10.124.11:8080/job/Pull%20Request%20Builder/460/artifact/build/artifact/wire-dev-PR2451-460.apk)
[Download build #461](http://10.10.124.11:8080/job/Pull%20Request%20Builder/461/artifact/build/artifact/wire-dev-PR2451-461.apk)
[Download build #462](http://10.10.124.11:8080/job/Pull%20Request%20Builder/462/artifact/build/artifact/wire-dev-PR2451-462.apk)
[Download build #469](http://10.10.124.11:8080/job/Pull%20Request%20Builder/469/artifact/build/artifact/wire-dev-PR2451-469.apk)
[Download build #471](http://10.10.124.11:8080/job/Pull%20Request%20Builder/471/artifact/build/artifact/wire-dev-PR2451-471.apk)
[Download build #477](http://10.10.124.11:8080/job/Pull%20Request%20Builder/477/artifact/build/artifact/wire-dev-PR2451-477.apk)
[Download build #485](http://10.10.124.11:8080/job/Pull%20Request%20Builder/485/artifact/build/artifact/wire-dev-PR2451-485.apk)
[Download build #502](http://10.10.124.11:8080/job/Pull%20Request%20Builder/502/artifact/build/artifact/wire-dev-PR2451-502.apk)
[Download build #503](http://10.10.124.11:8080/job/Pull%20Request%20Builder/503/artifact/build/artifact/wire-dev-PR2451-503.apk)
[Download build #504](http://10.10.124.11:8080/job/Pull%20Request%20Builder/504/artifact/build/artifact/wire-dev-PR2451-504.apk)


app/src/main/kotlin/com/waz/zclient/devices/data/source/remote/ClientsRemoteDataSourceImpl.kt
```diff
@@ -4,14 +4,14 @@ import com.waz.zclient.core.data.source.remote.SafeApiDataSource
 import com.waz.zclient.core.resources.Resource
 import com.waz.zclient.devices.data.model.ClientEntity
 
-class ClientsRemoteDataSourceImpl(private var deviceNetwork: ClientsNetwork = ClientsNetwork())
+class ClientsRemoteDataSourceImpl(private val clientsApi: ClientsApi)
```
android10: Avoid using **Impl** naming convention, this is just a ```ClientsRemoteDataSource``` and if we are calling ```ClientsRemoteDataSourceImpl``` it means there is an interface called ```ClientsRemoteDataSource``` and we reach the anti-pattern: one abstraction and one implementation, which means that given this case we probably do not need the interface if we only have one implementation.

Here is more info: https://stackoverflow.com/questions/2814805/java-interfaces-implementation-naming-convention

app/src/main/kotlin/com/waz/zclient/devices/data/ClientsRepositoryImpl.kt
```diff
@@ -20,7 +21,7 @@ class ClientsRepositoryImpl(private val remoteDataSource: ClientsRemoteDataSourc
         @VisibleForTesting
         internal var instance: ClientsRepositoryImpl? = null
```
android10: instance is not a good name for the subject under test by convention I would say just: ```clientRepository```

app/src/main/kotlin/com/waz/zclient/devices/data/ClientsRepositoryImpl.kt
```diff
@@ -20,7 +21,7 @@ class ClientsRepositoryImpl(private val remoteDataSource: ClientsRemoteDataSourc
         @VisibleForTesting
```
android10: This is code smell: we should not change the visibility or scope of something for the sake of testing, maybe we need to rethink the approach.

BradleyWire: Could potentially have a destroyInstance method and make the instance object private? I'd assume this is temporary until we decide on a DI framework (if any)

app/src/main/kotlin/com/waz/zclient/core/data/source/remote/SafeApiDataSource.kt
```diff
@@ -0,0 +1,29 @@
+package com.waz.zclient.core.data.source.remote
+
+import com.waz.zclient.core.resources.Resource
+import retrofit2.Response
+import timber.log.Timber
+
+abstract class SafeApiDataSource {
+
+    protected suspend fun <T> getRequestResult(responseCall: suspend () -> Response<T>): Resource<T> {
```
android10: Maybe rename this method to ```requestResult()``` since the ```get`` prefix comes from the old java days and is considered deprecated in new generation programming languages.

app/src/main/kotlin/com/waz/zclient/core/data/source/remote/SafeApiDataSource.kt
```diff
@@ -0,0 +1,29 @@
...
+        try {
+            val response = responseCall()
+            if (response.isSuccessful) {
+                val body = response.body()
+                if (body != null) {
+                    return Resource.success(body)
+                }
+            }
+            return error(" ${response.code()} ${response.message()}")
+        } catch (e: Exception) {
```
android10: I'm interested on how we are doing error propagation through different layers. 
In this case, the caller or client of this class should know about the potential exception without a clear contract. 

Maybe we can use a more functional approach here: ```Either<L, R>```

```Either<L, R>``` is referred as a disjoint function, which means that this structure is designed to hold either a Left<T> or Right<T> value but never both. It is a functional programming monadic type not yet existent in the Kotlin Standard Library but I implemented mine some time ago:

```kotlin
/**
 * Represents a value of one of two possible types (a disjoint union).
 * Instances of [Either] are either an instance of [Left] or [Right].
 * FP Convention dictates that: 
 *      [Left] is used for "failure".
 *      [Right] is used for "success".
 *
 * @see Left
 * @see Right
 */
sealed class Either<out L, out R> {
    /** * Represents the left side of [Either] class which by convention is a "Failure". */
    data class Left<out L>(val a: L) : Either<L, Nothing>()
    /** * Represents the right side of [Either] class which by convention is a "Success". */
    data class Right<out R>(val b: R) : Either<Nothing, R>()

    val isRight get() = this is Right<R>

    val isLeft get() = this is Left<L>

    fun either(fnL: (L) -> Any, fnR: (R) -> Any): Any =
            when (this) {
                is Either.Left -> fnL(a)
                is Either.Right -> fnR(b)
            }

    fun <T> flatMap(fn: (R) -> Either<L, T>): Either<L, T> {...}
    fun <T> map(fn: (R) -> (T)): Either<L, T> {...}
}
```

I also wrote about it some time ago, check the error handling section:
https://fernandocejas.com/2018/05/07/architecting-android-reloaded/

android10: And we propagate ```Either``` all over the place and the client always get either a success (Right) or failure(Left) which forces it to check for the result and treat the error properly or process the success :)

BradleyWire: So, the current way we're propagating errors is that the remote data source would catch the exception if we get one as such and then it'll send back a Resource object back with an error status. If we don't catch that at the remote level i.e. we use a different remote data source or local one, it'd then be caught at the domain layer inside the use-cases. It's one of the pitfalls of using co-routines is that you can suspend your function and would need to catch the exception that comes through that. 

I very much like the look of using an Either and then letting the presentation layer determine the loading logic as then we can just pass back the same way we're doing it now. I guess it doesn't open up the class for abuse (with too many responsibilities) in the future. 

Interested to hear other's thoughts on this.

app/src/main/kotlin/com/waz/zclient/core/resources/Resource.kt
```diff
@@ -1,12 +1,12 @@
 package com.waz.zclient.core.resources
 
-data class Resource<T>(
+data class Resource<out T>(
```
android10: What is a Resource? If it is something that we are using for Api Requests, maybe we can pick a friendlier name such as Result or a Response.

app/src/main/kotlin/com/waz/zclient/devices/data/ClientsRepository.kt
```diff
@@ -0,0 +1,11 @@
+package com.waz.zclient.devices.data
+
+import com.waz.zclient.core.resources.Resource
+import com.waz.zclient.devices.domain.model.Client
+
+interface ClientsRepository {
+
+    suspend fun getClientById(clientId: String): Resource<Client>
```
android10: Rename these to:
```clientById()```
```allClients()```

File: app/src/main/kotlin/com/waz/zclient/devices/data/ClientsRepository.kt [hunk 8]
```diff
@@ -0,0 +1,11 @@
...
+
+    suspend fun getAllClients(): Resource<List<Client>>
```
android10: This would look better when reading the code: ```Response<List<Client>>```

app/src/main/kotlin/com/waz/zclient/core/usecase/coroutines/UseCase.kt
```diff
@@ -0,0 +1,20 @@
...
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.async
+import kotlinx.coroutines.launch
+
+abstract class UseCase<out Type, in Params> where Type : Any {
+
+    abstract suspend fun run(params: Params): Resource<Type>
+
+    open operator fun invoke(
+        scope: CoroutineScope,
```
android10: I like this. Correct me if I'm wrong: To keep track of the coroutines and cancel them all when appropriate.

app/src/main/kotlin/com/waz/zclient/devices/data/ClientsRepositoryImpl.kt
```diff
@@ -0,0 +1,31 @@
...
+
+    override suspend fun getAllClients(): Resource<List<Client>> = remoteDataSource.getAllClients().toDomainList()
+
+    companion object {
+
+        @Volatile
+        @VisibleForTesting
+        internal var instance: ClientsRepositoryImpl? = null
+
+        fun getInstance(remoteDataSource: ClientsRemoteDataSource = ClientsRemoteDataSourceImpl(ClientsNetwork().getClientsApi())): ClientsRepository =
```
android10: I guess this is temporal until we use a DI Framework for injecting this as a ```@Singleton``` right?

app/src/main/kotlin/com/waz/zclient/devices/data/source/local/ClientPropertiesDataSource.kt
```diff
@@ -0,0 +1,6 @@
+package com.waz.zclient.devices.data.source.local
+
+interface ClientPropertiesDataSource {
+
+    fun getSelfClientId(): String
```
android10: rename to: ```selfClientId()```

app/src/main/kotlin/com/waz/zclient/devices/data/source/local/ClientPropertiesDataSourceImpl.kt
```diff
@@ -0,0 +1,9 @@
+package com.waz.zclient.devices.data.source.local
+
+class ClientPropertiesDataSourceImpl : ClientPropertiesDataSource {
```
android10: If we have only one implementation of this, I would get rid of the interface

app/src/main/kotlin/com/waz/zclient/devices/data/source/remote/ClientsApi.kt
```diff
@@ -0,0 +1,15 @@
...
+
+import com.waz.zclient.devices.data.model.ClientEntity
+import retrofit2.Response
+import retrofit2.http.GET
+import retrofit2.http.Path
+
+interface ClientsApi {
+
+    @GET("/clients/{clientId}")
+    suspend fun getClientByIdAsync(@Path("clientId") clientId: String): Response<ClientEntity>
```
android10: maybe ```clientById()```? Why the async suffix? I guess that is implicit since all api request should be asynchronous. 
```

File: app/src/main/kotlin/com/waz/zclient/devices/data/source/remote/ClientsApi.kt [hunk 14]
```diff
@@ -0,0 +1,15 @@
...
+
+    @GET("/clients")
+    suspend fun getAllClientsAsync(): Response<Array<ClientEntity>>
```
android10: ```allClients()```

android10: Check all the places where the ```get``` suffix exist as part of a function name :)

app/src/main/kotlin/com/waz/zclient/devices/domain/GetAllClientsUseCase.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.waz.zclient.core.usecase.coroutines.UseCase
+import com.waz.zclient.devices.data.ClientsRepository
+import com.waz.zclient.devices.domain.model.Client
+import timber.log.Timber
+
+class GetAllClientsUseCase(private val clientsRepository: ClientsRepository)
+    : UseCase<List<Client>, Unit>() {
+
+    override suspend fun run(params: Unit): Resource<List<Client>> {
+        return try {
```
android10: We should avoid using this try catch as much as possible, that is why I was suggesting the usage of ```Either<L, R>```.

We may end up eating the exception if the caller is not aware of the possible check. This approach is very prone to "hard to find bugs" since we loose awareness but let's open a discussion: 
https://softwareengineering.stackexchange.com/questions/365427/try-catch-log-rethrow-is-anti-pattern

app/src/main/kotlin/com/waz/zclient/devices/data/ClientsRepositoryImpl.kt
BradleyWire: Yes, this would be temporary.

app/src/main/kotlin/com/waz/zclient/core/usecase/coroutines/UseCase.kt
BradleyWire: So we have a handy extension which comes with the kotlin ktx library where we can define a scope and it'll clear that scope when the ViewModel is cleared.

app/src/main/kotlin/com/waz/zclient/devices/data/ClientsRepository.kt
BradleyWire: I've migrated this over to Either<L, R> now

PR #2458, opened 2019-12-05
Crash Controller Investigation [WIP] 

BradleyWire:
### Issues

Application was getting an ANR on syncing. 

### Causes

Catching exceptions on the UI thread 

### Solutions

Remove it completely and find all the crashes it has been eating

### Testing

regression will be needed. 
#### APK
[Download build #514](http://10.10.124.11:8080/job/Pull%20Request%20Builder/514/artifact/build/artifact/wire-dev-PR2458-514.apk)


app/src/main/scala/com/waz/zclient/appentry/AppEntryActivity.scala
```diff
@@ -317,7 +317,7 @@ class AppEntryActivity extends BaseActivity {
       .setTransition(FragmentTransaction.TRANSIT_FRAGMENT_FADE)
       .add(new CountryDialogFragment, CountryDialogFragment.TAG)
       .addToBackStack(CountryDialogFragment.TAG)
-      .commit
+      .commitAllowingStateLoss
```
gizemb: is losing the state here acceptable for us?

app/src/main/scala/com/waz/zclient/appentry/AppEntryActivity.scala
```diff
@@ -334,7 +334,7 @@ class AppEntryActivity extends BaseActivity {
     transaction
       .replace(R.id.fl_main_content, f, tag)
       .addToBackStack(tag)
-      .commit
+      .commitAllowingStateLoss
```
gizemb: same here?

PR #2466, opened 2019-12-09
Preare the terrain for the Network Re-Write

android10:
## What's new in this PR?

The idea Is to rewrite the networking of the Android Client. This PR aims to prepare the terrain in order to make it happen. 

- Kluent Fluent Assertion library for kotlin was added. 
- Either<L,R> class added.

#### APK
[Download build #552](http://10.10.124.11:8080/job/Pull%20Request%20Builder/552/artifact/build/artifact/wire-dev-PR2466-552.apk)

### DIFF
--- a/app/src/main/kotlin/com/waz/zclient/core/functional/Either.kt
+++ b/app/src/main/kotlin/com/waz/zclient/core/functional/Either.kt
@@ -1,0 +1,102 @@
+/**
+ * Copyright (C) 2019 Fernando Cejas Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.waz.zclient.core.functional
+
+/**
+ * Represents a value of one of two possible types (a disjoint union).
+ * Instances of [Either] are either an instance of [Left] or [Right].
+ * FP Convention dictates that [Left] is used for "failure"
+ * and [Right] is used for "success".
+ *
+ * @see Left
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/com/waz/zclient/core/functional/Either.kt
```diff
@@ -0,0 +1,102 @@
+/**
+ * Copyright (C) 2019 Fernando Cejas Open Source Project
```
gizemb: Can you update your copyrights? 🙂

app/src/test/kotlin/com/waz/zclient/core/functional/EitherTest.kt
```diff
@@ -0,0 +1,65 @@
...
+
+import com.waz.zclient.UnitTest
+import com.waz.zclient.core.functional.Either.Left
+import com.waz.zclient.core.functional.Either.Right
+import org.amshove.kluent.shouldBe
+import org.amshove.kluent.shouldBeInstanceOf
+import org.amshove.kluent.shouldEqualTo
+import org.junit.Test
+
+class EitherTest : UnitTest() {
```
gizemb: inheritance in tests might slow down the tests.
https://www.petrikainulainen.net/programming/unit-testing/3-reasons-why-we-should-not-use-inheritance-in-our-tests/

app/src/main/kotlin/com/waz/zclient/core/functional/Either.kt
android10: As discussed offline, we need to add copyright to the original author every time we use code from someone else. We use a GPL License. 

Maybe @marcoconti83 can build on this?

app/src/test/kotlin/com/waz/zclient/core/functional/EitherTest.kt
android10: The first rule of performance is measurement to assert that something not obvious could have an impact on performance. 

In the example you pointed out, it is a bit misleading since it is inheriting from from a class that has access to a DB which is IO, which means being slow, even though  I do not see any numbers that demonstrates that (they are assumptions).

In this case, the main purpose is to reuse methods and avoid repeating ourselves with ```Mockito.initMocks``` for example. There is no logic involved. 

Favor composition over inheritance for reusing logic and this case there is no logic involved, only configuration sharing.

android10:
build again

Issue #680, opened 2017-03-04
Caller A hears loud echo in voice calls when caller B uses Nexus 6P Android 7.1.1


fortran77:
I have a predictable situation of caller A hearing a loud distracting echo when caller B uses a Nexus 6P (Android 7.1.1) for a voice call. In the experiments described below, caller A always uses a smartphone. The voice calls were voice only, with no video.

When B uses the Nexus 6P with Wire 2.28.317, A hears a loud distracting echo, and B hears significant distortion.

When B uses a computer (an iMac running OS X 10.10), with Wire (version 2017-02-24) running in a web browser tab encapsulated in the Rambox software, audio quality is excellent for both A and B.

When B uses the Nexus 6P and both callers are using WhatsApp for a voice call, audio quality is excellent for both A and B.

When B uses the Nexus 6P and both callers are using conventional cellphone voice service, voice quality is satisfactory for both A and B.

So it appears that the culprit is B's nexus 6P, when running Wire, but not when running WhatsApp or using the cellphone voice network.

The echo and distortion are independent of who initiates the voice call.

mykola-mokhnach:
@jspittka FYI

mykola-mokhnach:
@fortran77 Thanks for your report. The issue has been investigated by the team and is expected to be fixed as soon as Android will pick up the updated version of the corresponding component in the future releases.

fortran77:
Issue #684 cited above mentions a "2.29 milestone". Presumably that means that a fix would be in version 2.9? I am not sure.

Anyway, Wire version 2.31.324, which I guess would be a later version than 2.29, on a Nexus 6P, is still causing the other party to hear an echo.

fortran77:
The current version 2.36.343 on a Nexus 6P still causes the other party to hear an echo.

fortran77:
The current version 2.38.352 on a Nexus 6P no longer causes the other party to hear an echo.

This is only after brief testing, but thank-you for the fix!

fortran77:
I reported "no longer causes the other party to hear an echo" previously, but that was an anomalous observation.

The echo issue has continued to exist on my Nexus 6P.

After various upgrades, I now have Wire 3.3.374 and Android 8.0.0 on this Nexus 6P.

Wire continues to work without any echo on an Android tablet and on a MacOS machine.

serussell:
[...]

PR #2437, opened 2019-11-22
Settings feature in Kotlin

mejdoo:
## What's new in this PR?

New Architecture: Writing Screen Settings in Kotlin
#### APK
[Download build #388](http://10.10.124.11:8080/job/Pull%20Request%20Builder/388/artifact/build/artifact/wire-dev-PR2437-388.apk)
[Download build #398](http://10.10.124.11:8080/job/Pull%20Request%20Builder/398/artifact/build/artifact/wire-dev-PR2437-398.apk)
[Download build #402](http://10.10.124.11:8080/job/Pull%20Request%20Builder/402/artifact/build/artifact/wire-dev-PR2437-402.apk)
[Download build #403](http://10.10.124.11:8080/job/Pull%20Request%20Builder/403/artifact/build/artifact/wire-dev-PR2437-403.apk)
[Download build #404](http://10.10.124.11:8080/job/Pull%20Request%20Builder/404/artifact/build/artifact/wire-dev-PR2437-404.apk)
[Download build #405](http://10.10.124.11:8080/job/Pull%20Request%20Builder/405/artifact/build/artifact/wire-dev-PR2437-405.apk)
[Download build #421](http://10.10.124.11:8080/job/Pull%20Request%20Builder/421/artifact/build/artifact/wire-dev-PR2437-421.apk)
[Download build #425](http://10.10.124.11:8080/job/Pull%20Request%20Builder/425/artifact/build/artifact/wire-dev-PR2437-425.apk)
[Download build #426](http://10.10.124.11:8080/job/Pull%20Request%20Builder/426/artifact/build/artifact/wire-dev-PR2437-426.apk)
[Download build #427](http://10.10.124.11:8080/job/Pull%20Request%20Builder/427/artifact/build/artifact/wire-dev-PR2437-427.apk)
[Download build #428](http://10.10.124.11:8080/job/Pull%20Request%20Builder/428/artifact/build/artifact/wire-dev-PR2437-428.apk)
[Download build #433](http://10.10.124.11:8080/job/Pull%20Request%20Builder/433/artifact/build/artifact/wire-dev-PR2437-433.apk)
[Download build #434](http://10.10.124.11:8080/job/Pull%20Request%20Builder/434/artifact/build/artifact/wire-dev-PR2437-434.apk)
[Download build #441](http://10.10.124.11:8080/job/Pull%20Request%20Builder/441/artifact/build/artifact/wire-dev-PR2437-441.apk)
[...]

PR #2468, opened 2019-12-09
Kotlin core network rewrite part 2

android10:
## What's new in this PR?

 - This is the second part of the Networking layer. Many changes overlap with the current work done with the new Architecture. Check the comments, because there is a bunch of stuff that is needed temporarily in order to not break our build. It is being refactored in #2484 

### Testing

 - A couple of things have been tested but mostly here there are interfaces and abstractions. Check #2484 for more implementation details. 

#### APK
[Download build #678](http://10.10.124.11:8080/job/Pull%20Request%20Builder/678/artifact/build/artifact/wire-dev-PR2468-678.apk)
[Download build #680](http://10.10.124.11:8080/job/Pull%20Request%20Builder/680/artifact/build/artifact/wire-dev-PR2468-680.apk)


android10:
build again

app/src/main/kotlin/com/waz/zclient/core/network/ApiService.kt
```diff
@@ -0,0 +1,45 @@
...
+import com.waz.zclient.core.exception.Failure
+import com.waz.zclient.core.exception.Failure.*
+import com.waz.zclient.core.extension.failFastIfUIThread
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.functional.Either.Left
+import com.waz.zclient.core.functional.Either.Right
+import retrofit2.Call
+import retrofit2.Response
+
+internal class ApiService(private val networkHandler: NetworkHandler) {
```
gizemb: if we make this internal, then all the service classes will have to be in the network module. I think having the services & apis in their feature modules would be better. 
we don't need to change it right now but i'm commenting anyway as a reminder that we should discuss the module structure later.

app/src/main/kotlin/com/waz/zclient/di/Injector.kt
```diff
@@ -0,0 +1,41 @@
...
+        if (BuildConfig.DEBUG) {
+            val loggingInterceptor = HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BASIC)
+            okHttpClientBuilder.addInterceptor(loggingInterceptor)
+        }
+        return okHttpClientBuilder.build()
+    }
+
+    fun networkHandler() = NetworkHandler(context())
+
+    fun context(): Context = WireApplication.APP_INSTANCE().applicationContext
```
gizemb: this should be a compilation error since kotlin code cannot access scala code.

gizemb: you should make a Java class in between. sorry...

app/src/main/kotlin/com/waz/zclient/di/Injector.kt
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Temporary class for providing dependencies.
+ * TODO: Remove once a DI Framework is used.
+ */
+object Injector {
+
+    private const val BASE_URL = "https://staging-nginz-https.zinfra.io"
+    private const val API_TOKEN = "-HFiw9hJvFhu2djSIt1WZ0fwQg3J0a0RUFCQ3OXzXKB8yrJn9m1xcPsGXoQHDZjHCIk0bbuQwOnYam3cFbItDw==.v=1.k=1.d=1575391626.t=a.l=.u=4555f7b2-f97b-409f-8c3a-333a473ac1b9.c=2878560529346928087"
+
+    fun retrofit(): Retrofit {
```
gizemb: this can be a singleton I believe since there's no point in creating this for every service.

android10:
build again

gizemb:
build again

gizemb:
build again

app/src/main/kotlin/com/waz/zclient/core/di/Injector.kt
```diff
@@ -0,0 +1,50 @@
...
+import com.waz.zclient.ContextProvider
+import com.waz.zclient.ZApplication
+import com.waz.zclient.core.network.*
+import okhttp3.OkHttpClient
+import okhttp3.logging.HttpLoggingInterceptor
+import retrofit2.Retrofit
+import retrofit2.converter.gson.GsonConverterFactory
+
+/**
+ * Temporary class for providing dependencies.
```
android10: Avoid reviewing this class since with #2495 it is gone.

app/src/main/kotlin/com/waz/zclient/core/exception/Failure.kt
```diff
@@ -12,6 +12,12 @@ sealed class DatabaseFailure : Failure()
 
 object NetworkConnection : NetworkFailure()
 object NetworkServiceError : NetworkFailure()
+object ServerError : NetworkFailure()
```
android10: @BradleyWire we are going to fine tune this. For the time being we just rebased with what you did on #2493

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenAuthenticator.kt
```diff
@@ -0,0 +1,54 @@
...
+ * is delivered to the caller. This authenticator blocks all requests while a token
+ * refresh is being performed. In-flight requests that fail with a 401 (unauthorized)
+ * are automatically retried.
+ */
+class AccessTokenAuthenticator(private val authToken: AuthToken) : Authenticator {
+
+    /**
+     * This authenticate() method is called when server returns 401 Unauthorized.
+     */
+    override fun authenticate(route: Route?, response: Response): Request? {
```
android10: This logic is no longer valid and it is being changed in #2484 but we need this class to build on top of it.

app/src/main/kotlin/com/waz/zclient/core/network/ApiService.kt
```diff
@@ -0,0 +1,51 @@
...
+            when (response.isSuccessful) {
+                true -> Right(response.body() ?: default)
+                false -> handleRequestError(response)
+            }
+        } catch (exception: Throwable) {
+            Left(ServerError)
+        }
+    }
+
+    private fun <T> handleRequestError(response: Response<T>): Either<Failure, T> {
```
android10: This is going to merged with your work @BradleyWire on #2493

app/src/main/kotlin/com/waz/zclient/core/network/ClientsRepositorySample.kt
```diff
@@ -0,0 +1,44 @@
...
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.functional.map
+import com.waz.zclient.core.network.api.client.ClientsService
+import com.waz.zclient.features.clients.ClientEntity
+
+/**
+ * This demonstrates the usage of the Network API.
+ * TODO: Remove this when used and implemented somewhere else.
+ */
+abstract class ClientsRepository(private val remoteDataSource: ClientsRemoteDataSource) {
```
android10: Avoid reviewing this class since it is only an Example of usage for the framework being written. 
Feel free to provide feedback dough.

app/src/main/kotlin/com/waz/zclient/features/clients/ClientEntity.kt
```diff
@@ -0,0 +1,31 @@
+package com.waz.zclient.features.clients
```
android10: This class is going to be removed in exchange for ```ClientApi``` created in #2437

app/src/main/kotlin/com/waz/zclient/core/di/Injector.kt
```diff
@@ -0,0 +1,50 @@
...
+        if (BuildConfig.DEBUG) {
+            val loggingInterceptor = HttpLoggingInterceptor().setLevel(HttpLoggingInterceptor.Level.BASIC)
+            okHttpClientBuilder.addInterceptor(loggingInterceptor)
+        }
+        return okHttpClientBuilder.build()
+    }
+
+    fun networkHandler() = NetworkHandler(context())
+
+    fun context(): Context = ContextProvider.getApplicationContext()
```
mejdoo: ContextProvider is completely removed here #2495, we may face some conflicts later...

app/src/main/kotlin/com/waz/zclient/core/network/AuthToken.kt
```diff
@@ -0,0 +1,18 @@
+package com.waz.zclient.core.network
+
+class AuthToken(private val tokenRepository: AccessTokenRepository) {
```
mejdoo: What's the difference between AuthToken and AccessTokenRepository? What should be the entry point to connect with the NetworkService Layer?

app/src/main/kotlin/com/waz/zclient/di/Injector.kt
android10: With the DI framework we are very safe here :)

app/src/main/kotlin/com/waz/zclient/core/network/api/client/ClientsApi.kt
```diff
@@ -0,0 +1,18 @@
+package com.waz.zclient.core.network.api.client
+
+import com.waz.zclient.features.clients.ClientEntity
+import retrofit2.Call
+import retrofit2.http.GET
+import retrofit2.http.Path
+
+interface ClientsApi {
```
mejdoo: This class should be removed. Done in  #2477

app/src/main/kotlin/com/waz/zclient/core/network/api/client/ClientsService.kt
```diff
@@ -0,0 +1,19 @@
+package com.waz.zclient.core.network.api.client
+
+import com.waz.zclient.core.di.Injector
+import com.waz.zclient.core.exception.Failure
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.network.ApiService
+import com.waz.zclient.core.network.NetworkClient
+import com.waz.zclient.features.clients.ClientEntity
+
+class ClientsService(private val networkClient: NetworkClient) : ApiService(Injector.networkHandler()) {
```
mejdoo: Same here

app/src/test/kotlin/com/waz/zclient/core/network/ApiServiceTest.kt
```diff
@@ -0,0 +1,24 @@
+package com.waz.zclient.core.network
+
+import com.waz.zclient.UnitTest
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mock
+
+class ApiServiceTest : UnitTest() {
```
mejdoo: Could be removed if not finalised.

app/src/test/kotlin/com/waz/zclient/core/network/api/client/ClientsServiceTest.kt
```diff
@@ -0,0 +1,29 @@
+package com.waz.zclient.core.network.api.client
+
+import com.waz.zclient.UnitTest
+import com.waz.zclient.core.network.NetworkClient
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mock
+
+class ClientsServiceTest : UnitTest() {
```
mejdoo: Same here

app/src/main/kotlin/com/waz/zclient/core/di/Injector.kt
android10: This is our own implementation. For some reason we thought exactly the same way. This is gone with the next refactor. We did not have a DI framework here.

app/src/main/kotlin/com/waz/zclient/core/network/api/client/ClientsApi.kt
android10: This is duplicated so we do not break any changes. We want to integrate with the current work. This branch is already rebased from `develop`

app/src/main/kotlin/com/waz/zclient/core/network/AuthToken.kt
android10: We have a NetworkService but you are using the `Network` class. We need to decide which is a proper entry point. This is not the integration. This is the structure of the framework. For now we want to keep these names until we merge with what you have. This PR is not touching at all anything of what you have done, which in terms of networking was temporal as discussed.

app/src/test/kotlin/com/waz/zclient/core/network/ApiServiceTest.kt
android10: The reason is because we are doing most of the job in the next part #2484

app/src/test/kotlin/com/waz/zclient/core/network/api/client/ClientsServiceTest.kt
android10: I would never say to remove a test class :). At least I want to have as a reminder. No code should go to production without tests. If you do not feel comfortable with this approach I can remove it and add it in the next PR.

app/src/main/kotlin/com/waz/zclient/core/network/AuthToken.kt
mejdoo: I know this is not touching anything of the settings feature. I think it would be great to define how should we interact with NetworkService layer later... the name AccessTokenRepository is a bit confusing here. Is it possible to call this class from any Usecase from the domain layer? Basically, the network layer will have interactions only with datasources I guess.

gizemb: I am thinking of moving the all network logic to a separate module, and declaring these classes as `internal` so clients won't have access to authentication code. We can come up with other solutions too.

The examples of how to interact with the service layer will come in future PRs. We don't have any here since we _just_ merged the architecture part.

app/src/test/kotlin/com/waz/zclient/core/network/api/client/ClientsServiceTest.kt
gizemb: this is a test of a temporary sample class. Will remove that in future PRs, when the sample is also removed.

app/src/main/kotlin/com/waz/zclient/core/network/AuthToken.kt
android10: This is a networking framework. From a caller perspective you will never have to deal with it at such level. You only interact from the ```NetworkServices``` and the different ```ServiceApi``` you create. Keep in mind that Repository is a pattern, that abstract you from your data sources. This is a core component, for instance it has nothing to do with our current Clean Architecture approach we are taking at feature development level.

mejdoo: Ok it we would be great the move this to a separate module like "storage" module

android10: Before we merge the latest changes, You will see the interaction with the components. We are trying to come up with a friendly and easy to use ```api```.

android10: @gizemb what are the arguments of moving this to a separate module? I do not see it. If it is only for scoping then we should find another approach. I do not see as a valid reason to have it there.

mejdoo:
Let's merge it for now by the way and think about moving the network layer to a separate module in part 3 if possible.

app/src/main/kotlin/com/waz/zclient/core/network/AuthToken.kt
android10: @mejdoo I think @gizemb is talking about a separated android module, whereas you are talking about a separation in terms of DI modules? Is that correct?

mejdoo: No Im talking also about an android module. Let's discuss tomorrow about that.

PR #2471, opened 2019-12-10
Refactor: Delete unused authentication related classes

gizemb:
### Issues

AccountStorage2 and "new" AuthRequestInterceptor classes were not used in production code and were only accessed by tests. Removed these redundant classes and modifies test code to use mocks and classes used in production code.

### Causes

These 2nd generation classes were created to replace the old ones but never made it that far. Since we're planning to write the network layer from scratch, there is no need to enforce a migration for them, they can be basically deleted and we can move on with the old implementations.

### Testing

There's no change in production code. The unit tests are fixed to use the original implementations.

#### APK
[Download build #572](http://10.10.124.11:8080/job/Pull%20Request%20Builder/572/artifact/build/artifact/wire-dev-PR2471-572.apk)


gizemb:
build again

Issue #648, opened 2017-02-23
Automatically Unmute Muted Conversation


cwmke:
It would be nice to be able to mute a conversation for a specified time period instead of permanently. For example, if there is a group that is talking about something you're not interested in, you mute the conversation and an option pops up with a list of times (15min, 30min, 1hr, 6hr, 12hr, 1 day) you can mute for as well as an option to mute forever.

bjoernherbig:
thanks you for the suggestion, but this one is not going to be on the roadmap anytime soon. small team

cwmke:
No problem! I just wanted to add it to what I'm sure is an exceptionally long list of desired features 😃. I'm thrilled with the app and am happy to wait.

PR #2477, opened 2019-12-12
Feature: Settings with coroutines 

BradleyWire:
## What's new in this PR?

Kotlin migration of the settings screen using kotlin-coroutines, MVVM, LiveData
#### APK
[Download build #648](http://10.10.124.11:8080/job/Pull%20Request%20Builder/648/artifact/build/artifact/wire-dev-PR2477-648.apk)


app/src/main/kotlin/com/waz/zclient/core/config/Config.kt
```diff
@@ -0,0 +1,13 @@
+package com.waz.zclient.core.config
+
+import com.waz.zclient.BuildConfig
+
+
+object Config {
+    fun isDeveloperSettingsEnabled(): Boolean = BuildConfig.DEVELOPER_FEATURES_ENABLED
```
android10: Remove the ```is``` from all these methods. In this case ```developerSettingsEnabled()``` is meaningful enough.

app/src/main/kotlin/com/waz/zclient/core/extension/Activity.kt
```diff
@@ -0,0 +1,24 @@
+package com.waz.zclient.core.extension
+
+
+import androidx.appcompat.app.AppCompatActivity
+import androidx.fragment.app.Fragment
+
+
+fun AppCompatActivity.addFragment(frameId: Int, fragment: Fragment) {
```
android10: Try to avoid grouping statements between {} as much as possible (which is very javaish). Kotlin is very concise and this could be:

```kotlin
fun AppCompatActivity.addFragment(frameId: Int, fragment: Fragment) =
    supportFragmentManager.doTransaction { add(frameId, fragment) }

fun AppCompatActivity.replaceFragment(frameId: Int, fragment: Fragment, addToBackStack: Boolean) =
    supportFragmentManager.doTransaction {
        replace(frameId, fragment).apply {
            if (addToBackStack) {
                addToBackStack(fragment.tag)
            }
        }
    }

fun AppCompatActivity.removeFragment(fragment: Fragment) =
    supportFragmentManager.doTransaction { remove(fragment) }
```

app/src/main/kotlin/com/waz/zclient/core/extension/Context.kt
```diff
@@ -1,8 +1,14 @@
...
 import android.net.NetworkInfo
 
-val Context.networkInfo: NetworkInfo? get() =
-    (this.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager).activeNetworkInfo
+val Context.networkInfo: NetworkInfo?
+    get() = (this.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager).activeNetworkInfo
+
+fun Context.stringArrayFromResource(id: Int): Array<String> = resources.getStringArray(id)
+
+fun Context.stringFromResource(id: Int): String = resources.getString(id)
```
android10: The compilers infers the type here: ```String can be removed```. I personally only explicitly specify the type when required.

app/src/main/kotlin/com/waz/zclient/core/config/Config.kt
android10: Also you can get rid of all the types since they are inferred by the compiler:
```kotlin
    fun developerSettingsEnabled() = BuildConfig.DEVELOPER_FEATURES_ENABLED
    fun appLockForced() = BuildConfig.FORCE_APP_LOCK
    fun hideScreenContentForced()= BuildConfig.FORCE_HIDE_SCREEN_CONTENT
    fun versionName() = BuildConfig.VERSION_NAME
    fun websiteUrl() = BuildConfig.WEBSITE_URL
    fun accountsUrl() = BuildConfig.ACCOUNTS_URL
```

app/src/main/kotlin/com/waz/zclient/core/extension/Fragment.kt
```diff
@@ -0,0 +1,25 @@
...
+
+
+import android.content.Intent
+import android.net.Uri
+import android.os.Bundle
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.FragmentManager
+import androidx.fragment.app.FragmentTransaction
+
+inline fun FragmentManager.doTransaction(func: FragmentTransaction.() ->
```
android10: Some indentation here and conciseness:
```kotlin
inline fun FragmentManager.doTransaction(func: FragmentTransaction.() -> FragmentTransaction) =
    beginTransaction().func().commit()

inline fun <T : Fragment> T.withArgs(argsBuilder: Bundle.() -> Unit): T =
    this.apply {
        arguments = Bundle().apply(argsBuilder)
    }
```

app/src/main/kotlin/com/waz/zclient/core/extension/Fragment.kt
```diff
@@ -0,0 +1,25 @@
...
+    beginTransaction().func().commit()
+}
+
+inline fun <T : Fragment> T.withArgs(
+    argsBuilder: Bundle.() -> Unit): T =
+    this.apply {
+        arguments = Bundle().apply(argsBuilder)
+    }
+
+fun Fragment.openUrl(url: String) {
```
android10: You can inline the variable since it is used only once:
```kotlin
fun Fragment.openUrl(url: String) =
    requireActivity().startActivity(Intent(Intent.ACTION_VIEW, Uri.parse(url)))
```

app/src/main/kotlin/com/waz/zclient/core/functional/Either.kt
```diff
@@ -100,3 +100,5 @@ fun <L, R> Either<L, R>.getOrElse(value: R): R =
         is Either.Left -> value
         is Either.Right -> b
     }
+
+data class Failure(val message: String)
```
android10: This should not be here. There is already the ```Failure``` class in: ```package com.waz.zclient.core.exception```

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
```diff
@@ -0,0 +1,53 @@
+package com.waz.zclient.core.network
```
android10: This class is very likely to change because of the work being done by @gizemb and I. But good to have something up and running for the feature. I'm not being very strict with this one.

app/src/main/kotlin/com/waz/zclient/core/network/SafeApiDataSource.kt
```diff
@@ -0,0 +1,56 @@
+package com.waz.zclient.core.network
```
android10: Same as above. We will be refactoring this part and adapt our WIP networking framework.

app/src/main/kotlin/com/waz/zclient/devices/data/ClientsDataSource.kt
```diff
@@ -0,0 +1,12 @@
+package com.waz.zclient.devices.data
```
android10: This is a bit confusing. This looks more like a Repository rather than a Data Source.

What is this for?

Unless I'm missing something the natural thing to have would be an abstraction representing the Repository with multiple data sources acting as collaborators.

storage/src/main/kotlin/com/waz/zclient/storage/clients/migration/ClientsTableMigration.kt
```diff
@@ -0,0 +1,124 @@
+package com.waz.zclient.storage.clients.migration
+
+import android.content.ContentValues
+import androidx.room.OnConflictStrategy
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+import org.json.JSONObject
+
+class ClientsTableMigration : Migration(124, 125) {
```
gizemb: is it possible to test this migration? 😬

storage/src/main/kotlin/com/waz/zclient/storage/clients/migration/ClientsTableMigration.kt
```diff
@@ -0,0 +1,124 @@
...
+
+    }
+
+    private fun parseClientData(client: JSONObject?): ContentValues {
+        val contentValues = ContentValues()
+        client?.apply {
+            if (has(CLIENT_ID_KEY)) {
+                val clientId = getString(CLIENT_ID_KEY)
+                contentValues.put(CLIENT_ID_KEY, clientId)
+            }
```
gizemb: maybe we can extract this to a method

storage/src/main/kotlin/com/waz/zclient/storage/clients/migration/ClientsTableMigration.kt
```diff
@@ -0,0 +1,124 @@
...
+    companion object {
+        //Shared keys
+        private const val CLIENT_ID_KEY = "id"
+        private const val CLIENT_LABEL_KEY = "label"
+        private const val CLIENT_LOCATION_LAT_KEY = "lat"
+        private const val CLIENT_LOCATION_LONG_KEY = "lon"
+        private const val CLIENT_ENC_KEY = "encKey"
+        private const val CLIENT_MAC_KEY = "macKey"
+        private const val CLIENT_VERIFICATION_KEY = "verification"
+        private const val CLIENT_MODEL_KEY = "model"
```
gizemb: maybe we could use these keys in the ClientEntity class as well?

storage/src/main/kotlin/com/waz/zclient/storage/db/dao/UserDao.kt
```diff
@@ -0,0 +1,30 @@
...
+import androidx.room.Insert
+import androidx.room.Query
+import com.waz.zclient.storage.db.model.UserEntity
+import io.reactivex.Completable
+
+@Dao
+interface UserDao {
+
+    @Insert
+    fun insert(user: UserEntity): Completable
```
gizemb: there's an rx dependency here?

storage/src/main/kotlin/com/waz/zclient/storage/db/migration/UserDatabaseMigration.kt
```diff
@@ -0,0 +1,52 @@
+package com.waz.zclient.storage.db.migration
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+
+class UserDatabaseMigration : Migration(124, 125) {
```
gizemb: can we test this too?

storage/src/main/kotlin/com/waz/zclient/storage/db/migration/UserDatabaseMigration.kt
```diff
@@ -0,0 +1,52 @@
...
+        database.execSQL("CREATE TABLE IF NOT EXISTS `$USER_TABLE_NAME` (`_id` TEXT PRIMARY KEY NOT NULL, `teamId` TEXT , `name` TEXT ," +
+            " `email` TEXT , `phone` TEXT , `tracking_id` TEXT , `picture` TEXT , `accent` INTEGER , `skey` TEXT , `connection` TEXT , " +
+            "`conn_timestamp` INTEGER , `conn_msg` TEXT , `conversation` TEXT , `relation` TEXT , `timestamp` INTEGER , `display_name` TEXT ," +
+            " `verified` TEXT , `deleted` INTEGER NOT NULL , `availability` INTEGER , `handle` TEXT , `provider_id` TEXT , `integration_id` TEXT ," +
+            " `expires_at` INTEGER , `managed_by` TEXT , `self_permissions` INTEGER , `copy_permissions` INTEGER , `created_by` TEXT )")
+        database.execSQL("INSERT INTO $USER_TABLE_NAME SELECT * FROM $USERS_TABLE_NAME")
+        database.execSQL("DROP TABLE $USERS_TABLE_NAME")
+    }
+
+    companion object {
```
gizemb: again. I think we can use these constants in the Entity class

storage/src/main/kotlin/com/waz/zclient/storage/db/model/UserEntity.kt
```diff
@@ -0,0 +1,112 @@
...
+    @ColumnInfo(name = "self_permissions")
+    var selfPermission: Int?,
+
+    @ColumnInfo(name = "copy_permissions")
+    var copyPermission: Int?,
+
+    @ColumnInfo(name = "created_by")
+    var createdBy: String?
+) {
+    constructor() : this("", "", "", "", "", "", "",
```
gizemb: is this wise?

storage/src/main/kotlin/com/waz/zclient/storage/pref/GlobalPreferences.kt
```diff
@@ -0,0 +1,29 @@
+package com.waz.zclient.storage.pref
+
+import android.content.Context
+import android.content.SharedPreferences
+
+class GlobalPreferences(context: Context) {
+
+    private val sharedPreferences: SharedPreferences = context.getSharedPreferences("com.wire.preferences", Context.MODE_PRIVATE)
```
gizemb: why the "com.wire.preferences"?

storage/src/main/kotlin/com/waz/zclient/storage/pref/GlobalPreferences.kt
```diff
@@ -0,0 +1,29 @@
...
+
+    var activeUserId: String
+        get() = stringPreference(ACTIVE_ACCOUNT_GLOBAL_PREF_KEY)
+        set(value) = stringPreference(ACTIVE_ACCOUNT_GLOBAL_PREF_KEY, value)
+
+
+    private fun stringPreference(key: String, value: String) {
+        sharedPreferences.edit().putString(key, value).apply()
```
gizemb: can be an extension method to SharedPreferences

app/src/main/kotlin/com/waz/zclient/core/config/Config.kt
mejdoo: Fixed

app/src/main/kotlin/com/waz/zclient/core/extension/Context.kt
mejdoo: Fixed

app/src/main/kotlin/com/waz/zclient/core/extension/Activity.kt
mejdoo: I'll fix this

app/src/main/kotlin/com/waz/zclient/core/extension/Fragment.kt
mejdoo: I'll fix this

app/src/main/kotlin/com/waz/zclient/core/extension/Fragment.kt
mejdoo: I'll fix this

app/build.gradle
```diff
@@ -291,6 +297,7 @@ android {
         exclude 'META-INF/NOTICE'
         exclude 'APK_LICENSE.txt'
         exclude 'LICENSE.txt'
+        exclude 'META-INF/proguard/coroutines.pro'
```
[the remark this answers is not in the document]
mejdoo: I'll fix this

app/build.gradle
```diff
@@ -363,6 +372,15 @@ dependencies {
     implementation "androidx.exifinterface:exifinterface:1.0.0"
     implementation "androidx.media:media:1.0.0"
 
+    implementation "androidx.lifecycle:lifecycle-extensions:2.1.0"
+    implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.2.0-rc03'
+    implementation "androidx.lifecycle:lifecycle-livedata-ktx:2.2.0-rc03"
+    implementation "androidx.lifecycle:lifecycle-viewmodel-ktx:2.1.0"
```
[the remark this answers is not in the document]
mejdoo: I will use versions.lifecycle with the latest version 2.2.0-rc03

app/build.gradle
```diff
@@ -363,6 +372,15 @@ dependencies {
...
+    implementation "androidx.room:room-ktx:$rootProject.ext.versions.room"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:1.3.2"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.2"
+    testImplementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.2"
```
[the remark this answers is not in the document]
mejdoo: yes

app/build.gradle
```diff
@@ -422,6 +442,11 @@ dependencies {
     testImplementation("$rootProject.deps.scalaTest") {
         exclude module: 'scala-library'
     }
+    testImplementation 'org.mockito:mockito-core:3.1.0'
```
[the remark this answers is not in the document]
mejdoo: redundant I will remove it

storage/src/main/kotlin/com/waz/zclient/storage/db/model/UserEntity.kt
mejdoo: This constructor is required to fix a room compilation problem related to @Ignore annotation. For the 'picture' field, the endpoint return an array of strings and in the database it's a string. We would like to use the same models for networking and local store that's why we added this @Ignore annotation.

app/src/main/kotlin/com/waz/zclient/core/functional/Either.kt
BradleyWire: I'll fix this

app/build.gradle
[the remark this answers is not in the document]
mejdoo: For androidx.lifecycle:lifecycle-extensions I will keep the version 2.1.0. We need to update the Scala code because getFragmentManager is deprecated . It would be better to do it in separate PR.

app/src/main/kotlin/com/waz/zclient/core/functional/Either.kt
BradleyWire: So we've not propagated and error up to the UI to be less verbose. If you look into the Failure class you'll see i've added attributes to the ServerError to pass an error code and error message back to the ViewModel (to make error handling more clear and easier to debug). You'll also see that I've added a CancellationError object so that we know when to handle a cancelled coroutines scope.

app/build.gradle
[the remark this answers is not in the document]
BradleyWire: I would suggest as-long as it's not breaking changes to keep the scala code as is

app/build.gradle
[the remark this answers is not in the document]
BradleyWire: Just spoke to @gizemb and we're looking at a new way of accessing dependencies across modules. There will be a separate PR raised.

app/build.gradle
```diff
@@ -407,12 +425,14 @@ dependencies {
...
     implementation "com.squareup.retrofit2:converter-gson:$retrofit_version"
-    implementation "com.squareup.retrofit2:adapter-rxjava2:$retrofit_version"
-    implementation 'com.squareup.okhttp3:logging-interceptor:3.12.0'
+    implementation "com.squareup.okhttp3:logging-interceptor:$versions.okHttp"
+    testImplementation "com.squareup.okhttp3:mockwebserver:$versions.okHttp"
 
     //Android Architecture Components
     implementation "androidx.lifecycle:lifecycle-extensions:$versions.lifecycle"
 
+    api project(":storage")
```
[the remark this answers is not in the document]
BradleyWire: I had issues with implementation around accessing storage objects within tests

storage/src/main/kotlin/com/waz/zclient/storage/db/model/UserEntity.kt
BradleyWire: I think there must be a better way to approach this rather than adding empty strings to the overloaded constructor, maybe actually populate the fields with the values from the primary constructor?

app/src/main/kotlin/com/waz/zclient/core/network/SafeApiDataSource.kt
```diff
@@ -0,0 +1,56 @@
...
+
+suspend fun <T> requestApi(responseCall: suspend () -> Response<T>): Either<Failure, T> {
+    try {
+        val response = responseCall()
+        if (response.isSuccessful) {
+            val body = response.body()
+            if (body != null) {
+                return Either.Right(body)
+            }
+        }
```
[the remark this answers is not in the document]
BradleyWire: Nice, i'll fix this.

app/src/main/kotlin/com/waz/zclient/devices/data/source/local/ClientsLocalDataSource.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import com.waz.zclient.ContextProvider
+import com.waz.zclient.core.network.requestLocal
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.functional.Failure
+import com.waz.zclient.storage.clients.model.ClientEntity
+import com.waz.zclient.storage.db.UserDatabase
+import com.waz.zclient.storage.pref.GlobalPreferences
+
+class ClientsLocalDataSource {
```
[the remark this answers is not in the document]
BradleyWire: I've updated the tests and injected the dependencies

app/src/main/kotlin/com/waz/zclient/devices/mapper/ClientMapper.kt
```diff
@@ -0,0 +1,26 @@
...
+    time = time,
+    label = label,
+    _class = _class,
+    type = type,
+    id = id,
+    model = model,
+    location = ClientLocation(lon, lat)
+)
+
+fun ClientLocationEntity.toClient() = ClientLocation(
```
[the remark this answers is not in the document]
BradleyWire: Will fix this.

app/src/main/kotlin/com/waz/zclient/devices/data/source/local/ClientsLocalDataSource.kt
```diff
@@ -0,0 +1,33 @@
...
+
+    private val globalPreferences = GlobalPreferences(ContextProvider.getApplicationContext())
+    private val userId = globalPreferences.activeUserId
```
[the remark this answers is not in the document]
BradleyWire: Removed this from this class anyway

app/src/main/kotlin/com/waz/zclient/devices/data/ClientsDataSource.kt
mejdoo: Fixed for now : 
Repository ---> Interface
DataSource ---> Implementation

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
```diff
@@ -0,0 +1,90 @@
...
+
+    private fun initToolbar() {
+        activity?.title = getString(R.string.pref_account_screen_title)
+    }
+
+    private fun initViewModel() {
+        settingsAccountViewModel = ViewModelProvider(this, settingsViewModelFactory).get(SettingsAccountViewModel::class.java).also { viewModel ->
+            viewModel.loading.observe(viewLifecycleOwner, Observer { isLoading ->
+                updateLoadingVisibility(isLoading)
+            })
```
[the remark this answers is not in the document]
BradleyWire: Refactored for all fragments

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
```diff
@@ -0,0 +1,90 @@
...
+    private fun showErrorMessage(errorMessage: String) {
+        Toast.makeText(requireContext(), errorMessage, Toast.LENGTH_LONG).show()
+    }
+
+    private fun updateLoadingVisibility(isLoading: Boolean?) {
+        //Show hide progress indicator
+    }
+
+    private fun updateProfile(userProfileItem: UserProfileItem) {
+        preferences_account_name_title.text = userProfileItem.name
```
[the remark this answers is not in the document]
BradleyWire: I guess this is up to you when you write up the coding style guide, but, i'll offer my opinion here and day i'm not a massive fan of accessing the ids directly unless they're only used once. If we're using them multiple times I think it should be assigned to a view of some kind. As for camel case ids, I don't know.

app/src/main/kotlin/com/waz/zclient/settings/devices/detail/SettingsDeviceDetailFragment.kt
```diff
@@ -0,0 +1,76 @@
...
+
+    private fun bindError(errorMessage: String?) {
+        //Show error when we need to
+    }
+
+    private fun bindLoading(loading: Boolean?) {
+        //Show visibility of loading indicator
+    }
+
+    private fun bindDataToView(clientItem: ClientItem) {
```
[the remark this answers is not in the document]
BradleyWire: I don't think we should have the fragment knowing about the domain objects, the client item is used as a wrapper for the domain objects. It doesn't look pretty but it keeps atleast the fragment separated from the usecases.

app/src/main/kotlin/com/waz/zclient/settings/devices/detail/SettingsDeviceDetailFragment.kt
```diff
@@ -0,0 +1,76 @@
...
+        lifecycleScope.launchWhenResumed {
+            val id = arguments?.getString(SettingsDeviceConstants.DEVICE_ID_BUNDLE_KEY)
+            id?.let { deviceDetailsViewModel.loadData(it) }
+        }
+    }
+
+    companion object {
+        fun newInstance(deviceId: String) = SettingsDeviceDetailFragment()
+            .withArgs {
+                putString(SettingsDeviceConstants.DEVICE_ID_BUNDLE_KEY, deviceId)
```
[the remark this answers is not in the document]
BradleyWire: We're accessing this in multiple classes though?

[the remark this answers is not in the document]
BradleyWire: This is access across multiple classes

app/src/main/kotlin/com/waz/zclient/settings/devices/list/SettingsDeviceListFragment.kt
```diff
@@ -0,0 +1,113 @@
...
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        lifecycleScope.launchWhenResumed {
+            deviceListViewModel.loadData()
+        }
+    }
+
+    private fun initRecyclerView(rootView: View) {
+        devicesRecyclerView = rootView.findViewById(R.id.device_list_recycler_view)
+        val linearLayoutInflater = LinearLayoutManager(requireContext(), LinearLayoutManager.VERTICAL, false)
```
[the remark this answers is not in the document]
BradleyWire: Will fix

app/src/main/kotlin/com/waz/zclient/settings/devices/list/adapter/DevicesRecyclerViewAdapter.kt
```diff
@@ -0,0 +1,64 @@
...
+
+    private val deviceId: TextView by lazy {
+        itemView.findViewById<TextView>(R.id.item_device_id)
+    }
+
+    private val deviceVerifiedIcon: ImageView by lazy {
+        itemView.findViewById<ImageView>(R.id.item_device_verification_icon)
+    }
+
+    fun bind(clientItem: ClientItem, itemClickListener: RecyclerViewItemClickListener<ClientItem>?) {
```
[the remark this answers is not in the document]
BradleyWire: ClientItem is the UI model

app/src/main/kotlin/com/waz/zclient/settings/main/list/SettingsMainListAdapter.kt
```diff
@@ -0,0 +1,34 @@
...
+
+    override fun onBindViewHolder(holder: SettingsViewHolder, position: Int) {
+        holder.bind(settingsItems[position])
+    }
+
+    override fun getItemCount(): Int = settingsItems.size
+
+    inner class SettingsViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
+        fun bind(settingsItem: SettingsMainItem) {
+            with(itemView) {
```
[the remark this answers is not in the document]
mejdoo: I think so

app/src/main/res-v2/layout/item_view_devices.xml
```diff
@@ -0,0 +1,64 @@
...
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-->
+
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:layout_width="match_parent"
+    android:layout_height="@dimen/preference_button_height"
```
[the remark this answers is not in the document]
BradleyWire: We can properly refactor the UI when we're completely finished with the end to end as this is just replicating was it was previously from an attribute aspect rather than identical. I was going to go back and redo all of this anyway

app/src/main/res-v2/layout/settings_switch.xml
```diff
@@ -0,0 +1,51 @@
...
+    <androidx.appcompat.widget.AppCompatTextView
+        android:id="@+id/settings_switch_title"
+        android:layout_width="0dp"
+        android:layout_height="wrap_content"
+        android:layout_gravity="center"
+        android:layout_marginStart="@dimen/preference_button_padding"
+        android:layout_marginEnd="@dimen/preference_button_padding"
+        android:layout_marginBottom="@dimen/preference_button_padding"
+        android:ellipsize="end"
+        android:text="@string/pref_devices_device_verified"
```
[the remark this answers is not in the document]
BradleyWire: This actually is the text it needs

app/src/main/res-v2/layout/settings_switch.xml
```diff
@@ -0,0 +1,51 @@
...
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-->
+
+<merge xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <androidx.appcompat.widget.AppCompatTextView
```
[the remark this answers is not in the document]
BradleyWire: Because it sets the text on the switch button whereas we need the switch button and text to be separate and side by side

app/src/main/kotlin/com/waz/zclient/user/data/mapper/UserEntityMapper.kt
```diff
@@ -0,0 +1,17 @@
+package com.waz.zclient.user.data.mapper
+
+import com.waz.zclient.storage.db.model.UserEntity
+
+import com.waz.zclient.user.domain.model.User
+
+fun UserEntity.toUser() = User(id = id, teamId = teamId, name = name, handle = handle, email = email, phone = phone,
```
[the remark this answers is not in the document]
mejdoo: Check the table **Users** in the database and the network response for /self/profile. This entity combine both of them.

storage/src/main/kotlin/com/waz/zclient/storage/pref/GlobalPreferences.kt
mejdoo: What should we put instead?

app/src/test/kotlin/com/waz/zclient/devices/data/source/remote/ClientsRemoteDataSourceTest.kt
```diff
@@ -0,0 +1,133 @@
...
+    @Test(expected = CancellationException::class)
+    fun `Given getAllClients() is called, when api response is an error, then return an error response`() {
+        runBlocking {
+            `when`(clientsApi.allClients()).thenReturn(allClientsResponse)
+
+            remoteDataSource.allClients()
+
+            verify(clientsApi).allClients()
+
+            cancel(CancellationException(TEST_EXCEPTION_MESSAGE))
```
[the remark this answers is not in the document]
BradleyWire: So we're running this test inside a runBlocking {} closure which enabled you to run suspend functions within it to act as a coroutines job essentially. This cancel method is how you would cancel a job in the production environment and i'm just calling cancel (to suspend all suspend functions within the given data stream)

app/src/test/kotlin/com/waz/zclient/devices/domain/GetAllClientsUseCaseTest.kt
```diff
@@ -0,0 +1,60 @@
...
+    private lateinit var repository: ClientsDataSource
+
+    @Before
+    fun setup() {
+        MockitoAnnotations.initMocks(this)
+        getAllClientsUseCase = GetAllClientsUseCase(repository)
+    }
+
+    @Test
+    fun `given clients response is successful, then repository returns list of a clients`() {
```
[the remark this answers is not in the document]
BradleyWire: Again, this could be something to come up within the style guide. I personal like using Gherkin as it's clear and concise (but can get lengthy)

storage/src/main/kotlin/com/waz/zclient/storage/pref/GlobalPreferences.kt
mejdoo: done

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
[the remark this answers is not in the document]
mejdoo: It's an open discussion...We should define our coding style by the way.

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
```diff
@@ -0,0 +1,90 @@
...
+
+            viewModel.profile.observe(viewLifecycleOwner, Observer { profile ->
+                updateProfile(profile)
+            })
+
+        }
+    }
+
+    private fun setupListeners() {
+        preferences_account_reset_password.setOnClickListener { openUrl(getString(R.string.url_password_forgot).replaceFirst(Accounts, Config.accountsUrl())) }
```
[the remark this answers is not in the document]
mejdoo: You talking here about openUrl(.....) I guess?

app/src/main/kotlin/com/waz/zclient/settings/main/SettingsMainFragment.kt
```diff
@@ -0,0 +1,64 @@
...
+            DEVICES -> replaceFragment(SettingsDeviceListFragment.newInstance())
+            OPTIONS -> replaceFragment(SettingsOptionsFragment.newInstance())
+            ADVANCED -> replaceFragment(SettingsAdvancedFragment.newInstance())
+            SUPPORT -> replaceFragment(SettingsSupportFragment.newInstance())
+            ABOUT -> replaceFragment(SettingsAboutFragment.newInstance())
+        }
+    }
+
+    private fun replaceFragment(fragment: Fragment) {
+        (activity as AppCompatActivity).replaceFragment(R.id.layout_container, fragment, true)
```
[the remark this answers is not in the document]
mejdoo: ok, but we have to make sure **"NO UI COMPONENTS IN THE VIEW MODELS"**

app/src/main/kotlin/com/waz/zclient/core/network/SafeApiDataSource.kt
```diff
@@ -0,0 +1,56 @@
...
+}
+
+suspend fun <R> requestData(request: suspend () -> Either<Failure, R>): Either<Failure, R> =
+    try {
+        request()
+    } catch (e: CancellationException) {
+        Either.Left(Failure(e.localizedMessage))
+    }
+
+suspend fun <R> requestLocal(localRequest: suspend () -> R): Either<Failure, R> =
```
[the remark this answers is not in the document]
BradleyWire: There would be a clash if they had the same name as they're not overloaded functions. They both require the same amount of parameters unless we have them in different files?

app/src/main/kotlin/com/waz/zclient/core/network/SafeApiDataSource.kt
```diff
@@ -0,0 +1,56 @@
...
+    try {
+        Either.Right(localRequest())
+    } catch (e: CancellationException) {
+        Either.Left(Failure(e.localizedMessage))
+    }
+
+suspend fun <R> resultEither(databaseRequest: suspend () -> Either<Failure, R>,
```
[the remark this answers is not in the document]
BradleyWire: Save call request is to save the data from the network request in one go? I don't think we can generally changes the order of the requests without removing the save function?

app/src/main/kotlin/com/waz/zclient/user/data/source/local/UsersLocalDataSource.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.functional.Failure
+import com.waz.zclient.core.network.requestLocal
+
+import com.waz.zclient.storage.db.UserDatabase
+import com.waz.zclient.storage.db.dao.UserDao
+import com.waz.zclient.storage.db.model.UserEntity
+import com.waz.zclient.storage.pref.GlobalPreferences
+
+class UsersLocalDataSource constructor(
```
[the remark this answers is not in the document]
mejdoo: We should inject the global pref and the user database here to avoid problems in unit tests...Do it make sense ?

app/src/main/kotlin/com/waz/zclient/user/domain/usecase/ChangeHandleUseCase.kt
```diff
@@ -0,0 +1,20 @@
...
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.functional.Failure
+import com.waz.zclient.core.usecase.UseCase
+import com.waz.zclient.user.data.UsersRepository
+
+class ChangeHandleUseCase(private val usersRepository: UsersRepository)
+    : UseCase<Any, ChangeHandle>() {
+
+    override suspend fun run(params: ChangeHandle): Either<Failure, Any> = requestData {
+        usersRepository.changePhone(params.handle)
```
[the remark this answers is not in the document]
mejdoo: oh no!

storage/src/main/kotlin/com/waz/zclient/storage/db/model/UserEntity.kt
BradleyWire: After a discussion with @android10 we will be separating the data source response into their own models and then "merging" the responses which will then map to a domain object through the repository implementation.

app/src/test/kotlin/com/waz/zclient/user/data/mapper/UserEntityMapperTest.kt
```diff
@@ -0,0 +1,43 @@
...
+
+import com.waz.zclient.userEntity
+import junit.framework.TestCase.assertEquals
+import org.junit.Test
+
+class UserEntityMapperTest {
+
+    @Test
+    fun `Mapping UserEntity to User should be correct`() {
+        val user = userEntity.toUser()
```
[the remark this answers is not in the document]
mejdoo: TestModels.kt

storage/src/main/kotlin/com/waz/zclient/storage/db/dao/UserDao.kt
mejdoo: fixed

storage/src/main/kotlin/com/waz/zclient/storage/db/model/UserEntity.kt
mejdoo: yes, it should be better. By the way, we don't need anymore as we gonna remove @ignore. We will have 2 separates data classes for api and Dao responses.

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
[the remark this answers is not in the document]
gizemb: @BradleyWire regarding the "multiple times" thing: i've read that kotlin android extensions "builds a local view cache. So the first time a property is used, it will do a regular findViewById. But next times, the view will be recovered from the cache, so the access will be faster." so I think assigning into a variable isn't needed (if the naming is good enough).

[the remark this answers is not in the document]
BradleyWire: okay, that's very handy. I've just now read that too: https://antonioleiva.com/kotlin-android-extensions/. I'll let you know decide when writing up the coding guide in that case.

app/src/main/kotlin/com/waz/zclient/utilities/DateAndTimeUtils.kt
```diff
@@ -0,0 +1,31 @@
...
+            val isLastTwoMins = now.minusMinutes(TWO_MINUTES).isBefore(localTime)
+            val isLastSixtyMinutes = now.minusMinutes(SIXTY_MINUTES).isBefore(localTime)
+
+            return when {
+                isLastTwoMins -> return "Just now"
+                isLastSixtyMinutes -> return "${Duration.between(localTime, now).toMinutes().toInt()} minutes ago"
+                else -> {
+                    time
+                }
+            }
```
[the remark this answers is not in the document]
BradleyWire: I agree with the utils part of this, but, as far as i'm aware this logic is used across multiple features. I'm not sure making this feature specific will help with reusability. i'll come up with something else though.

app/src/main/res-v2/values/pref_keys.xml
```diff
@@ -0,0 +1,36 @@
...
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-->
+<resources>
```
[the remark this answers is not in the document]
BradleyWire: My motto is - if the user won't ever see it, it shouldn't live in an xml file. These could be better placed as constants because they're for the developer?

app/src/test/kotlin/com/waz/zclient/devices/data/ClientsRepositoryTest.kt
```diff
@@ -0,0 +1,140 @@
...
+        assert(domainClient?._class == TEST_CLASS)
+        assert(domainClient?.type == TEST_TYPE)
+        assert(domainClient?.id == TEST_ID)
+        assert(domainClient?.model == TEST_MODEL)
+        assert(domainClient?.location?.long == TEST_LONGITUDE)
+        assert(domainClient?.location?.lat == TEST_LATITUDE)
+    }
+
+    private fun generateMockEntity(): ClientEntity {
+        return ClientEntity(TEST_ID, TEST_TIME, TEST_LABEL, TEST_COOKIE, TEST_TYPE, TEST_CLASS,
```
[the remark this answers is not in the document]
BradleyWire: We'll be changing this model today, so i'll update the testcases too :)

app/src/main/kotlin/com/waz/zclient/core/exception/Failure.kt
```diff
@@ -6,8 +6,9 @@ package com.waz.zclient.core.exception
  */
 sealed class Failure {
     object NetworkConnection : Failure()
-    object ServerError : Failure()
+    data class ServerError(val errorCode: Int, val errorMessage: String) : Failure()
+    object CancellationError : Failure()
```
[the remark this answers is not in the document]
android10: As far as I remember our current approach was not scaling up by using ```sealed classes```. Correct me if I'm wrong @BradleyWire. Something to keep in mind for the next iteration.

app/src/main/kotlin/com/waz/zclient/core/network/SafeApiDataSource.kt
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+suspend fun <R> requestLocal(localRequest: suspend () -> R): Either<Failure, R> =
+    try {
+        Either.Right(localRequest())
+    } catch (e: CancellationException) {
+        Either.Left(Failure.CancellationError)
+    }
+
+suspend fun <R> resultEither(mainRequest: suspend () -> Either<Failure, R>,
```
android10: As we agreed, this could be renamed. Let's keep all these comments here and once we merge, we start fixing them little by little. Otherwise this will grow more and more.

app/src/main/kotlin/com/waz/zclient/devices/data/ClientsDataSource.kt
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+    private fun clientByIdLocal(clientId: String): suspend () -> Either<Failure, Client> =
+        {
+            localDataSource.clientById(clientId).map {
+                clientMapper.toClient(it)
+            }
+        }
+
+    private fun allClientsLocal(): suspend () -> Either<Failure, List<Client>> =
```
android10: All these methods could be in one line. This indentation does not look like concise kotlin code. I guess this could be fixed automatically when @gizemb works on the code style.

android10: They should be like the ```saveClient()``` function :)

app/src/main/kotlin/com/waz/zclient/devices/data/source/ClientMapper.kt
```diff
@@ -0,0 +1,62 @@
+package com.waz.zclient.devices.data.source
+
+import com.waz.zclient.devices.data.source.remote.model.ClientApi
+import com.waz.zclient.devices.domain.model.Client
+import com.waz.zclient.devices.domain.model.ClientLocation
+import com.waz.zclient.storage.db.clients.model.ClientDao
+
+class ClientMapper {
```
android10: Just as a reminder: we can try to make this a bit more generic to reuse the logic somewhere else:
```kotlin
class DataMapper<T> {
    fun <R> convert(data: R, transform: (R) -> T) : T {
        return transform.invoke(data)
    }
}

val clientEntity: ClientEntity = ClientEntity()
val client = DataMapper<Client>().convert(clientEntity, { it.toClient() })
```

app/src/main/kotlin/com/waz/zclient/devices/data/source/local/ClientsLocalDataSource.kt
```diff
@@ -0,0 +1,26 @@
...
+
+    suspend fun clientById(clientId: String): Either<Failure, ClientDao> = requestLocal {
+        clientDbService.clientById(clientId)
+    }
+
+    suspend fun allClients(): Either<Failure, List<ClientDao>> = requestLocal {
+        clientDbService.allClients().toList()
+    }
+
+    fun updateClients(clients: List<ClientDao>) {
```
android10: Keep consistency with the rest of the methods. It is well done with both ```clientById()``` and ```allClients()```.

```kotlin
    fun updateClients(clients: List<ClientDao>) = 
        clientDbService.updateClients(clients) 

    fun updateClient(client: ClientDao) = 
        clientDbService.updateClient(client)
```

app/src/main/kotlin/com/waz/zclient/devices/data/source/remote/ClientsNetworkService.kt
```diff
@@ -0,0 +1,15 @@
+package com.waz.zclient.devices.data.source.remote
+
+import com.waz.zclient.devices.data.source.remote.model.ClientApi
+import retrofit2.Response
+import retrofit2.http.GET
+import retrofit2.http.Path
+
+interface ClientsNetworkService {
```
android10: Within our work for (@gizemb and I) we use these naming the other way around. This is actually a ```ClientApi``` with ```retrofit``` annotations and we have a ```ClientsService``` that uses this Api in order to retrieve data from the network. 

In any case, we need to agree with naming conventions. This is common engineering language that is key to use/understand the current architecture. I will add an item to the android collective document for further discussion.

app/src/main/kotlin/com/waz/zclient/devices/data/source/remote/model/ClientsApi.kt
```diff
@@ -0,0 +1,35 @@
+package com.waz.zclient.devices.data.source.remote.model
+
+import com.google.gson.annotations.SerializedName
+
+data class ClientApi(
```
android10: File is called ```ClientsApi``` whereas class ```ClientApi```. This is also a bit confusing. As in the other comment, we need to discuss this. I would say something like ```ClientInfoApi``` (or something better)

app/src/main/kotlin/com/waz/zclient/devices/domain/GetAllClientsUseCase.kt
```diff
@@ -0,0 +1,16 @@
...
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.network.requestData
+import com.waz.zclient.core.usecase.UseCase
+import com.waz.zclient.devices.data.ClientsRepository
+import com.waz.zclient.devices.domain.model.Client
+
+class GetAllClientsUseCase(private val clientsRepository: ClientsRepository)
+    : UseCase<List<Client>, Unit>() {
+
+    override suspend fun run(params: Unit): Either<Failure, List<Client>> = requestData {
```
android10: Indentation and format:
```kotlin
    override suspend fun run(params: Unit): Either<Failure, List<Client>> = 
        requestData { clientsRepository.allClients() }
```

app/src/main/kotlin/com/waz/zclient/devices/domain/GetSpecificClientUseCase.kt
```diff
@@ -0,0 +1,18 @@
+package com.waz.zclient.devices.domain
+
+import com.waz.zclient.core.exception.Failure
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.network.requestData
+import com.waz.zclient.core.usecase.UseCase
+import com.waz.zclient.devices.data.ClientsRepository
+import com.waz.zclient.devices.domain.model.Client
+
+class GetSpecificClientUseCase(private val clientsRepository: ClientsRepository)
```
android10: The name is a bit misleading and ambiguos to me: 
 - ```GetClientUseCase``` is more meaningful since then when running the use case we force to pass a ```clientId```

app/src/main/kotlin/com/waz/zclient/settings/about/SettingsAboutFragment.kt
```diff
@@ -0,0 +1,36 @@
...
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        return inflater.inflate(R.layout.fragment_about, container, false)
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+        super.onViewCreated(view, savedInstanceState)
+        activity?.title = getString(R.string.pref_about_screen_title)
+        preferences_about_website.setOnClickListener { openUrl(Config.websiteUrl()) }
+        preferences_about_terms.setOnClickListener {}
+        preferences_about_privacy.setOnClickListener { openUrl(getString(R.string.url_privacy_policy).replaceFirst(WEBSITE, Config.websiteUrl())) }
```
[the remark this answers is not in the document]
android10: This is clear a use case. The Fragment is UI and is performing logic: Opening a url. 
Plus I want to be able to test this use case scenario. Fragments should either contain ViewModels or UI stuff.

app/src/main/kotlin/com/waz/zclient/settings/about/SettingsAboutFragment.kt
```diff
@@ -0,0 +1,36 @@
...
+        preferences_about_license.setOnClickListener { openUrl(getString(R.string.url_third_party_licences).replaceFirst(WEBSITE, Config.websiteUrl())) }
+        preferences_about_version.text = getString(R.string.pref_about_version_title, Config.versionName())
+    }
+
+    companion object {
+        fun newInstance() = SettingsAboutFragment()
+        const val WEBSITE = "|WEBSITE|"
```
android10: This constant is not clear to me. There should not be any of this inside a Fragment. They are UI components.

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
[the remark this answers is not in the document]
android10: As I commented before, this is a use case. Also Utils, Managers and Helpers are misleading and they seem to be a black hole where we throw things in when it is not clear what are the responsibilities of a funcion/method/class/interface.

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
```diff
@@ -0,0 +1,90 @@
...
+    private fun updateProfile(userProfileItem: UserProfileItem) {
+        preferences_account_name_title.text = userProfileItem.name
+        preferences_account_email_title.text = userProfileItem.email
+        preferences_account_handle_title.text = userProfileItem.handle
+        preferences_account_phone_title.text = userProfileItem.phone
+    }
+
+    companion object {
+        fun newInstance() = SettingsAccountFragment()
+        private const val Accounts = "|ACCOUNTS|"
```
android10: What is the purpose of this constant?

app/src/main/kotlin/com/waz/zclient/core/exception/Failure.kt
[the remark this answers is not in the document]
BradleyWire: As discussed this will get changed once we know what we're trying to achieve at the UI level

app/src/main/kotlin/com/waz/zclient/settings/about/SettingsAboutFragment.kt
BradleyWire: Open url could be something similar to this: https://github.com/android10/Android-CleanArchitecture-Kotlin/blob/master/app/src/main/kotlin/com/fernandocejas/sample/features/movies/PlayMovie.kt

But, we don't do this until this is complete: https://wearezeta.atlassian.net/browse/AN-6553

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
BradleyWire: Same as: https://github.com/wireapp/wire-android/pull/2477#discussion_r358715429

app/src/main/kotlin/com/waz/zclient/devices/data/source/ClientMapper.kt
BradleyWire: Created a story for this: https://wearezeta.atlassian.net/browse/AN-6544

app/src/main/kotlin/com/waz/zclient/settings/about/SettingsAboutFragment.kt
mejdoo: @android10 this constant is used to force the website url in compile time. We can remove it and do that with string formatting.

app/src/main/kotlin/com/waz/zclient/settings/main/SettingsMainFragment.kt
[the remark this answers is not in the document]
BradleyWire: Opened a subtask for this: https://wearezeta.atlassian.net/browse/AN-6556

app/src/main/res-v2/values/pref_keys.xml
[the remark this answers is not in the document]
BradleyWire: Created a subtask for this: https://wearezeta.atlassian.net/browse/AN-6557

app/src/main/res-v2/values/styles.xml
```diff
@@ -0,0 +1,65 @@
...
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-->
+<resources>
+    <style name="settingsButton" parent="@android:style/TextAppearance.Widget.TextView">
```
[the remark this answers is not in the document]
BradleyWire: Made a subtask for this: https://wearezeta.atlassian.net/browse/AN-6558

app/src/main/kotlin/com/waz/zclient/settings/main/SettingsMainFragment.kt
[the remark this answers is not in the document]
gizemb: yes, you'd be passing a navigation event, not the intent or the view.

app/src/test/kotlin/com/waz/zclient/devices/data/source/remote/ClientsRemoteDataSourceTest.kt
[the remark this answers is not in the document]
gizemb: yep, thanks :)

app/src/main/AndroidManifest.xml
```diff
@@ -85,12 +85,56 @@
...
+                   https://developers.google.com/maps/documentation/android-sdk/config#specify_requirement_for_apache_http_legacy_library
+        -->
         <uses-library android:name="org.apache.http.legacy" android:required="false"/>
 
+        <activity
+            android:name="com.waz.zclient.settings.main.SettingsMainActivity"
+            android:hardwareAccelerated="true"
+            android:theme="@style/Theme.Dark.Preferences"
+            android:launchMode="singleTask"
+            android:label="@string/empty_string"/>
```
gizemb: I think we're gonna check here again, right?

app/src/main/kotlin/com/waz/zclient/devices/data/source/ClientMapper.kt
```diff
@@ -0,0 +1,66 @@
...
+            macKey = macKey,
+            verification = verification
+        )
+    }
+
+    fun toListOfClientDao(list: List<Client>): List<ClientDao> = list.map {
+        toClientDao(it)
+    }
+
+    @JvmName("clientDaoToClients")
```
gizemb: why do we need JvmName?

app/src/test/kotlin/com/waz/zclient/devices/data/ClientsRepositoryTest.kt
```diff
@@ -0,0 +1,144 @@
...
+import com.waz.zclient.storage.db.clients.model.ClientDao
+import kotlinx.coroutines.runBlocking
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mock
+import org.mockito.Mockito.*
+import org.mockito.MockitoAnnotations
+
+class ClientsRepositoryTest {
```
gizemb: this is now ClientsDataSourceTest, isn't it?

app/src/test/kotlin/com/waz/zclient/devices/data/ClientsRepositoryTest.kt
```diff
@@ -0,0 +1,144 @@
...
+    fun `Given getAllClients() is called, when the local data source succeeded, then map the data response to domain`() {
+        runBlocking {
+            `when`(localDataSource.allClients()).thenReturn(Either.Right(listOf(generateMockDao())))
+
+            repository.allClients()
+
+            verify(localDataSource).allClients()
+
+            localDataSource.allClients().map {
+                verify(clientMapper).toListOfClients(it)
```
gizemb: i like that the test now follows the definition without unnecessary mapping logic here :)

app/src/test/kotlin/com/waz/zclient/user/data/UserRepositoryTest.kt
```diff
@@ -0,0 +1,110 @@
...
+import kotlinx.coroutines.runBlocking
+import org.junit.Assert.assertEquals
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mock
+import org.mockito.Mockito.`when`
+import org.mockito.Mockito.verify
+import org.mockito.MockitoAnnotations
+
+class UserRepositoryTest {
```
gizemb: UsersDataSourceTest

app/src/test/scala/com/waz/zclient/SecurityChecklistTest.scala
```diff
@@ -1,161 +0,0 @@
-/**
```
gizemb: why is this deleted??

storage/src/main/kotlin/com/waz/zclient/storage/db/UserDatabase.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.room.RoomDatabase
+import com.waz.zclient.storage.db.clients.model.ClientDao
+import com.waz.zclient.storage.db.clients.service.ClientDbService
+import com.waz.zclient.storage.db.users.migration.UserDatabaseMigration
+import com.waz.zclient.storage.db.users.model.UserDao
+import com.waz.zclient.storage.db.users.model.UserPreferenceDao
+import com.waz.zclient.storage.db.users.service.UserDbService
+import com.waz.zclient.storage.db.users.service.UserPreferenceDbService
+
+@Database(entities = [UserPreferenceDao::class, UserDao::class, ClientDao::class], version = 125, exportSchema = false)
```
gizemb: shouldn't it be 126?

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/service/ClientDbService.kt
```diff
@@ -0,0 +1,29 @@
...
+    @Transaction
+    suspend fun updateClients(clients: List<ClientDao>) {
+        deleteAllClients()
+        insertAll(clients)
+    }
+
+    @Insert
+    suspend fun insertAll(clients: List<ClientDao>)
+
+    @Query("DELETE FROM client")
```
gizemb: we can use @ Delete maybe

storage/src/main/kotlin/com/waz/zclient/storage/extension/SharedPreference.kt
```diff
@@ -0,0 +1,11 @@
...
+
+import android.content.SharedPreferences
+
+fun SharedPreferences.string(key: String, value: String) {
+    edit().putString(key, value).apply()
+}
+
+fun SharedPreferences.string(key: String): String {
+    return getString(key, "")
+}
```
gizemb: this is a very weird naming imo. what if want to have a sharedPref getter with the default value?? then it'll be `SharedPreferences.string(String, String)` too. which is the same signature with put method. we should change these to putString and getString

BradleyWire:
build again

storage/src/main/kotlin/com/waz/zclient/storage/db/UserDatabase.kt
mejdoo: bumped for now

storage/src/main/kotlin/com/waz/zclient/storage/extension/SharedPreference.kt
mejdoo: @android10 what do you suggest? How to apply Kotlin idiomatic style for this case?

mejdoo:
build again

PR #616, opened 2017-02-17
Upgrade AVS with SE for API changes

deanrobertcook:
#### APK
[Download build #8482](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8482/artifact/build/artifact/wire-dev-PR616-8482.apk)

PR #2479, opened 2019-12-12
refactor: harden account wiping code

CalumMcCall:
### Issues

The `wipeData` method in the `AccountsService` would abort account wiping and logout, allowing a user to bypass the account lock and gain full access to Wire, if an exception was thrown during the execution of `wipeData`. [A ticket for the issue is here](https://wearezeta.atlassian.net/browse/AN-6541).

### Causes

Thrown exceptions, for example from file operations, would cause the whole method to abort, skipping the logout check and further wiping steps.

### Solutions

The method was rewritten to wrap all risky operations in `Try` constructs. This way, we can skip failing steps and always continue to at least logging the user out, if we are unable to wipe all the data. Unfortunately, this means the method is less easy to read now, but I think in this case it makes sense to harden this particular method given the security risk. I'm open to suggestions for making it all nicer. 

### Testing

Testing was done manually.
#### APK
[Download build #616](http://10.10.124.11:8080/job/Pull%20Request%20Builder/616/artifact/build/artifact/wire-dev-PR2479-616.apk)

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/AccountsService.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/AccountsService.scala
@@ -29,7 +29,6 @@ import com.waz.log.LogShow.SafeToLog
 import com.waz.model.AccountData.Password
 import com.waz.model._
 import com.waz.service.backup.BackupManager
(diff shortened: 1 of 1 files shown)


AndroidBob:

marcoconti83:
ok to build

makingthematrix:
build again

PR #2481, opened 2019-12-13
Refactor: Combine dependencies & versions into a single file

gizemb:
### Issues

It is hard to track dependencies in separate modules. We combined all into a single file to be able to search and update them from a single place.

### Notes

Shouldn't change any behavior since there's no version update.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -1,0 +1,43 @@
+/**
+ * Wire
+ * Copyright (C) 2019 Wire Swiss GmbH
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


dependencies.gradle
```diff
@@ -0,0 +1,45 @@
...
+        kluent      : "1.14",
+        lifecycle   : "2.0.0",
+        okHttp      : "3.12.0",
+        circe       : "0.9.3",
+        retrofit    : "2.6.2",
+        macwire     : "2.3.3",
+        glide       : "4.10.0"
+    ]
+
+    deps = [
```
BradleyWire: Would we not want more key dependencies in this collection/ Maybe at-least the ones shared across multiple modules?

android10: This file In general is growing in size, making it less and less readable and hard to understand. We will need to extract many of this stuff to separated files. AFAIK dependencies could sit in a ```dependencies.gradle``` that can be merged with the main ```gradle``` file

android10:
@gizemb what is the current state of this one?

android10:
@gizemb closing this one in favor of #2514. Great job in any case :+1: :tada: :)

Issue #665, opened 2017-02-27
Network switch during the call is not handeled


RebelCoderRU:
Hello.

Wire does not handle network switch on the fly.

Scenario:
1) Use Cell Network to call from Android to Android.
2) During the call, switch on/walk in to the WiFi area, to make your device connect to it
3) After network switch has accrued, it will  cause the call to stay active, still counting the call time, but no data is being sent between callers.

Scenario can be used for both cases Use Cellular network and switch to WiFi, or use WiFi and switch to 4G during the active call.

As an example, Skype handles this good and re-connects with a new IP and continues the Video/Audio call.

Tested on:
Devices: Samsung Galaxy A9 Pro, Galaxy A7.
Android 6.0.1

Thanks.

someoneEsle:
Was going to open a similar issue myself! In my case, after switching from wifi to data during a call, the call just dropped. I think it may be because my phone took too long to reconnect after it lost the wifi connection (?). Happened on a Nexus 5X, Android 7.1

bjoernherbig:
Thank you for reporting. You can expect changes regarding this soon

makingthematrix:
This was fixed at some point in the past. The switching still cause a small glitch but all works now.

PR #2491, opened 2019-12-17
fix: flaky test in SignalSpec

CalumMcCall:
### Issues

SignalSpec was failing about 1/3 of the time. Ticket: [AN-6566](https://wearezeta.atlassian.net/browse/AN-6566)

### Causes

Since this is a recurring issue, I'll give some background for future maintainers. When I disabled all tests except the flaky one, I managed to reproduce the exception(an example of this is in the ticket) in the for comprehension test reliably. Then from printing out state inside the `eventually` block on line 204, it became clear that `received` wasn't actually being updated with the value 2 until several attempts had been made. So it seems that the value 2 wasn't being saved in `received` until a considerable delay, by which point the final `eventually` block on line 211 had failed.

### Solutions

Checking for the state required before proceeding using an `eventually` block seems to do the trick.
#### APK
[Download build #665](http://10.10.124.11:8080/job/Pull%20Request%20Builder/665/artifact/build/artifact/wire-dev-PR2491-665.apk)


AndroidBob:

marcoconti83:
ok to test

marcoconti83:
build again

PR #676, opened 2017-03-02
[AN-5067] Updates adapter on any type of changes

JPonte:
#### APK

[Download build #8564](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8564/artifact/build/artifact/wire-dev-PR676-8564.apk)
[Download build #8569](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8569/artifact/build/artifact/wire-dev-PR676-8569.apk)

PR #2496, opened 2019-12-19
Fix for conversation roles in UserSearch and ConversationList

makingthematrix:
The bug was caused by using conversation roles when `currentConvId` was not yet set - after the first start of the app. On top of conversation roles we have the team roles. Sometimes the two overlap and that was the source of confusion.For example, in the User Search, we should use team roles to check if the user is able to add services to a conversation,not the conversation roles.

In the process of fixing I made some small changes to `IntegrationDetailsFragment`, but I decided not to rewrite it more than needed.It should be either refactored in another PR, or rewritten in Kotlin completely.

Test:
1. Log in as a team user to a freshly installed app.
2. Go to UserSearch

Expected result:
You see the "Create group" button, as well as some contacts if you have them.

Actual result:
No "Create group" button and the list is empty.
#### APK
[Download build #674](http://10.10.124.11:8080/job/Pull%20Request%20Builder/674/artifact/build/artifact/wire-dev-PR2496-674.apk)
[Download build #676](http://10.10.124.11:8080/job/Pull%20Request%20Builder/676/artifact/build/artifact/wire-dev-PR2496-676.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/common/controllers/UserAccountsController.scala
+++ b/app/src/main/scala/com/waz/zclient/common/controllers/UserAccountsController.scala
@@ -81,6 +81,8 @@ class UserAccountsController(implicit injector: Injector, context: Context, ec: 
         decodeBitmask(bitmask)
       }
 
+  lazy val isAdmin: Signal[Boolean] = selfPermissions.map(AdminPermissions.subsetOf)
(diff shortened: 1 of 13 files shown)


AnastasiiaWire:
build again

PR #810, opened 2017-04-22
Code cleanup - mostly regarding deprecated api's.

pprucia:
Code cleanup - mostly regarding deprecated api's. Devices with newer api available will use it, older ones will use legacy api.
#### APK
[Download build #8800](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8800/artifact/build/artifact/wire-dev-PR810-8800.apk)

Issue #802, opened 2017-04-18
New messages: status notification canceled after reading one of them


theincogtion:
If I get a new message in wire it is shown in the status bar on android.
This also happens if 2 or more messages are unread.
If I read one of them the notification in the statusbar about the unread messages disappears.
I often forget to read the other one because Wire does not show unread messages for me.

makingthematrix:
Hi,
I'm going to close this issue. We had a lot of changes over time to the notifications feature. Please open a new ticket if this is still an issue.

Issue #801, opened 2017-04-17
I can invite people who are already on Wire


someoneEsle:
## The problem

When I go to my contacts, I can see a list that includes:
1. people who are on Wire, with their profile pic and name/nickname
2. people who are not on Wire, with a big "Invite" button
3. people who *are* on Wire, who still display the big "Invite" button as if they didn't have the app

People from group 3 are indistinguishable from people from group 2. They are active users, and they have all signed up with their phones (some of them may have added an email to their accounts, but not all of them). Note that not all of people from group 1 have their counterpart in group 3 -- the two are disjoint. I cannot figure out what people in group 3 have in common :/
It may be worth saying that all of the phone numbers on my phone are saved with their respective country code (+1, +44, etc).

## Environment

* Wire version that exhibits the issue: 2.31.324
* Android OS version used to run Wire: stock Android 7.1.2
* Mobile phone model/manufacturer: Nexus 5X

forteller:
Yeah, I've gotten an invite sms from someone once even though I am registered with my phone number

Issue #750, opened 2017-03-26
Allow use of chat while been in call


avently:
For now you can't use chat when you are talking with somebody. I think we should have a way to use the chat for all features except audio message. 
Interface needs to have a button which hides call activity and then shows the activity back when it's needed.

theincogtion:
By the way: In Windows it is possible to chat but not in Android :(

mandreyel:
+1

This can be rather annoying (e.g. when sharing a picture being discussed in a call) and seems like an arbitrary restriction. Is there any specific reason this feature is not implemented?

tariq-chameleon:
Very often I find myself willing to clarify myself to the person I am in voice-chat with using a text/picture/link/video/local file but I am not being able to do that unless I go to my desktop machine and use the desktop version.

someoneEsle:
Hey everyone! We just pushed a beta to the play store that adds this functionality. Let us know if you find any bugs!

PR #2510, opened 2019-12-23
Kotlin network framework part2

gizemb:
## General Architecture
network_architecture

WIP
#### APK
[Download build #716](http://10.10.124.11:8080/job/Pull%20Request%20Builder/716/artifact/build/artifact/wire-dev-PR2510-716.apk)
[Download build #717](http://10.10.124.11:8080/job/Pull%20Request%20Builder/717/artifact/build/artifact/wire-dev-PR2510-717.apk)
[Download build #720](http://10.10.124.11:8080/job/Pull%20Request%20Builder/720/artifact/build/artifact/wire-dev-PR2510-720.apk)
[Download build #725](http://10.10.124.11:8080/job/Pull%20Request%20Builder/725/artifact/build/artifact/wire-dev-PR2510-725.apk)
[Download build #726](http://10.10.124.11:8080/job/Pull%20Request%20Builder/726/artifact/build/artifact/wire-dev-PR2510-726.apk)
[Download build #732](http://10.10.124.11:8080/job/Pull%20Request%20Builder/732/artifact/build/artifact/wire-dev-PR2510-732.apk)
[Download build #735](http://10.10.124.11:8080/job/Pull%20Request%20Builder/735/artifact/build/artifact/wire-dev-PR2510-735.apk)
[Download build #750](http://10.10.124.11:8080/job/Pull%20Request%20Builder/750/artifact/build/artifact/wire-dev-PR2510-750.apk)
[Download build #801](http://10.10.124.11:8080/job/Pull%20Request%20Builder/801/artifact/build/artifact/wire-dev-PR2510-801.apk)
[Download build #802](http://10.10.124.11:8080/job/Pull%20Request%20Builder/802/artifact/build/artifact/wire-dev-PR2510-802.apk)
[Download build #803](http://10.10.124.11:8080/job/Pull%20Request%20Builder/803/artifact/build/artifact/wire-dev-PR2510-803.apk)
[Download build #810](http://10.10.124.11:8080/job/Pull%20Request%20Builder/810/artifact/build/artifact/wire-dev-PR2510-810.apk)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -83,6 +83,7 @@
         android:theme="@style/Theme.Splash"
         android:vmSafeMode="${applicationVmSafeMode}"
         tools:replace="android:allowBackup"
+        android:networkSecurityConfig="@xml/network_security_config"
         >
 
         <!--    TODO: not required when using play-services-maps:16.1.0 and above
--- a/app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
+++ b/app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
@@ -5,7 +5,17 @@ import com.waz.zclient.BuildConfig
 import com.waz.zclient.core.di.NetworkDependencyProvider.createHttpClient
 import com.waz.zclient.core.di.NetworkDependencyProvider.createHttpClientForToken
 import com.waz.zclient.core.di.NetworkDependencyProvider.retrofit
-import com.waz.zclient.core.network.*
+import com.waz.zclient.core.network.AccessTokenAuthenticator
+import com.waz.zclient.core.network.AccessTokenInterceptor
+import com.waz.zclient.core.network.AccessTokenLocalDataSource
+import com.waz.zclient.core.network.AccessTokenRemoteDataSource
+import com.waz.zclient.core.network.AccessTokenRepository
+import com.waz.zclient.core.network.ApiService
+import com.waz.zclient.core.network.AuthTokenHandler
+import com.waz.zclient.core.network.NetworkClient
+import com.waz.zclient.core.network.NetworkHandler
+import com.waz.zclient.core.network.RawResponseRegistry
+import com.waz.zclient.core.network.RetrofitClient
 import com.waz.zclient.core.network.api.token.TokenApi
 import com.waz.zclient.core.network.api.token.TokenService
 import com.waz.zclient.core.threading.ThreadHandler
@@ -54,7 +64,7 @@ val networkModule: Module = module {
     single { createHttpClient(get(), get()) }
     single { retrofit(get()) }
     single { AccessTokenRemoteDataSource(get()) }
-    factory {
+    single {
         AccessTokenLocalDataSource(
             //TODO: get from other(global?) module
             androidContext().getSharedPreferences("DUMMY_USER_PREFS", Context.MODE_PRIVATE)
@@ -75,6 +85,6 @@ val networkModule: Module = module {
     val networkClientForToken = "NETWORK_CLIENT_FOR_TOKEN"
     single<NetworkClient>(named(networkClientForToken)) { RetrofitClient(retrofit(createHttpClientForToken())) }
     single(named(apiServiceForToken)) { ApiService(get(), get(), get(named(networkClientForToken)), get()) }
-    factory { get<ApiService>(named(apiServiceForToken)).createApi(TokenApi::class.java) }
+    single { get<ApiService>(named(apiServiceForToken)).createApi(TokenApi::class.java) }
     single { TokenService(get(named(apiServiceForToken)), get()) }
 }
--- a/app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
+++ b/app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
@@ -53,7 +53,7 @@ object NetworkDependencyProvider {
         }.build()
 
     private fun OkHttpClient.Builder.addLoggingInterceptor() {
-        val loggingInterceptor = HttpLoggingInterceptor().setLevel(Level.BASIC)
+        val loggingInterceptor = HttpLoggingInterceptor().setLevel(Level.BODY)
         addInterceptor(loggingInterceptor)
     }
 }
@@ -71,6 +71,7 @@ val networkModule: Module = module {
     single { AccessTokenRepository(get(), get()) }
     single { AuthTokenHandler(get()) }
     single { AccessTokenAuthenticator(get()) }
+    single { AccessTokenInterceptor(get()) }
     single<NetworkClient> { RetrofitClient(get()) }
     single { ApiService(get(), get(), get()) }
 
--- a/app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
+++ b/app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
@@ -2,8 +2,6 @@ package com.waz.zclient.core.di
 
 import android.content.Context
 import com.waz.zclient.BuildConfig
-import com.waz.zclient.core.di.NetworkDependencyProvider.API_SERVICE_FOR_TOKEN
-import com.waz.zclient.core.di.NetworkDependencyProvider.NETWORK_CLIENT_FOR_TOKEN
 import com.waz.zclient.core.di.NetworkDependencyProvider.createHttpClient
 import com.waz.zclient.core.di.NetworkDependencyProvider.createHttpClientForToken
 import com.waz.zclient.core.di.NetworkDependencyProvider.retrofit
@@ -25,9 +23,6 @@ object NetworkDependencyProvider {
 
     private const val BASE_URL = "https://staging-nginz-https.zinfra.io"
 
-    const val API_SERVICE_FOR_TOKEN = "API_SERVICE_FOR_TOKEN"
-    const val NETWORK_CLIENT_FOR_TOKEN = "NETWORK_CLIENT_FOR_TOKEN"
-
     fun retrofit(okHttpClient: OkHttpClient): Retrofit =
         Retrofit.Builder()
             .baseUrl(BASE_URL)
@@ -37,24 +32,19 @@ object NetworkDependencyProvider {
 
     fun createHttpClient(accessTokenInterceptor: AccessTokenInterceptor,
                          accessTokenAuthenticator: AccessTokenAuthenticator): OkHttpClient =
-        OkHttpClient.Builder().apply {
-            addInterceptor(accessTokenInterceptor)
-            authenticator(accessTokenAuthenticator)
-            if (BuildConfig.DEBUG) {
-                addLoggingInterceptor()
-            }
-        }.build()
+        OkHttpClient.Builder()
+            .addInterceptor(accessTokenInterceptor)
+            .authenticator(accessTokenAuthenticator)
+            .addLoggingInterceptor()
+            .build()
 
     fun createHttpClientForToken(): OkHttpClient =
-        OkHttpClient.Builder().apply {
-            if (BuildConfig.DEBUG) {
-                addLoggingInterceptor()
-            }
-        }.build()
+        OkHttpClient.Builder().addLoggingInterceptor().build()
 
-    private fun OkHttpClient.Builder.addLoggingInterceptor() {
-        val loggingInterceptor = HttpLoggingInterceptor().setLevel(Level.BODY)
-        addInterceptor(loggingInterceptor)
+    private fun OkHttpClient.Builder.addLoggingInterceptor() = this.apply {
+        if (BuildConfig.DEBUG) {
+            addInterceptor(HttpLoggingInterceptor().setLevel(Level.BODY))
+        }
     }
 }
 
@@ -64,10 +54,12 @@ val networkModule: Module = module {
     single { createHttpClient(get(), get()) }
     single { retrofit(get()) }
     single { AccessTokenRemoteDataSource(get()) }
-    factory { AccessTokenLocalDataSource(
-        //TODO: get from other(global?) module
-        androidContext().getSharedPreferences("DUMMY_USER_PREFS", Context.MODE_PRIVATE)
-    ) }
+    factory {
+        AccessTokenLocalDataSource(
+            //TODO: get from other(global?) module
+            androidContext().getSharedPreferences("DUMMY_USER_PREFS", Context.MODE_PRIVATE)
+        )
+    }
     single { AccessTokenRepository(get(), get()) }
     single { AuthTokenHandler(get()) }
     single { AccessTokenAuthenticator(get()) }
@@ -76,9 +68,11 @@ val networkModule: Module = module {
     single { ApiService(get(), get(), get()) }
 
     //Token manipulation
-    single<NetworkClient>(named(NETWORK_CLIENT_FOR_TOKEN)) { RetrofitClient(retrofit(createHttpClientForToken())) }
-    single(named(API_SERVICE_FOR_TOKEN)) { ApiService(get(), get(), get(named(NETWORK_CLIENT_FOR_TOKEN))) }
-    factory { get<ApiService>(named(API_SERVICE_FOR_TOKEN)).createApi(TokenApi::class.java)}
-    single { TokenService(get(named(API_SERVICE_FOR_TOKEN)), get()) }
+    val apiServiceForToken = "API_SERVICE_FOR_TOKEN"
+    val networkClientForToken = "NETWORK_CLIENT_FOR_TOKEN"
+    single<NetworkClient>(named(networkClientForToken)) { RetrofitClient(retrofit(createHttpClientForToken())) }
+    single(named(apiServiceForToken)) { ApiService(get(), get(), get(named(networkClientForToken))) }
+    factory { get<ApiService>(named(apiServiceForToken)).createApi(TokenApi::class.java) }
+    single { TokenService(get(named(apiServiceForToken)), get()) }
 }
 
--- a/app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
+++ b/app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
@@ -1,20 +1,19 @@
 package com.waz.zclient.core.di
 
 import com.waz.zclient.BuildConfig
-import com.waz.zclient.core.network.AccessTokenAuthenticator
-import com.waz.zclient.core.network.AccessTokenInterceptor
-import com.waz.zclient.core.network.AccessTokenRepository
-import com.waz.zclient.core.network.ApiService
-import com.waz.zclient.core.network.AuthTokenHandler
-import com.waz.zclient.core.network.NetworkClient
-import com.waz.zclient.core.network.NetworkHandler
-import com.waz.zclient.core.network.RetrofitClient
+import com.waz.zclient.core.di.NetworkDependencyProvider.API_SERVICE_FOR_TOKEN
+import com.waz.zclient.core.di.NetworkDependencyProvider.NETWORK_CLIENT_FOR_TOKEN
+import com.waz.zclient.core.di.NetworkDependencyProvider.createHttpClient
+import com.waz.zclient.core.di.NetworkDependencyProvider.createHttpClientForToken
+import com.waz.zclient.core.di.NetworkDependencyProvider.retrofit
+import com.waz.zclient.core.network.*
 import com.waz.zclient.core.threading.ThreadHandler
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
 import okhttp3.logging.HttpLoggingInterceptor.Level
 import org.koin.android.ext.koin.androidContext
 import org.koin.core.module.Module
+import org.koin.core.qualifier.named
 import org.koin.dsl.module
 import retrofit2.Retrofit
 import retrofit2.converter.gson.GsonConverterFactory
@@ -23,6 +22,9 @@ object NetworkDependencyProvider {
 
     private const val BASE_URL = "https://staging-nginz-https.zinfra.io"
 
+    const val API_SERVICE_FOR_TOKEN = "API_SERVICE_FOR_TOKEN"
+    const val NETWORK_CLIENT_FOR_TOKEN = "NETWORK_CLIENT_FOR_TOKEN"
+
     fun retrofit(okHttpClient: OkHttpClient): Retrofit =
         Retrofit.Builder()
             .baseUrl(BASE_URL)
@@ -30,26 +32,44 @@ object NetworkDependencyProvider {
             .addConverterFactory(GsonConverterFactory.create())
             .build()
 
-    fun createClient(accessTokenInterceptor: AccessTokenInterceptor,
-                     accessTokenAuthenticator: AccessTokenAuthenticator): OkHttpClient =
+    fun createHttpClient(accessTokenInterceptor: AccessTokenInterceptor,
+                         accessTokenAuthenticator: AccessTokenAuthenticator): OkHttpClient =
         OkHttpClient.Builder().apply {
             addInterceptor(accessTokenInterceptor)
             authenticator(accessTokenAuthenticator)
             if (BuildConfig.DEBUG) {
-                val loggingInterceptor = HttpLoggingInterceptor().setLevel(Level.BASIC)
-                addInterceptor(loggingInterceptor)
+                addLoggingInterceptor()
             }
         }.build()
+
+    fun createHttpClientForToken(): OkHttpClient =
+        OkHttpClient.Builder().apply {
+            if (BuildConfig.DEBUG) {
+                addLoggingInterceptor()
+            }
+        }.build()
+
+    private fun OkHttpClient.Builder.addLoggingInterceptor() {
+        val loggingInterceptor = HttpLoggingInterceptor().setLevel(Level.BASIC)
+        addInterceptor(loggingInterceptor)
+    }
 }
 
 val networkModule: Module = module {
+    single { ThreadHandler() }
     single { NetworkHandler(androidContext()) }
-    single { NetworkDependencyProvider.createClient(get(), get()) }
-    single { NetworkDependencyProvider.retrofit(get()) }
-    single { AccessTokenRepository() }
+    single { createHttpClient(get(), get()) }
+    single { retrofit(get()) }
+    single { AccessTokenRemoteDataSource(get()) }
+    single { AccessTokenRepository(get()) }
     single { AuthTokenHandler(get()) }
     single { AccessTokenAuthenticator(get()) }
     single<NetworkClient> { RetrofitClient(get()) }
-    single { ThreadHandler() }
     single { ApiService(get(), get(), get()) }
+
+    //Token manipulation
+    single<NetworkClient>(named(NETWORK_CLIENT_FOR_TOKEN)) { RetrofitClient(retrofit(createHttpClientForToken())) }
+    single(named(API_SERVICE_FOR_TOKEN)) { ApiService(get(), get(), get(named(NETWORK_CLIENT_FOR_TOKEN))) }
+    single { TokenService(get(named(API_SERVICE_FOR_TOKEN))) }
 }
+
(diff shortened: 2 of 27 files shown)


storage/src/main/kotlin/com/waz/zclient/storage/extension/SharedPreference.kt
```diff
@@ -2,10 +2,8 @@ package com.waz.zclient.storage.extension
...
 
-fun SharedPreferences.string(key: String, value: String) {
-    edit().putString(key, value).apply()
-}
+fun SharedPreferences.putString(key: String, value: String) = edit().putString(key, value).apply()
 
-fun SharedPreferences.string(key: String): String {
-    return getString(key, "")
-}
+fun SharedPreferences.string(key: String, defValue: String = ""): String? = getString(key, defValue)
```
gizemb: this is something temporary but there were 2 problems with the previous implementation:
1- it wasn't returning a non-null value, since getString() may return null if sb intentionally put a null value with a given key before (so the signature was wrong and prune to runtime errors).
2- there were no means to provide a default value, which imo is restricting the api

however, now, this method is essentially only changing naming from getString to string and doing nothing else meaningful, which is again weird :)

we can discuss whether such tailoring makes sense in framework methods.

app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
```diff
@@ -23,33 +25,60 @@ object NetworkDependencyProvider {
...
             if (BuildConfig.DEBUG) {
-                val loggingInterceptor = HttpLoggingInterceptor().setLevel(Level.BASIC)
-                addInterceptor(loggingInterceptor)
+                addLoggingInterceptor()
+            }
+        }.build()
+
+    fun createHttpClientForToken(): OkHttpClient =
+        OkHttpClient.Builder().apply {
+            if (BuildConfig.DEBUG) {
```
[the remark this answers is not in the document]
android10:
```kotlin
    fun createHttpClient(accessTokenInterceptor: AccessTokenInterceptor,
                         accessTokenAuthenticator: AccessTokenAuthenticator): OkHttpClient =
        OkHttpClient.Builder().apply {
            addInterceptor(accessTokenInterceptor)
            authenticator(accessTokenAuthenticator)
            addLoggingInterceptor()
        }.build()

    fun createHttpClientForToken(): OkHttpClient =
        OkHttpClient.Builder().apply {
            addLoggingInterceptor()
        }.build()
```

app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
```diff
@@ -23,33 +25,60 @@ object NetworkDependencyProvider {
...
+    single { ThreadHandler() }
     single { NetworkHandler(androidContext()) }
-    single { NetworkDependencyProvider.createClient(get(), get()) }
-    single { NetworkDependencyProvider.retrofit(get()) }
-    single { AccessTokenRepository() }
+    single { createHttpClient(get(), get()) }
+    single { retrofit(get()) }
+    single { AccessTokenRemoteDataSource(get()) }
+    factory { AccessTokenLocalDataSource(
+        //TODO: get from other(global?) module
```
android10: :+1: When we re consider how we organize global dependencies, we can definitely get that from another module.

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenLocalDataSource.kt
```diff
@@ -0,0 +1,29 @@
...
+    fun refreshToken(): String? = userPreferences.string(KEY_REFRESH_TOKEN)
+
+    fun updateRefreshToken(newRefreshToken: String) =
+        userPreferences.putString(KEY_REFRESH_TOKEN, newRefreshToken)
+
+    fun wipeOutAccessToken() = userPreferences.removeString(KEY_ACCESS_TOKEN)
+
+    fun wipeOutRefreshToken() = userPreferences.removeString(KEY_REFRESH_TOKEN)
+
+    companion object {
```
android10: This is only make up but as companion object conventions this should be at the top of the class.

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenRemoteDataSource.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.waz.zclient.core.exception.Failure
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.network.api.token.AccessTokenResponse
+import com.waz.zclient.core.network.api.token.TokenService
+
+class AccessTokenRemoteDataSource(private val tokenService: TokenService) {
+
+    fun renewAccessToken(refreshToken: String) : Either<Failure, AccessTokenResponse> =
+        tokenService.renewAccessToken(refreshToken)
+
```
android10: Remove new line

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenRepository.kt
```diff
@@ -1,42 +1,32 @@
...
+    fun refreshToken(): String = localDataSource.refreshToken() ?: String.empty()
 
-    fun updateRefreshToken(newRefreshToken: String) {
-        //TODO: Save the refresh token somewhere: User Preferences?
-    }
+    fun updateRefreshToken(newRefreshToken: String) =
+        localDataSource.updateRefreshToken(newRefreshToken)
 
-    fun renewAccessToken(refreshToken: String) : Either<Failure, String> = TODO("implement")
+    //TODO: do we need an intermediary Access Token repository model here?
```
android10: What do you mean here?

app/src/main/kotlin/com/waz/zclient/core/network/AuthTokenHandler.kt
```diff
@@ -17,7 +18,9 @@ class AuthTokenHandler(private val tokenRepository: AccessTokenRepository) {
     fun updateRefreshToken(newRefreshToken: String) = tokenRepository.updateRefreshToken(newRefreshToken)
 
     fun renewAccessToken(refreshToken: String): Either<Failure, String> =
-        tokenRepository.renewAccessToken(refreshToken)
+        tokenRepository.renewAccessToken(refreshToken).map {
```
android10: Only format and personal taste here: 
```kotlin
fun renewAccessToken(refreshToken: String): Either<Failure, String> =
        tokenRepository.renewAccessToken(refreshToken).map { it.token }
```

app/src/main/AndroidManifest.xml
```diff
@@ -83,6 +83,7 @@
         android:theme="@style/Theme.Splash"
         android:vmSafeMode="${applicationVmSafeMode}"
         tools:replace="android:allowBackup"
+        android:networkSecurityConfig="@xml/network_security_config"
```
gizemb: note that this will affect the whole app once merged to develop. It's only enabled for debug though, so shouldn't affect the production code. I opened [an issue here](https://wearezeta.atlassian.net/browse/AN-6589) for testing

app/src/main/kotlin/com/waz/zclient/core/network/api/token/TokenService.kt
```diff
@@ -0,0 +1,18 @@
...
+import com.waz.zclient.core.exception.Failure
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.network.ApiService
+
+class TokenService(private val apiService: ApiService,
+                   private val tokenApi: TokenApi) {
+
+    fun renewAccessToken(refreshToken: String): Either<Failure, AccessTokenResponse> =
+        apiService.request(
+            tokenApi.access(mapOf(
```
android10: Some indentation? Definitely all this should be automatic when you setup the code style :)    
```kotlin
apiService.request( 
      tokenApi.access(mapOf( "Cookie" to "zuid=$refreshToken")), AccessTokenResponse.EMPTY
)
```

app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,28 @@
...
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-->
+<network-security-config>
```
android10: Just out of curiosity. What is this?

app/src/test/kotlin/com/waz/zclient/core/network/AccessTokenRepositoryTest.kt
```diff
@@ -0,0 +1,94 @@
...
+
+import com.waz.zclient.UnitTest
+import com.waz.zclient.core.extension.empty
+import org.amshove.kluent.shouldBe
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mock
+import org.mockito.Mockito.*
+
+class AccessTokenRepositoryTest : UnitTest() {
```
android10: Well done :heart:

app/src/test/kotlin/com/waz/zclient/core/network/api/token/TokenServiceTest.kt
```diff
@@ -0,0 +1,42 @@
...
+
+        tokenService.renewAccessToken(refreshToken)
+
+        verify(apiService).request(any(), any<AccessTokenResponse>())
+
+        val argumentCaptor = ArgumentCaptor.forClass(Map::class.java) as ArgumentCaptor<Map<String, String>>
+        verify(tokenApi).access(capture(argumentCaptor))
+        argumentCaptor.value `should contain` ("Cookie" to "zuid=$refreshToken")
+    }
+
```
android10: Remove space

storage/src/main/kotlin/com/waz/zclient/storage/extension/SharedPreference.kt
android10: Totally agree! We should discuss how and when we want to extend functionality. My point is that this creates static methods which most of the time are not being tested (or they are hard to test) and they are prone to introduce bugs.

app/src/main/res/xml/network_security_config.xml
gizemb: I needed to sniff and manipulate network requests/responses to test authentication logic --otherwise I have to wait for 15 mins for the token to expire--. So, I used Charles, which does this job by installing its own certificate to device (which makes it a user-installed certificate). But for devices running on or above Android 7, we need to add this file to the app in order to make the app allow user installed certificates. I think debugging the network requests will be useful in general, so I added this file to project instead of keeping it locally.
more info: https://developer.android.com/training/articles/security-config
jira issue for testing: https://wearezeta.atlassian.net/browse/AN-6589

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenRepository.kt
gizemb: currently this method returns AccessTokenResponse, which is the model class of the network response. I asked whether we should have a repository model class that we should map to such as: NetworkResponse -> RepositoryModel. we will probably do, once local data source has its own model class (like LocalEntity -> RepositoryModel), but for now we don't need it, and I didn't create one.

app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
```diff
@@ -23,33 +25,60 @@ object NetworkDependencyProvider {
 
     private const val BASE_URL = "https://staging-nginz-https.zinfra.io"
 
+    const val API_SERVICE_FOR_TOKEN = "API_SERVICE_FOR_TOKEN"
```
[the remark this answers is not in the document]
gizemb: i agree, but can't create a const val in that scope but changed anyway for the sake of scoping.

app/src/test/kotlin/com/waz/zclient/core/network/AccessTokenRepositoryTest.kt
gizemb: thanks :)

app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
```diff
@@ -23,33 +25,60 @@ object NetworkDependencyProvider {
...
+        }.build()
+
+    fun createHttpClientForToken(): OkHttpClient =
+        OkHttpClient.Builder().apply {
+            if (BuildConfig.DEBUG) {
+                addLoggingInterceptor()
             }
         }.build()
+
+    private fun OkHttpClient.Builder.addLoggingInterceptor() {
```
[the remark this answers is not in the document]
gizemb: done. also changed the method a bit to apply builder convention (since the class is a builder). now the code above does not need apply blocks and looks cleaner imo.

app/src/main/res/xml/network_security_config.xml
android10: As soon as we don't open another door for breaking the security of the app, which is a core component of our client. What is wrong by just using normal logs? I need more arguments on why this is useful. I'm a bit skeptical to let our app to accept user installed certificates.

app/src/main/AndroidManifest.xml
android10: Maybe a topic for discussion? I need to ask more questions about accepting these kind of certificates.

app/src/main/res/xml/network_security_config.xml
gizemb: because i want to be able to manipulate the requests/responses in real time. such as, I want to send an http 401 even if the backend sends me 200 success, to be able to debug this functionality in my app. This also lets you test other cases such as throttling etc. which you can't create only by code.

this is a very powerful testing mechanism imo, and it should have already been enabled for the app.

if you read the issue, the documentation link & the config xml file, you'll see that it has the `<debug-overrides/>` method, which trusts the given certificates only in debug builds. I don't think this is a security concern since it's a developer configuration. I think android framework documentation + qa and dev tests are enough to verify that we don't break anything on production code.

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenRemoteDataSource.kt
```diff
@@ -0,0 +1,12 @@
+package com.waz.zclient.core.network
+
+import com.waz.zclient.core.exception.Failure
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.network.api.token.AccessTokenResponse
+import com.waz.zclient.core.network.api.token.TokenService
+
+class AccessTokenRemoteDataSource(private val tokenService: TokenService) {
```
BradleyWire: we're missing test cases for this class

app/src/test/kotlin/com/waz/zclient/core/network/AccessTokenRepositoryTest.kt
```diff
@@ -0,0 +1,94 @@
...
+
+        val token = accessTokenRepository.refreshToken()
+
+        verify(localDataSource).refreshToken()
+        token shouldBe String.empty()
+    }
+
+    @Test
+    fun `when updateRefreshToken is called, calls localDataSource with given token`() {
+        accessTokenRepository.updateRefreshToken("newToken")
```
BradleyWire: If we're repeating hardcoded strings, i'd used constants

app/src/test/kotlin/com/waz/zclient/core/network/RawResponseRegistryTest.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    @Before
+    fun setUp() {
+        rawResponseRegistry = RawResponseRegistry()
+    }
+
+    @Test
+    fun `given an action is added, performs the action with given response when notifyRawResponse is called`() {
+        val action: (Response<*>) -> Unit = mock(Function1::class.java) as (Response<*>) -> Unit
+        val response= mock(Response::class.java)
```
BradleyWire: formatting

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenRemoteDataSource.kt
gizemb: didn't notice! done.

app/src/test/kotlin/com/waz/zclient/core/network/AccessTokenRepositoryTest.kt
gizemb: done

app/src/test/kotlin/com/waz/zclient/core/network/RawResponseRegistryTest.kt
gizemb: done

app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
```diff
@@ -30,26 +30,51 @@ object NetworkDependencyProvider {
...
-    single { NetworkDependencyProvider.createClient(get(), get()) }
-    single { NetworkDependencyProvider.retrofit(get()) }
-    single { AccessTokenRepository() }
-    single { AuthTokenHandler(get()) }
+    single { createHttpClient(get(), get()) }
+    single { retrofit(get()) }
+    single { AccessTokenRemoteDataSource(get()) }
+    factory {
+        AccessTokenLocalDataSource(
+            //TODO: get from other(global?) module
```
android10: Good Point! I guess this is something to be done once we integrate these changes.

app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
```diff
@@ -30,26 +30,51 @@ object NetworkDependencyProvider {
...
+        OkHttpClient.Builder().addLoggingInterceptor().build()
+
+    private fun OkHttpClient.Builder.addLoggingInterceptor() = this.apply {
+        if (BuildConfig.DEBUG) {
+            addInterceptor(HttpLoggingInterceptor().setLevel(Level.BODY))
+        }
+    }
 }
 
 val networkModule: Module = module {
```
android10: I'm also wondering if it would be a good idea to put inside each feature package its dependency injection module rather than grouping all together in one place. :thinking:

app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
```diff
@@ -30,26 +30,51 @@ object NetworkDependencyProvider {
...
+    single { ThreadHandler() }
     single { NetworkHandler(androidContext()) }
-    single { NetworkDependencyProvider.createClient(get(), get()) }
-    single { NetworkDependencyProvider.retrofit(get()) }
-    single { AccessTokenRepository() }
-    single { AuthTokenHandler(get()) }
+    single { createHttpClient(get(), get()) }
+    single { retrofit(get()) }
+    single { AccessTokenRemoteDataSource(get()) }
+    factory {
```
android10: I do not see a reason to use a ```factory``` here since ```Koin``` will be creating a new instance every time we inject this collaborator, which in this case is not the best case since Network requests happen regularly. A singleton should to the job. Remember that ```@Singleton``` are not bad if they are not saving global state, which turn them into anti-pattern.

app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
```diff
@@ -30,26 +30,51 @@ object NetworkDependencyProvider {
...
-    single { ThreadHandler() }
-    single { ApiService(get(), get(), get()) }
+    single { ApiService(get(), get(), get(), get()) }
+
+    //Token manipulation
+    val apiServiceForToken = "API_SERVICE_FOR_TOKEN"
+    val networkClientForToken = "NETWORK_CLIENT_FOR_TOKEN"
+    single<NetworkClient>(named(networkClientForToken)) { RetrofitClient(retrofit(createHttpClientForToken())) }
+    single(named(apiServiceForToken)) { ApiService(get(), get(), get(named(networkClientForToken)), get()) }
+    factory { get<ApiService>(named(apiServiceForToken)).createApi(TokenApi::class.java) }
```
android10: Same case here: Is there any specific reason to use a ```factory``` instead of a ```single```?

app/src/main/kotlin/com/waz/zclient/core/network/api/token/TokenService.kt
```diff
@@ -0,0 +1,15 @@
...
+import com.waz.zclient.core.network.ApiService
+
+class TokenService(private val apiService: ApiService,
+                   private val tokenApi: TokenApi) {
+
+    fun renewAccessToken(refreshToken: String): Either<Failure, AccessTokenResponse> =
+        apiService.request(
+            tokenApi.access(mapOf("Cookie" to "zuid=$refreshToken")), AccessTokenResponse.EMPTY
+        )
+
```
android10: Extra space here. Hopefully all this can be solved with style check and lint.

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenRemoteDataSource.kt
android10: I like this! You both made my day. Checking for tests existence and writing them. That is the first thing I do when reviewing code. :tada:

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenRepository.kt
android10: This is a good point. After checking this in more details, maybe it is better that instead of manipulating tokens as string, we create some kind of abstraction (interface) ```Token``` which might have 2 implementations: ```AccessToken``` and ```RefreshToken``` because we are only storing on ```SharedPreferences``` the token string literal, but from the ```api``` we get way more fields that could be useful:

```kotlin
data class AccessTokenResponse(
    @SerializedName("access_token")
    val token: String,
    @SerializedName("token_type")
    val type: String,
    @SerializedName("user")
    val userId: String,
    @SerializedName("expires_in")
    val expiresIn: String) {
    ...
}
```
These these classes should contain all these fields and we should serialize them and save them to preferences. When obtaining we map what we get from the preferences to the mentioned classes. 

If we go for this option, we do not need to do it in this PR.

app/src/main/kotlin/com/waz/zclient/core/network/ApiService.kt
```diff
@@ -17,7 +10,8 @@ import retrofit2.Response
 
 class ApiService(private val networkHandler: NetworkHandler,
                  private val threadHandler: ThreadHandler,
-                 private val networkClient: NetworkClient) {
+                 private val networkClient: NetworkClient,
+                 private val rawResponseRegistry: RawResponseRegistry) {
```
android10: Is this for logging the response? If so and it does not have an impact on testing then I would get rid of it and call it statically as a logging good practice. Same way as we use either the Android Logger or Timber. (This should be wrapped in our own implementation)
Also, if logging is the main purpose, unless I'm missing something, the name is not very friendly. Please correct me if I'm wrong.

app/src/main/kotlin/com/waz/zclient/core/network/AuthTokenHandler.kt
```diff
@@ -16,9 +20,15 @@ class AuthTokenHandler(private val tokenRepository: AccessTokenRepository) {
...
+    fun updateRefreshTokenFromResponse(response: Response<*>) =
+        response.headers()[RESPONSE_HEADER_REFRESH_TOKEN_KEY]?.let {
+            if (refreshToken() != it) {
+                updateRefreshToken(it)
+            }
+        }
+
     fun renewAccessToken(refreshToken: String): Either<Failure, String> =
-        tokenRepository.renewAccessToken(refreshToken)
+        tokenRepository.renewAccessToken(refreshToken).map { it.token }
```
android10: This was my point regarding that we receive way more info from the api and we are only using the token string literal. 
At scala level, do you know if we are storing all this info?

app/src/main/kotlin/com/waz/zclient/core/network/RawResponseRegistry.kt
```diff
@@ -0,0 +1,16 @@
+package com.waz.zclient.core.network
+
+import retrofit2.Response
+
+class RawResponseRegistry {
```
android10: I need a better understanding of this class.

app/src/main/kotlin/com/waz/zclient/core/network/AuthTokenHandler.kt
```diff
@@ -16,9 +20,15 @@ class AuthTokenHandler(private val tokenRepository: AccessTokenRepository) {
     fun refreshToken() = tokenRepository.refreshToken()
     fun updateRefreshToken(newRefreshToken: String) = tokenRepository.updateRefreshToken(newRefreshToken)
 
+    fun updateRefreshTokenFromResponse(response: Response<*>) =
```
android10: What is the difference between this method and ```updateRefreshToken```? It looks like from an architectural perspective there should not be anything that tells us what is the source of information: that is the purpose of the repository pattern. 
Maybe we can re think this a little bit more?

app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
gizemb: you mean separating networkModule like, tokenModule, (core)NetworkModule etc?

app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
gizemb: well it's like a singleton in practice since the repository class that uses this is a singleton. converted it.

app/src/main/kotlin/com/waz/zclient/core/di/NetworkModule.kt
gizemb: same as above. changed that too.

app/src/main/kotlin/com/waz/zclient/core/network/api/token/TokenService.kt
gizemb: yeap, there's a rule that checks extra spaces.

app/src/main/kotlin/com/waz/zclient/core/network/ApiService.kt
gizemb: no, not for logging.

this is the logic i tried to implement here:
_"update the refresh token whenever a response header has "Cookie""_

I thought if we do it in the ApiService level, then no clients using ApiService would have to deal with it. `performRequest` method here returns the `body` of the response (line 33) but i need the `header` to be able to extract the "Cookie". The first thing i thought was to pass the `AuthTokenHandler` to `ApiService` as a dependency, and call `authTokenHandler.updateRefreshToken(extractCookieFromResponse(response))` whenever i get a response. But this solution has 2 problems:
1. Gives `ApiService` class extra responsibility about token logic
2. Creates a cyclic dependency (ApiService -> AuthTokenHander -> AccessTokenRepository -> AccessTokenRemoteDataSource -> TokenService -> ApiService again)

so what i did is I created a mediator class `RawResponseRegistry` that holds some `listeners`. On one hand, there are classes that registers themselves as listeners to `RawResponseRegistry` (currently there's only `AuthTokenHandler`). These listeners are notified whenever `notifyRawResponseReceived` method is called. 
On the other hand, here in `ApiService`, whenever I get a successful response i tell RawResponseRegistry to notify the listeners that are registered to it (via `notifyRawResponseReceived`). 
So, once `AuthTokenHandler` registers itself to this class, it gets notified from then on when a new response is received, and can check the header of the response whether it contains a "Cookie" information.

app/src/main/kotlin/com/waz/zclient/core/network/AuthTokenHandler.kt
gizemb: yes, we are. I'll change the models and create a repository model that we can use in the coming PRs.

app/src/main/kotlin/com/waz/zclient/core/network/RawResponseRegistry.kt
gizemb: tried to explain above.

app/src/main/kotlin/com/waz/zclient/core/network/AuthTokenHandler.kt
gizemb: then I'll try to implement some mapper in between since we need to extract the "Cookie" information from response somewhere here.

gizemb: but this is not the repository class, it's the handler. if there's no logic inside in this class, then this would be a 1:1 mapping of the repository class, wouldn't it?
I can create a mapper here, but I think I didn't understand the purpose of this class, let's discuss it.

android10:
@gizemb good Job! :heart: 
Let's make sure we do the proper refactor in another PR as agreed.

PR #2513, opened 2019-12-27
Experiment in FP

makingthematrix:
TBD

https://docs.google.com/document/d/1iuxDLg-6Epu6e-WHJb7jkLHd8tS7Dg653f1TE6YqxDQ/edit?folder=0ABayswLXSBGhUk9PVA
#### APK
[Download build #771](http://10.10.124.11:8080/job/Pull%20Request%20Builder/771/artifact/build/artifact/wire-dev-PR2513-771.apk)

### DIFF
--- a/app/src/main/kotlin/com/waz/zclient/settings/main/list/SettingsMainListFactory.kt
+++ b/app/src/main/kotlin/com/waz/zclient/settings/main/list/SettingsMainListFactory.kt
@@ -9,17 +9,16 @@ import com.waz.zclient.settings.main.model.SettingsMainItem
 class SettingsMainListFactory {
 
     companion object {
-        fun generateList(context: Context): List<SettingsMainItem> =
-                listOf(
-                        context.stringArrayFromResource(R.array.settings_titles)
-                                .zip(context.stringArrayFromResource(R.array.settings_icons)),
-                        if (Config.developerSettingsEnabled())
-                            context.stringArrayFromResource(R.array.settings_developer_titles)
-                                    .zip(context.stringArrayFromResource(R.array.settings_developer_icons))
-                        else emptyList()
-                ).flatten()
-                 .map {
-                    SettingsMainItem(it.first, it.second)
-                }
+        fun generateList(context: Context): List<SettingsMainItem> {
+            val titles =
+                    context.stringArrayFromResource(R.array.settings_titles)
+                        .zip(context.stringArrayFromResource(R.array.settings_icons))
+            val devTitles =
+                    if (Config.developerSettingsEnabled())
+                        context.stringArrayFromResource(R.array.settings_developer_titles)
+                                .zip(context.stringArrayFromResource(R.array.settings_developer_icons))
+                    else emptyList()
+            return listOf(titles, devTitles).flatten().map { SettingsMainItem(it.first, it.second) }
+        }
     }
 }
--- a/app/src/main/kotlin/com/waz/zclient/settings/main/list/SettingsMainListFactory.kt
+++ b/app/src/main/kotlin/com/waz/zclient/settings/main/list/SettingsMainListFactory.kt
@@ -9,22 +9,17 @@ import com.waz.zclient.settings.main.model.SettingsMainItem
 class SettingsMainListFactory {
 
     companion object {
-        fun generateList(context: Context): List<SettingsMainItem> {
-            val titles = context.stringArrayFromResource(R.array.settings_titles)
-            val icons = context.stringArrayFromResource(R.array.settings_icons)
-            val developerTitles = context.stringArrayFromResource(R.array.settings_developer_titles)
-            val developerIcons = context.stringArrayFromResource(R.array.settings_developer_icons)
-
-            val settingItems = ArrayList<SettingsMainItem>()
-            for (i in titles.indices) {
-                settingItems.add(SettingsMainItem(titles[i], icons[i]))
-            }
-            if (Config.developerSettingsEnabled()) {
-                for (i in developerTitles.indices) {
-                    settingItems.add(SettingsMainItem(developerTitles[i], developerIcons[i]))
+        fun generateList(context: Context): List<SettingsMainItem> =
+                listOf(
+                        context.stringArrayFromResource(R.array.settings_titles)
+                                .zip(context.stringArrayFromResource(R.array.settings_icons)),
+                        if (Config.developerSettingsEnabled())
+                            context.stringArrayFromResource(R.array.settings_developer_titles)
+                                    .zip(context.stringArrayFromResource(R.array.settings_developer_icons))
+                        else emptyList()
+                ).flatten()
+                 .map {
+                    SettingsMainItem(it.first, it.second)
                 }
-            }
-            return settingItems
-        }
     }
 }
--- a/storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
+++ b/storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
@@ -6,119 +6,105 @@ import androidx.room.migration.Migration
 import androidx.sqlite.db.SupportSQLiteDatabase
 import org.json.JSONObject
 
+import kotlin.collections.*
+
+import com.waz.zclient.storage.db.clients.migration.Option.*
+import com.waz.zclient.storage.db.clients.migration.ContentKey.*
+import com.waz.zclient.storage.db.clients.migration.ContentValue.*
+
 class ClientsTableMigration : Migration(125, 126) {
 
     override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL("""
+            CREATE TABLE '$NEW_CLIENT_TABLE_NAME' 
+            ('${CLIENT_ID_KEY.str}' TEXT NOT NULL, 
+             '${NEW_CLIENT_TIME_KEY.str}' NUMERIC, 
+             '${CLIENT_LABEL_KEY.str}' TEXT, 
+             '${NEW_CLIENT_TYPE_KEY.str}' TEXT, 
+             '${CLIENT_MODEL_KEY.str}' TEXT, 
+             '${CLIENT_LOCATION_LAT_KEY.str}' LONG, 
+             '${CLIENT_LOCATION_LONG_KEY.str}' DOUBLE, 
+             '${NEW_CLIENT_LOCATION_NAME_KEY.str}' TEXT, 
+             '${CLIENT_VERIFICATION_KEY.str}' TEXT, 
+             '${CLIENT_ENC_KEY.str}' TEXT, 
+             '${CLIENT_MAC_KEY.str}' TEXT, 
+             PRIMARY KEY('${CLIENT_ID_KEY.str}')
+            )""".trimIndent())
 
-        // "Clients" to "client" migration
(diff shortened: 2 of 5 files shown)


app/src/main/kotlin/com/waz/zclient/settings/main/list/SettingsMainListFactory.kt
```diff
@@ -10,21 +10,15 @@ class SettingsMainListFactory {
```
makingthematrix: There are two issues here which in this particular case can be overlooked, but they give us a hint that similar things may happen in other places in the code. 
1. The code treats `titles` and `icons` as separate arrays and assume they are of the same length. The app may fail at line 20 if the length of the `icons` array is smaller than `titles`.
2. The code retrieves `developerTitles` and `developerIcons` first, and only then checks if they are necessary.

Ad 1. This can be easily fixed by using the`zip` method on the array. After that we can also switch from `for/in` to `map`. It's a small step, but it already tells us something: search for collection methods before writing your won code.
Ad 2. Here we can use the fact that `if/else` is an expression. If we use developerTitles/Icons, let's retrieve them and zip them. Otherwise, let's use an empty list. Afterwards we need to use `listOf` and `flatten` to make one list of items from two lists of items where one of them can be empty. 
(I didn't find a method to simply join two lists - shouldn't there be one?)

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
```diff
@@ -6,119 +6,105 @@ import androidx.room.migration.Migration
...
+            )""".trimIndent())
+
+        val clientsArray =
+                JSONObject(
+                        database.query("SELECT * FROM $OLD_CLIENT_TABLE_NAME")
+                                .getString(DATA_COLUMN_INDEX)
+                )
+                .getJSONArray(OLD_CLIENTS_CLIENTS_ARRAY_KEY)
+
+        for (i in 0 until clientsArray.length())
```
makingthematrix: `clientsArray` is not really an array, so we can't use `foreach` here.

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
```diff
@@ -6,119 +6,105 @@ import androidx.room.migration.Migration
...
+             '${CLIENT_LOCATION_LAT_KEY.str}' LONG, 
+             '${CLIENT_LOCATION_LONG_KEY.str}' DOUBLE, 
+             '${NEW_CLIENT_LOCATION_NAME_KEY.str}' TEXT, 
+             '${CLIENT_VERIFICATION_KEY.str}' TEXT, 
+             '${CLIENT_ENC_KEY.str}' TEXT, 
+             '${CLIENT_MAC_KEY.str}' TEXT, 
+             PRIMARY KEY('${CLIENT_ID_KEY.str}')
+            )""".trimIndent())
+
+        val clientsArray =
```
makingthematrix: Readability: 
There's no hard rule here. Sometimes it's more readable to nest expressions one into another and just use new lines and indentation. Sometimes it's more readable to take the nested expressions out and assign them to local `val`s. Sometimes it even makes sense to use a local variable if that saves us from an unnecessarily complex alternative.

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
```diff
@@ -6,119 +6,105 @@ import androidx.room.migration.Migration
...
+                        database.query("SELECT * FROM $OLD_CLIENT_TABLE_NAME")
+                                .getString(DATA_COLUMN_INDEX)
+                )
+                .getJSONArray(OLD_CLIENTS_CLIENTS_ARRAY_KEY)
+
+        for (i in 0 until clientsArray.length())
+            database.insert(
+                    NEW_CLIENT_TABLE_NAME,
+                    OnConflictStrategy.REPLACE,
+                    parseClientData(clientsArray.getJSONObject(i)).toContentValues()
```
makingthematrix: `toContentValues` is a mix-in method. I like them, but I agree that sometimes they may make code unclear, esp. during debugging. This like could also look like this:
```
toContentValues(
    parseClientData(clientsArray.getJSONObject(i))
)
```

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ContentKey.kt
```diff
@@ -0,0 +1,25 @@
+package com.waz.zclient.storage.db.clients.migration
+
+enum class ContentKey(val str: String) {
```
makingthematrix: Type safety:
Keys as enums wrapping strings instead of strings themselves.

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ContentValue.kt
```diff
@@ -0,0 +1,13 @@
+package com.waz.zclient.storage.db.clients.migration
```
makingthematrix: Type safety:
Wrapper data classes extending a common sealed class. Mixed with `when/in` and overloaded `apply` methods makes the code more concise while improving safety.

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/Option.kt
```diff
@@ -0,0 +1,28 @@
+package com.waz.zclient.storage.db.clients.migration
```
makingthematrix: I'd change a few things in how this class is written, but it serves the purpose as it is now. By implementing `Iterable` I make it possible to `flatMap` over options - they work basically as lists containing 1 or 0 elements. This way it is possible to create a list with `listOf` from elements which sometimes can be null, and then flatten it.

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
```diff
@@ -6,119 +6,105 @@ import androidx.room.migration.Migration
...
+                                getString(it, CLIENT_MAC_KEY)
+                        )
+                    }.flatten()
+
+                    listOf(main, location, signaling).flatten().toMap()
+                }
+    }
+}
+
+fun Map<ContentKey, ContentValue>.toContentValues(): ContentValues {
```
makingthematrix: This converts a map of properties taken from JSON to plain `ContentValues` used later in SQL. Each value has one of three types which all extend a sealed class, so we are sure no other types are possible and we unwrap them with `when/is`.

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
```diff
@@ -6,119 +6,105 @@ import androidx.room.migration.Migration
...
+        private fun getDouble(json: JSONObject, key: ContentKey) =
+                if (json.has(key.str)) Some(key to DoubleValue(json.getDouble(key.str))) else None
+
+        private fun getLong(json: JSONObject, key: ContentKey) =
+                if (json.has(key.str)) Some(key to LongValue(json.getLong(key.str))) else None
+
+        private fun getJSONObject(json: JSONObject, key: ContentKey) =
+                if (json.has(key.str)) Some(json.getJSONObject(key.str)) else None
+
+        private fun parseClientData(client: JSONObject?) =
```
makingthematrix: `client` can be null and in this case we want to return an empty map. Otherwise, we build a map of properties, in some cases going one level deeper, and with each property we take into account that it can be null.

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
android10: The question is: If ```clientsArray``` is not an array, maybe we can rename to ```clientsCollection``` or ```clientsList```?

app/src/main/kotlin/com/waz/zclient/settings/main/list/SettingsMainListFactory.kt
android10: I agree with the ```map``` thing, especially since there is a correlation between titles and icons.

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
android10: What do you mean by a mix-in method?

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
android10: Here you are using the Scala ```Some``` and ```None``` right?

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
gizemb: it's a JSONArray so i think naming makes sense.
if we really want to use forEach, we can write an extension function. @android10 would that be a good use case for extension functions?

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
```diff
@@ -6,119 +6,105 @@ import androidx.room.migration.Migration
...
+             '${NEW_CLIENT_TYPE_KEY.str}' TEXT, 
+             '${CLIENT_MODEL_KEY.str}' TEXT, 
+             '${CLIENT_LOCATION_LAT_KEY.str}' LONG, 
+             '${CLIENT_LOCATION_LONG_KEY.str}' DOUBLE, 
+             '${NEW_CLIENT_LOCATION_NAME_KEY.str}' TEXT, 
+             '${CLIENT_VERIFICATION_KEY.str}' TEXT, 
+             '${CLIENT_ENC_KEY.str}' TEXT, 
+             '${CLIENT_MAC_KEY.str}' TEXT, 
+             PRIMARY KEY('${CLIENT_ID_KEY.str}')
+            )""".trimIndent())
```
gizemb: looks definitely better :)

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
gizemb: I'm not sure about introducing `Option`, `Some` and `None` to Kotlin. I like it better when I don't have to wrap my variables inside classes like that. We can achieve the same "fold" logic here with 
`client?.let { //do sth } ?: emptyMap()`
let's discuss if they have any other advantages

mejdoo: I agree with Gizem. It would be better to write Kotlin idiomatic code.

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
makingthematrix: Sorry, an extension method in Kotlin. Or an implicit method in Scala.

storage/src/main/kotlin/com/waz/zclient/storage/db/clients/migration/ClientsTableMigration.kt
makingthematrix: I use `Some` and `None` from the `Option` class I wrote. If there is a more Kotlin-idiomatic way to do it, sure, let's talk about it. But I also wanted to show how similar this solution is to `Either` and to collections.

Issue #762, opened 2017-04-02
Are conversations with Otto the bot private?


patrickwelker:
I just wondered if you gather any data from Otto the Bot conversations? My better half started to send herself passwords and other personal stuff via Otto. I just wanted to confirm that this is not the best idea.

mykola-mokhnach:
The bot immediately deletes all the received and decrypted strings on his side. It only uses them to generate reply if needed. Also we don't keep any logs or anything, which may affect your privacy in dialog with bot.

ImAnnoying2:
Why not close this issue if the question has been answered?

PR #758, opened 2017-03-31
[AN-5074] New single image fragment with action buttons

JPonte:
Ticket: https://wearezeta.atlassian.net/browse/AN-5074
#### APK
[Download build #8727](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8727/artifact/build/artifact/wire-dev-PR758-8727.apk)
[Download build #8733](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8733/artifact/build/artifact/wire-dev-PR758-8733.apk)
[Download build #8734](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8734/artifact/build/artifact/wire-dev-PR758-8734.apk)
[Download build #8741](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8741/artifact/build/artifact/wire-dev-PR758-8741.apk)
[Download build #8742](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8742/artifact/build/artifact/wire-dev-PR758-8742.apk)
[Download build #8747](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8747/artifact/build/artifact/wire-dev-PR758-8747.apk)
[Download build #8790](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8790/artifact/build/artifact/wire-dev-PR758-8790.apk)

Issue #754, opened 2017-03-30
Feature request: comfort noise


fortran77:
On Wire audio calls on an Android smartphone, I've sometimes heard the other person ask "Are you still there?", and I have done the same myself. When neither party is talking, if the codecs do silence suppression, i.e., transmit only a silence signal, it becomes impossible for the callers to distinguish between an active connection and a dropped connection.

In the telephone and cellphone world, the solution is for each end to generate comfort noise, i.e., a low level of background noise indicating to the person on the call that the connection is still active.

Wire should consider doing this.

jspittka:
Thanks for this suggestion! We consciously decided against CNG, that's why Wire calls are perceived as very crisp and natural. CNG works great in the old telephone world but for wideband calls most users perceive CNG as annoying.

Our approach to signal users that a call is up or has gone down is by playing audio notifications. Otherwise reliability should be just so high that the question "are you still there" is not needed. Some of the audio notifications may have been broken during the roll out of the new calling protocol. We will review and fix potential defects soon.

PR #2516, opened 2019-12-30
Settings account - part1

mejdoo:
## What's new in this PR?

This is a PR for the settings account screen.  In part 1, I have implemented the UI and business logic for update name.

Please review but don't merge. There is issue to fix with PUT request using retrofit.
#### APK
[Download build #742](http://10.10.124.11:8080/job/Pull%20Request%20Builder/742/artifact/build/artifact/wire-dev-PR2516-742.apk)
[Download build #754](http://10.10.124.11:8080/job/Pull%20Request%20Builder/754/artifact/build/artifact/wire-dev-PR2516-754.apk)
[Download build #758](http://10.10.124.11:8080/job/Pull%20Request%20Builder/758/artifact/build/artifact/wire-dev-PR2516-758.apk)
[Download build #762](http://10.10.124.11:8080/job/Pull%20Request%20Builder/762/artifact/build/artifact/wire-dev-PR2516-762.apk)
[Download build #768](http://10.10.124.11:8080/job/Pull%20Request%20Builder/768/artifact/build/artifact/wire-dev-PR2516-768.apk)


app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
```diff
@@ -52,36 +55,42 @@ class SettingsAccountFragment : Fragment() {
...
-        preferences_account_name_title.text = userProfileItem.name
-        preferences_account_email_title.text = userProfileItem.email
-        preferences_account_handle_title.text = userProfileItem.handle
-        preferences_account_phone_title.text = userProfileItem.phone
+        with(userProfileItem) {
+            preferences_account_name_title.text = name
+            preferences_account_handle_title.text = handle
+
+            if (!email.isNullOrEmpty()) preferences_account_email_title.text = email
+            else preferences_account_email_title.text = getString(R.string.pref_account_add_email_title)
```
gizemb: this looks somewhat cleaner:
`preferences_account_email_title.text = if (!email.isNullOrEmpty) email else getString(...)`

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
```diff
@@ -52,36 +55,42 @@ class SettingsAccountFragment : Fragment() {
...
+
+            if (!phone.isNullOrEmpty()) preferences_account_phone_title.text = phone
```
gizemb: same here

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
```diff
@@ -52,36 +55,42 @@ class SettingsAccountFragment : Fragment() {
     }
 
     private fun setupListeners() {
+        preferences_account_name.setOnClickListener {
+            val editNameDialogFragment: EditTextDialogFragment = EditTextDialogFragment.newInstance(
+                preferences_account_name_title.text.toString())
+            editNameDialogFragment.listener = this
+            editNameDialogFragment.show(requireActivity().supportFragmentManager, "")
```
gizemb: can also be:
```
val editNameDialogFragment: EditTextDialogFragment = EditTextDialogFragment.newInstance(
                preferences_account_name_title.text.toString()).also {
                       it.listener = this
                }.show(...)
```
also, do we really want to user _activity's_ fragmentManager? then, we're manipulating activity's ui from fragment again, which i am not a fan as you know :)

app/src/main/res-v2/layout/dialog_fragment_edit_text.xml
```diff
@@ -0,0 +1,34 @@
...
+    This program is distributed in the hope that it will be useful,
+    but WITHOUT ANY WARRANTY; without even the implied warranty of
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    GNU General Public License for more details.
+
+    You should have received a copy of the GNU General Public License
+    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+-->
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
```
gizemb: ignore if this page is going to change but, no need to wrap into linearlayout for now.

app/src/main/kotlin/com/waz/zclient/core/network/SafeApiDataSource.kt
```diff
@@ -58,3 +58,16 @@ private fun <R> performFallback(fallbackRequest: suspend () -> Either<Failure, R
             }
         }
     }
+
+suspend fun <R> saveData(saveToNetwork: suspend () -> Either<Failure, R>,
```
gizemb: is there really no other way to reuse `performFallback` method instead of this?

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/EditTextDialogFragment.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.waz.zclient.core.extension.withArgs
+
+class EditTextDialogFragment : DialogFragment() {
+
+    var listener: EditTextDialogFragmentListener? = null
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+
+        val initialValue = arguments?.getString(DEFAULT_TEXT_BUNDLE_KEY, "")
+        val inflater = requireActivity().layoutInflater
```
gizemb: is it the same inflater that is passed to `onCreateView`? if not, is there any difference?

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/EditTextDialogFragment.kt
```diff
@@ -0,0 +1,48 @@
...
+        val builder = AlertDialog.Builder(requireContext())
+        val view = inflater.inflate(R.layout.dialog_fragment_edit_text, null)
```
gizemb: again, there's a container in `onCreateView`. could it be useful somehow (in terms of measuring, context, theming, etc)?

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/EditTextDialogFragment.kt
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        builder.setTitle(getString(R.string.pref_account_edit_name_title))
+        builder.setView(view)
+        builder.setPositiveButton(getString(android.R.string.ok)) { _, _ ->
+            val newValue = editText.text.toString().trim()
+            listener?.onTextEdited(newValue)
+            dismiss()
+        }
+        builder.setNegativeButton(getString(android.R.string.cancel)) { _, _ -> dismiss() }
```
gizemb: since these are builder methods, they can be chained i guess?

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
mejdoo: done

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
mejdoo: Please prepare for us a small proof of concept how to avoid calling activity methods from fragments...

app/src/main/res-v2/layout/dialog_fragment_edit_text.xml
mejdoo: done

app/src/main/kotlin/com/waz/zclient/core/network/SafeApiDataSource.kt
mejdoo: It's not exactly the same implementation specially for `saveToDatabase()`, but we can optimise it.

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/EditTextDialogFragment.kt
mejdoo: yes done

app/src/main/kotlin/com/waz/zclient/core/network/SafeApiDataSource.kt
BradleyWire: Could we not move the onFailure {} logic to a private method atleast as that seems duplicated?

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/EditTextDialogFragment.kt
```diff
@@ -0,0 +1,45 @@
...
+        }
+
+        builder.setTitle(getString(R.string.pref_account_edit_name_title)).setView(view).setPositiveButton(getString(android.R.string.ok)) { _, _ ->
+            val newValue = editText.text.toString().trim()
+            listener?.onTextEdited(newValue)
+            dismiss()
+        }.setNegativeButton(getString(android.R.string.cancel)) { _, _ -> dismiss() }
+
+
+        return builder.create()
```
BradleyWire: Can we not return the built object? Rather than assigning the builder and creating the object on different lines? I would also put each chained call onto a new line for readability.

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
```diff
@@ -52,36 +57,41 @@ class SettingsAccountFragment : Fragment() {
     }
 
     private fun setupListeners() {
+        preferences_account_name.setOnClickListener {
```
BradleyWire: Couldn't we do something like initPreferencesAccountTitle() and initAccountResetPassword() for readability so we can define the characteristics for the particular UI in their own private methods. If we're looking to read something around the account title it'd probably be better to read it in one place rather than multiple different methods.

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
```diff
@@ -52,36 +57,41 @@ class SettingsAccountFragment : Fragment() {
...
-            settingsAccountViewModel.loadData()
+            settingsAccountViewModel.loadProfile()
         }
     }
 
     private fun showErrorMessage(errorMessage: String) {
         Toast.makeText(requireContext(), errorMessage, Toast.LENGTH_LONG).show()
     }
 
-    private fun updateLoadingVisibility(isLoading: Boolean?) {
```
BradleyWire: Let's have a discussion with Victor (for the long-term solution) around the UI for when content takes a while to load. It'd be good to notify the user when we're loading content.

app/src/main/kotlin/com/waz/zclient/user/data/source/remote/model/UserApi.kt
```diff
@@ -4,29 +4,29 @@ import com.google.gson.annotations.SerializedName
 
 data class UserApi(
     @SerializedName("id")
-    var id: String,
+    var id: String? = null,
```
BradleyWire: Id, name aren't optional on the api. How come they need to be here?

mejdoo: I"m using now this data class with retrofit @Body param for update name/handle/phone (PUT requests). Do you suggest to create a new data class instead and keep this as it is?

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
mejdoo: let's dot it

app/src/main/kotlin/com/waz/zclient/user/data/source/remote/model/UserApi.kt
BradleyWire: No, I was just curious more than anything

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
BradleyWire: I've just realised why loading isn't accounted for as we in the spike are relying on api calls but when this hits production we'll be retrieving the latest data from the database so the loading is essentially the latest data stored on the device. No discussion is needed with Victor.

app/src/main/kotlin/com/waz/zclient/core/network/Network.kt
```diff
@@ -15,6 +15,7 @@ open class Network {
 
                 val newRequest = chain.request().newBuilder()
                     .addHeader("Content-Type", "application/json")
+                    .addHeader("Accept", "application/json")
```
android10: is this something we need to all the api requests? if so, we might have to implement it at the networking level. cc @gizemb

app/src/main/kotlin/com/waz/zclient/core/network/SafeApiDataSource.kt
```diff
@@ -7,18 +7,21 @@ import com.waz.zclient.core.functional.onSuccess
...
             val body = response.body()
-            body?.let {
-                return Either.Right(body)
-            }
+            body?.let { return Either.Right(body) }
         }
         return Either.Left(HttpError(response.code(), response.message()))
-    } catch (e: Exception) {
+    }
+    catch (e: EOFException) {
```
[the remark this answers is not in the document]
android10: @mejdoo interesting point. As we discussed, when integrating #2510 we will propose a solution and there are going to be issues that will come up.

app/src/main/kotlin/com/waz/zclient/core/network/SafeApiDataSource.kt
```diff
@@ -58,3 +61,16 @@ private fun <R> performFallback(fallbackRequest: suspend () -> Either<Failure, R
             }
         }
     }
+
+suspend fun <R> saveData(saveToNetwork: suspend () -> Either<Failure, R>,
+                         saveToDatabase: suspend () -> Either<Failure, R>): Either<Failure, R> =
+    with(saveToNetwork()) {
```
android10: ```saveToNetwork``` is a confusing name. What is ```saveToNetwork``` suppose to do?

app/src/main/kotlin/com/waz/zclient/core/network/SafeApiDataSource.kt
```diff
@@ -58,3 +61,16 @@ private fun <R> performFallback(fallbackRequest: suspend () -> Either<Failure, R
...
+        onSuccess { runBlocking { saveToDatabase() } }
+        onFailure {
+            when (it) {
+                is NetworkServiceError -> Timber.e("Network request failed with generic error ")
```
android10: This is something I would like to do. We should not use 3rd party logging libraries without encapsulating them. 
How do we test this?

This is out of scope of this PR but I would like to see some kind of collaborator ```Logger``` that wraps this ```Timber``` static object.

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/EditTextDialogFragment.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    var listener: EditTextDialogFragmentListener? = null
+
+    override fun onCreateDialog(savedInstanceState: Bundle?): Dialog {
+
+        val initialValue = arguments?.getString(DEFAULT_TEXT_BUNDLE_KEY, "")
+        val inflater = requireActivity().layoutInflater
+        val builder = AlertDialog.Builder(requireContext())
+        val view = inflater.inflate(R.layout.dialog_fragment_edit_text, null)
+        val editText: EditText = view.findViewById(R.id.edit_text) as EditText
```
android10:
```
apply plugin: 'kotlin-android'
apply plugin: 'kotlin-android-extensions'
```
We definitely need to get rid of these ```findViewById``` asap plus this casting.

android10: Something to discuss soon.

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/EditTextDialogFragment.kt
```diff
@@ -0,0 +1,45 @@
...
+            val newValue = editText.text.toString().trim()
+            listener?.onTextEdited(newValue)
+            dismiss()
+        }.setNegativeButton(getString(android.R.string.cancel)) { _, _ -> dismiss() }
+
+
+        return builder.create()
+    }
+
+    companion object {
```
android10: This could Include the listener as a parameter and make this private:
```kotlin
private var listener: EditTextDialogFragmentListener? = null
```

We want to encapsulate and do not let external clients to modify internals from a class. Plus the ```newInstance``` is not really returning any new instance here. 

This is how the code should look like:
```kotlin
companion object {
        fun newInstance(defaultValue: String, dialogListener: EditTextDialogFragmentListener): EditTextDialogFragment {
            return EditTextDialogFragment().withArgs {
                putString(
                    DEFAULT_TEXT_BUNDLE_KEY,
                    defaultValue
                )
            }.also { it.listener = dialogListener }
        }
        private const val DEFAULT_TEXT_BUNDLE_KEY = "defaultTextBundleKey"
    }
```

And when you call this method from ```SettingsAccountFragment```:
```kotlin
preferences_account_name.setOnClickListener {
            EditTextDialogFragment.newInstance(preferences_account_name_title.text.toString(),
                this).show(requireActivity().supportFragmentManager, "")
        }
```

android10: also the ```setOnClickListener``` is not actually setting a listener...it is showing a ```dialog``` :thinking:

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/EditTextDialogFragment.kt
mejdoo: I agree

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
mejdoo: Good catch

app/src/main/kotlin/com/waz/zclient/user/data/mapper/UserMapper.kt
```diff
@@ -36,13 +36,13 @@ class UserMapper {
...
     fun toUserDao(user: User) = with(user) {
-        UserDao(id = id, teamId = teamId, name = name,
+        UserDao(id = id.toString(), teamId = teamId, name = name.toString(),
             handle = handle, email = email, phone = phone,
             trackingId = trackingId, picture = picture,
             accentId = accentId, sKey = sKey,
             connection = connection, connectionTimestamp = connectionTimestamp,
             connectionMessage = connectionMessage, conversation = conversation, relation = relation,
-            timestamp = timestamp, displayName = displayName, verified = verified, deleted = deleted,
+            timestamp = timestamp, displayName = displayName, verified = verified, deleted = deleted!!.toInt(),
```
gizemb: are we sure that `deleted` won't be null here?

app/src/test/kotlin/com/waz/zclient/user/data/source/local/UserLocalDataSourceTest.kt
```diff
@@ -71,8 +71,40 @@ class UserLocalDataSourceTest {
 
     }
 
+    @Test
```
gizemb: please format these. too many unnecessary new lines here. 
the outermost `{}`s can be omitted since there's only runBlocking block inside.
given what, and after doing what, we assert changeName() is successful? please provide more meaningful method names. I will also propose test method names in the form of "given when then" today in the meeting.

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
gizemb: you can get activity's viewmodel and notify it with some event. then, the viewmodel decides what to do with the event.
alternatively, activity's viewmodel can observe fragment's viewmodel and gets notified by livedata. then again, activity's viewmodel would decide what to do.

android10:
@mejdoo is this PR still valid? Or it was replaced by #2530 ? If so, please close this one.

mejdoo:
@android10 I will close It as soon as we make sure that we will use Flow API.

BradleyWire:
Closing this as too many changes have been done in other PRs and are an improvement.

Issue #794, opened 2017-04-14
Add a possibility to save received images (photos) to a memory like in Signal, Telegram or other messengers


ghsrcgh:
Please, add a possibility to save received images (photos) to a memory like in Signal, Telegram or other messengers

mykola-mokhnach:
You can save an image by long tapping it and selecting Save. Would you like something else to be implemented?

ghsrcgh:
I have no this option, tried a lot of times, before created ticket.
I tried to see it with all access rights (restricted) removed. Then I had enabled access to the Memory and tried again. Nothing available.

mykola-mokhnach:
Could you please provide a screenshot of the view, where you expect to see the option?

ghsrcgh:
I checked again and understand the issue.
1. There is no possibility to save an image, if it is opened on a full screen. You may tap so long as you want, but nothing appears. I'm awaiting a possibility to save like it made in Telegram or Signal (there is a ":" menu or Save icon). It is obviously for user and working well (understandable usecase). Anyway, I want to have a possibility to save from a full screen view.
2. If you long taping on a preview, yes, menu appears and I able to select a "Save".

makingthematrix:
I think this was a bug that was fixed a long time ago. I will close this ticket now.

PR #2523, opened 2020-01-06
Refactor UserData.displayName

makingthematrix:
From time to time a bug comes up in our UI, usually about that the user name does not appear correctly in one place or another. After investigation, we find out that we used `UserData.name` or
`UserData.displayName` instead of `UserData.getDisplayName`.
This PR removes both`UserData.displayName` and  `UserData.getDisplayName`, and adds a migration to remove the column from the Users table in DB as well. From now on we can safely use `name` in all cases.
#### APK
[Download build #786](http://10.10.124.11:8080/job/Pull%20Request%20Builder/786/artifact/build/artifact/wire-dev-PR2523-786.apk)
[Download build #794](http://10.10.124.11:8080/job/Pull%20Request%20Builder/794/artifact/build/artifact/wire-dev-PR2523-794.apk)

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/content/UsersStorage.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/content/UsersStorage.scala
@@ -117,27 +117,27 @@ class UsersStorageImpl(context: Context, storage: ZmsDatabase)
     cs.get(user.id) foreach { n =>
       if (n.first != name.first) {
         contactsByName.removeBinding(n.first, user.id)
-        displayNameUpdater ! n.first
+        nameUpdater ! n.first
       }
     }
 
     cs(user.id) = name
     contactsByName.addBinding(name.first, user.id)
-    displayNameUpdater ! name.first
+    nameUpdater ! name.first
     name
   }
 
-  val displayNameUpdater: SerialProcessingQueue[String] = new SerialProcessingQueue[String]({ firstNames =>
+  val nameUpdater: SerialProcessingQueue[String] = new SerialProcessingQueue[String]({ firstNames =>
     contactNameParts map { cs =>
       firstNames.toSet foreach { (first: String) =>
-        updateDisplayNamesWithSameFirst(contactsByName.getOrElse(first, Set()).toSeq, cs)
+        updateNamesWithSameFirst(contactsByName.getOrElse(first, Set()).toSeq, cs)
       }
     }
-  }, "UsersDisplayNameUpdater")
+  }, "UsersNameUpdater")
 
-  def updateDisplayNamesWithSameFirst(users: Seq[UserId], cs: mutable.HashMap[UserId, NameParts]): Unit = {
-    def setFullName(user: UserId) = update(user, { (u : UserData) => u.copy(dispName = u.name) })
-    def setDisplayName(user: UserId, name: String) = update(user, (_: UserData).copy(dispName = Name(name)))
+  def updateNamesWithSameFirst(users: Seq[UserId], cs: mutable.HashMap[UserId, NameParts]): Unit = {
+    def setFullName(user: UserId) = update(user, { (u : UserData) => u.copy(name = u.name) })
+    def setDisplayName(user: UserId, name: String) = update(user, (_: UserData).copy(name = Name(name)))
 
     if (users.isEmpty) CancellableFuture.successful(())
     else if (users.size == 1) {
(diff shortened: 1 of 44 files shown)


makingthematrix:
build again

makingthematrix:
@android10 
The original code was there for a long time as a sort of a preparation for a feature where users can change names of their contacts. So `name` would be the original name of the contact, and `displayName` would be the name made up by the user. Then, `getDisplayName()` would be used to get one or the other (`displayName`, or if it's not set, then `name`).
But the functionality was never finished and tbh it was never clear which one of the three should be used when. Every few months we got a bug that the name was not displayed or not updated in some circumstances, because the wrong field was used.

But this is no longer important. As I worked on this code, I realized I can simply remove `displayName` altogether. In current code it just carries the copy of `name`.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/db/ZMessagingDB.scala
```diff
@@ -390,6 +390,49 @@ object ZMessagingDB {
...
+        s"""
+           | UPDATE Users
+           | SET name = display_name
+           | WHERE display_name IS NOT NULL AND display_name IS NOT "" AND name IS NOT display_name
+        """.stripMargin)
+      db.execSQL(
+        """
+          | CREATE TABLE UsersCopy (
+          | _id TEXT PRIMARY KEY,
+          | teamId TEXT, name TEXT, email TEXT, phone TEXT, tracking_id TEXT,
```
gizemb: i guess you dropped the `displayName` and unified them under `name` property, right?

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/content/UsersStorage.scala
```diff
@@ -117,27 +117,27 @@ class UsersStorageImpl(context: Context, storage: ZmsDatabase)
...
     }
-  }, "UsersDisplayNameUpdater")
+  }, "UsersNameUpdater")
 
-  def updateDisplayNamesWithSameFirst(users: Seq[UserId], cs: mutable.HashMap[UserId, NameParts]): Unit = {
-    def setFullName(user: UserId) = update(user, { (u : UserData) => u.copy(displayName = u.name) })
-    def setDisplayName(user: UserId, name: String) = update(user, (_: UserData).copy(displayName = Name(name)))
+  def updateNamesWithSameFirst(users: Seq[UserId], cs: mutable.HashMap[UserId, NameParts]): Unit = {
+    def setFullName(user: UserId) = update(user, { (u : UserData) => u.copy(name = u.name) })
+    def setDisplayName(user: UserId, name: String) = update(user, (_: UserData).copy(name = Name(name)))
```
gizemb: what's the difference between setFullName and setDisplayName now?

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/log/LogShowInstancesSE.scala
```diff
@@ -146,7 +146,7 @@ trait LogShowInstancesSE {
   implicit val UserDataLogShow: LogShow[UserData] =
     LogShow.createFrom { u =>
       import u._
-      l"UserData(id: $id | teamId: $teamId | name: $name | displayName: $displayName | email: $email | phone: $phone | handle: $handle | deleted: $deleted)"
+      l"UserData(id: $id | teamId: $teamId | name: $name | displayName: $name | email: $email | phone: $phone | handle: $handle | deleted: $deleted)"
```
gizemb: we can omit displayName now

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/db/ZMessagingDB.scala
gizemb: i saw your comment now. marking this as resolved.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/log/LogShowInstancesSE.scala
makingthematrix: done

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/content/UsersStorage.scala
makingthematrix: code removed

Issue #790, opened 2017-04-13
Long pressing a message with a link in it opens the link


someoneEsle:
## The problem
- Send a message to A, anything that has a url in it
- Long press the message (say that you want to forward it, or copy it, or delete it, whatever)
- Browser opens instead of options menu

## Environment
* Wire version that exhibits the issue: 2.31.324
* Last Wire version that did not exhibit the issue (if applicable): NA
* Android OS version used to run Wire: 7.1.2
* Is this a custom firmware or a stock one: Stock
* Mobile phone model/manufacturer: Nexus 5X
* Mobile network type (EDGE/LTE/Wi-Fi/Offline): --

red-avtovo:
Thank you for you report, @someoneEsle 
Unfortunately we know that issue, but we haven't fixed that yet

PR #789, opened 2017-04-12
Group d'etat: Group calling v3!

hakonbo:
V3 group call
#### APK
[Download build #8749](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8749/artifact/build/artifact/wire-dev-PR789-8749.apk)
[Download build #8759](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8759/artifact/build/artifact/wire-dev-PR789-8759.apk)
[Download build #8782](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8782/artifact/build/artifact/wire-dev-PR789-8782.apk)
[Download build #8784](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8784/artifact/build/artifact/wire-dev-PR789-8784.apk)
[Download build #8789](http://192.168.10.18:8080/job/Pull%20Request%20Builder/8789/artifact/build/artifact/wire-dev-PR789-8789.apk)

PR #2527, opened 2020-01-07
Android10: Incoming call notifications fix 

BradleyWire:
### Issues

On Android10 we need consistent behaviour when the application isn't in the foreground instead of showing an overlay and notification. 

### Causes

New restrictions on Android10 means we can't automatically show an overlay when an incoming call comes in, so we have to show a notification to bring the app to the foreground. 

### Solutions

Show just a notification for when the app is in background and an overlay when the app is active for Android10 + devices. 

#### APK
[Download build #789](http://10.10.124.11:8080/job/Pull%20Request%20Builder/789/artifact/build/artifact/wire-dev-PR2527-789.apk)
[Download build #790](http://10.10.124.11:8080/job/Pull%20Request%20Builder/790/artifact/build/artifact/wire-dev-PR2527-790.apk)

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -682,7 +682,7 @@
 
     <!-- System notifications -->
     <string name="system_notification__silence_call">Decline</string>
-    <string name="system_notification__join_call">Accept</string>
+    <string name="system_notification__join_call">Open Call</string>
     <string name="system_notification__leave_call">End call</string>
     <string name="system_notification__calling_one">Calling…</string>
     <string name="system_notification__calling_group">%1$s is calling…</string>
--- a/app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
@@ -23,7 +23,7 @@ import com.waz.api.Verification
 import com.waz.avs.VideoPreview
 import com.waz.content.GlobalPreferences
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
-import com.waz.model.{Picture, _}
+import com.waz.model._
 import com.waz.service.ZMessaging.clock
 import com.waz.service.call.Avs.VideoState
 import com.waz.service.call.CallInfo.CallState.{SelfJoining, _}
--- a/app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
@@ -17,14 +17,13 @@
  */
 package com.waz.zclient.calling.controllers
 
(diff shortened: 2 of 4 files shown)


app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
```diff
@@ -17,14 +17,13 @@
...
-import android.os.PowerManager
+import android.os.{Build, PowerManager}
 import android.telephony.{PhoneStateListener, TelephonyManager}
 import com.waz.api.Verification
 import com.waz.avs.VideoPreview
 import com.waz.content.GlobalPreferences
-import com.waz.model.Picture
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
-import com.waz.model._
+import com.waz.model.{Picture, _}
```
gizemb: please either import whole package or specific classes

app/src/main/res/values/strings.xml
```diff
@@ -682,7 +682,7 @@
 
     <!-- System notifications -->
     <string name="system_notification__silence_call">Decline</string>
-    <string name="system_notification__join_call">Accept</string>
+    <string name="system_notification__join_call">Open Call</string>
```
gizemb: Does translations get notified when you change a string in English?
Also, it might be nice if it displayed "Accept" for overlays

BradleyWire: Aren't translations notified when we generate the build? The problem with accept is that it doesn't accept anything as there are issues with permissions and the way we implement them in the app. The solution for now until we look at refactoring the permissions is to open the overlay to give the user the option to accept or decline the call through this button.

gizemb: Open call seems fine then.
yes, they're notified. I'm sure it gets notified when you add or remove a string resource, but not sure about the value change

BradleyWire: Interesting, be good to know how this works.

app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
```diff
@@ -283,12 +282,13 @@ class CallController(implicit inj: Injector, cxt: WireContext, eventContext: Eve
...
   }
 
-  onCallStarted.on(Threading.Ui) { _ =>
+  onCallStarted.on(Threading.Ui) { activeUi =>
     (for {
       incoming <- isCallIncoming.head
       allowed  <- allowedByStatus.head
-    } yield !incoming || allowed).foreach {
-      case true  => CallingActivity.start(cxt)
+      shouldDisplayOverlay = activeUi || Build.VERSION.SDK_INT < 29
```
johnxnguyen: It might be helpful for the next developer to have a comment here stating that Android10+ devices will only notifications for calls in the background.

Issue #828, opened 2017-04-30
The call is still active after it is ended, when screen sharing


Asuanin:
I tested call betwen Adroid phone and Linux desktop, when screen sharing, and after end call, it is still active on Android and buttons and return has any reaction, help only kill process

Android: dial linux user
Linux: accept call and screen sharing
Android: not rescieve any video and beep sound more 10 second
Linux: call is ended
Android: call still active, time run and buttons and return has any reaction

Phone:
Lenovo P70A
Android 4.4.4 (stock rom)
Nova Launcher
Wire 2.31.325

PC:
Ubuntu 16.04.2 LTS
kernel 4.4.0-72 x86_64
LXDE, 4 virtual desktops
Wire 2.13.2721

mykola-mokhnach:
Please move the issue to https://github.com/wireapp/wire-desktop/issues, since it's a problem in web client, rather than Android one.

PR #2533, opened 2020-01-08
Kotlin network part 3: Simplify logic for updating refresh token

gizemb:
### Issues

The previous logic with RawResponseRegistry was too complicated for a single task.

### Solutions

Moved related code from ApiService to AccessTokenAuthenticator

### Testing

Unit tests provided.

#### APK
[Download build #828](http://10.10.124.11:8080/job/Pull%20Request%20Builder/828/artifact/build/artifact/wire-dev-PR2533-828.apk)


app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenInterceptor.kt
```diff
@@ -6,12 +6,18 @@ import okhttp3.Response
 
 class AccessTokenInterceptor(private val authTokenHandler: AuthTokenHandler) : Interceptor {
 
+    companion object {
```
android10: Since there is only one usage for this, maybe we can inline and remove the ```companion object``` and the ```constant```:
```kotlin
    private fun updateRefreshTokenFromResponse(response: Response) =
        response.headers()["Cookie"]?.let {
            if (authTokenHandler.refreshToken() != it) {
                authTokenHandler.updateRefreshToken(it)
            }
        }
```
Also if you feel that you need a name for the header you can also create a local variable inside the method.

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenInterceptor.kt
```diff
@@ -6,12 +6,18 @@ import okhttp3.Response
...
+    }
+
     override fun intercept(chain: Interceptor.Chain): Response {
         val token = authTokenHandler.accessToken()
 
         return when (token != String.empty()) {
             true -> addAuthHeader(chain, token)
             false -> chain.proceed(chain.request())
+        }.also {
+            updateRefreshTokenFromResponse(it)
```
android10: Naming, here response is implicit since you are passing it as a parameter. 
This is my suggestion: ```updateRefreshToken(it)```

app/src/main/kotlin/com/waz/zclient/core/network/ApiService.kt
```diff
@@ -28,10 +34,7 @@ class ApiService(private val networkHandler: NetworkHandler,
         return try {
             val response = call.execute()
             when (response.isSuccessful) {
-                true -> {
```
android10: Way cleaner! :heart:

app/src/main/kotlin/com/waz/zclient/core/network/AuthTokenHandler.kt
```diff
@@ -3,13 +3,10 @@ package com.waz.zclient.core.network
...
 import com.waz.zclient.core.functional.Either
 import com.waz.zclient.core.functional.map
-import retrofit2.Response
 
 class AuthTokenHandler(private val tokenRepository: AccessTokenRepository) {
 
     companion object {
-        private const val RESPONSE_HEADER_REFRESH_TOKEN_KEY = "Cookie"
-
         const val AUTH_HEADER = "Authorization"
```
android10: I'm wondering if we could move these 2 constants, since I do not if they belong to the ```TokenHandler```.
They are authentication header literals, so maybe move them to the ```AccessTokenAuthenticator``` class?

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenAuthenticator.kt
```diff
@@ -25,20 +30,26 @@ class AccessTokenAuthenticator(private val authTokenHandler: AuthTokenHandler) :
...
-        response.request().header(AuthTokenHandler.AUTH_HEADER)?.let {
+        response.request().header(AUTH_HEADER)?.let {
             response.request()
                 .newBuilder()
-                .removeHeader(AuthTokenHandler.AUTH_HEADER)
-                .addHeader(AuthTokenHandler.AUTH_HEADER,
-                    "${AuthTokenHandler.AUTH_HEADER_TOKEN_TYPE} $newAccessToken")
+                .removeHeader(AUTH_HEADER)
+                .addHeader(AUTH_HEADER,
+                    "$AUTH_HEADER_TOKEN_TYPE $newAccessToken")
```
android10: This can go in 1 line

app/src/main/kotlin/com/waz/zclient/core/network/AccessTokenAuthenticator.kt
```diff
@@ -25,20 +30,26 @@ class AccessTokenAuthenticator(private val authTokenHandler: AuthTokenHandler) :
...
                 .build()
         }
 
+    private fun updateRefreshToken(response: Response) =
+        response.headers()["Cookie"]?.let {
```
android10: I removed the constant since it was only being used in one place. If you feel you need a variable, extract to a local one inside the method.

gizemb: ok for now but detekt will probably fail here.

Issue #970, opened 2017-06-29
vibration recorded at start of audio messages


davidak:
## The problem

When i hold the record button to record an audio message, the phone vibrates once. This vibration is recorded on the start of every message.

[recording.m4a.zip](https://github.com/wireapp/wire-android/files/1113527/recording.m4a.zip)

## Environment

* Wire version that exhibits the issue: 2.36.343
* Last Wire version that did not exhibit the issue (if applicable): -
* Android OS version used to run Wire: Fairphone Open OS 17.04 (based on Android 5.1, without Google Play Serrvices)
* Is this a custom firmware or a stock one: alternative OS from manufacturer
* Mobile phone model/manufacturer: Fairphone 2
* Mobile network type (EDGE/LTE/Wi-Fi/Offline): LTE/Wifi

## Details

I have this issue with my Fairphone, but my friend with a Samsung Galaxy Note 4 don't has it. His don't vibrate at all when recording audio messages.

## Link to Debug logs

https://gist.github.com/davidak/eb86c0075ce720272bd345fa4a7d3b1a

jspittka:
Thanks for reporting this problem. We tried to make sure that we avoid recording the notifications but Android hardware differs quite a bit. It is possible that we missed this specific scenario.

davidak:
It don't vibrate in the latest version. That solves this problem.

PR #2536, opened 2020-01-09
Remove unused old assets and sound media code

makingthematrix:
This PR removes unused code:
* Old assets
* SoundCloud and Spotify views in UI

Notes:
1. I retained `AssetService.BitmapResult`. It shouldn't be necessary, but in the moment it's still use in `ImageNotificationsController`, solely as `Signal.empty[BitmapResult]`. This is to make sure that we won't forget to rewrite that functionality to new assets.
2. SoundCloud and Spotify can still be used in SE - another client may return a message content with one of them as the message type and we need to handle that. For all practical purposes, however, we treat them as weblinks.

Tests:
Our tests for asset sending and loading should be sufficient. No change in functionality was made (other than treating SoundClound and Spotify messages as regular weblinks).
#### APK
[Download build #868](http://10.10.124.11:8080/job/Pull%20Request%20Builder/868/artifact/build/artifact/wire-dev-PR2536-868.apk)
[Download build #970](http://10.10.124.11:8080/job/Pull%20Request%20Builder/970/artifact/build/artifact/wire-dev-PR2536-970.apk)


app/src/main/scala/com/waz/zclient/notifications/controllers/ImageNotificationsController.scala
```diff
@@ -65,8 +65,9 @@ class ImageNotificationsController(implicit cxt: WireContext, eventContext: Even
       }
     case _ => Signal.empty[BitmapResult]
   }.zip(savedImageUri).on(Threading.IO) {
-    case (BitmapResult.BitmapLoaded(bitmap, _), uri) => showBitmap(bitmap, uri)
-    case (_, uri) => showBitmap(null, uri)
+    //case (BitmapResult.BitmapLoaded(bitmap, _), uri) => showBitmap(bitmap, uri)
```
BradleyWire: Should we not just remove this?

makingthematrix: I want to investigate this piece a bit more before removing. As I understand, we should display the avatars in notifications and we don't do it now... or maybe we can make an explicit decision not to do it.

app/src/main/scala/com/waz/zclient/notifications/controllers/CallingNotificationsController.scala
```diff
@@ -76,11 +73,11 @@ class CallingNotificationsController(implicit cxt: WireContext, eventContext: Ev
     for {
       zs                     <- inject[AccountsService].zmsInstances
       (curCallId, allCallsF) <- filteredGlobalProfile
-      bitmaps                <- Signal.sequence(allCallsF.map { case (conv, (caller, account)) =>
+      bitmaps                <- Signal.sequence(allCallsF.map { case (conv, (_, account)) =>
                                   zs.find(_.selfUserId == account)
                                     .fold2(
                                       Signal.const(conv -> Option.empty[Bitmap]),
-                                      z => getBitmapSignal(z, caller).map(conv -> _)
+                                      z => Signal.const(conv -> Option.empty[Bitmap]) // //TODO: Use new assets engine to fetch the bitmap
```
mejdoo: Any plans to implement this TODO for the new assets engine ?

mejdoo:
I reviewed around 120 files... I know its complicated, but we need to find a solution how to split the refactoring in such cases into multiple PRs...

app/src/main/scala/com/waz/zclient/notifications/controllers/CallingNotificationsController.scala
makingthematrix: It's the same issue as above.
This functionality wasn't working before. My change only made it more clear that it's not working :) Of course I agree we should either fix it or remove the code, but I'd suggest to do it in another PR.

makingthematrix:
build again

makingthematrix:
build again

makingthematrix:
> I reviewed around 120 files... I know its complicated, but we need to find a solution how to split the refactoring in such cases into multiple PRs...

I guess this is a very special case. The deleted lines are just old unused code. The added lines are in 99% changes to names and locations.

PR #1082, opened 2017-08-17
[AN-5502] vibration on notifications works according to user preferences

makingthematrix:
#### APK
[Download build #9341](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9341/artifact/build/artifact/wire-dev-PR1082-9341.apk)
[Download build #9346](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9346/artifact/build/artifact/wire-dev-PR1082-9346.apk)
[Download build #9347](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9347/artifact/build/artifact/wire-dev-PR1082-9347.apk)
[Download build #9355](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9355/artifact/build/artifact/wire-dev-PR1082-9355.apk)
[Download build #9356](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9356/artifact/build/artifact/wire-dev-PR1082-9356.apk)
[Download build #9357](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9357/artifact/build/artifact/wire-dev-PR1082-9357.apk)
[Download build #9358](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9358/artifact/build/artifact/wire-dev-PR1082-9358.apk)
[Download build #9359](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9359/artifact/build/artifact/wire-dev-PR1082-9359.apk)

PR #2575, opened 2020-01-29
Remove ability for app to read phone state and some redundant code

BradleyWire:
### Issues

When migrating the edit phone functionality to Kotlin, I discovered some redundant code in the current codebase. When a phone number exists for the user already: we're checking the permission for READ_PHONE_STATE (we never request it in the app in real time for API 23 >, so it'll never be granted for 6.0 or above) and if the user has granted the permission at install time (for 21/22) it'll then retrieve the sim/network data on the device. The code itself doesn't add much value to the functionality as it behaves as normal from a UX standpoint without it.

### Solutions

Removed the permission from the manifest and the controller in which you can read the telephony service data to retrieve the country iso and phone number from the sim card and network. 

### Testing

1.  Install app on API 21 | 22 
2. Register with a personal account using email
3. Go to settings > Account > Add phone number and add a phone number 
4. Verify phone number via sms code 
5. Leave dialog
6. Come back phone should appear as the one you verified, not the one attached to your sim. 

#### APK
[Download build #1082](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1082/artifact/build/artifact/wire-dev-PR2575-1082.apk)


app/src/main/AndroidManifest.xml
```diff
@@ -42,8 +42,6 @@
     <!-- Needed for taking photos when setting profile image or sending images -->
     <uses-permission android:name="android.permission.CAMERA" />
     <uses-permission android:name="android.permission.FLASHLIGHT" />
-    <!-- get phone number for contact matching -->
```
BradleyWire: This comment was written in 2016 and doesn't state the correct function for this permission.

makingthematrix:
Great job! :)

PR #2576, opened 2020-01-29
AN-6612: Fixed the issue for Android 10 < devices for when we're declaring imp…

BradleyWire:
…ortant notifications

### Issues

1. Notification is displaying over the overlay for Android10 devices when we open the call from the action item on the notification
2. Notification is displaying as-well as the overlay for Android 10 or lower devices. 

### Causes

Set full screen intent is too high priority for Android10 or below devices to the point it sticks around when we don't want it to also with the foreground service was still running when we open the overlay 

### Solutions

De-prioritised setFullIntent for Android 10 or below, changed the showNotification logic + pending intent flags for the calling activity. 

#### Scenario 1 
1. Call a device with Android 10 or less inside the app 
2. Overlay should display with no calling push notifications

#### Scenario 2
1. Call a device with Android 10 or less with the app not in the background/foreground
2. Overlay should display with no calling push notifications 

#### Scenario 3
1. Call a device with Android 10 or less with the app is in the background
2. Overlay should display with no calling push notifications 

#### Scenario 4
1. Call a device with Android 10 or higher with the app is in the background
2. Push notification should display
3. Click open call
4. Call overlay should display and push notification should disappear

### DIFF
--- a/app/src/main/scala/com/waz/zclient/notifications/controllers/CallingNotificationsController.scala
+++ b/app/src/main/scala/com/waz/zclient/notifications/controllers/CallingNotificationsController.scala
@@ -228,14 +228,15 @@ object CallingNotificationsController {
       .setCategory(NotificationCompat.CATEGORY_CALL)
       .setPriority(priority)
       .setOnlyAlertOnce(true)
-      .setOngoing(true)
-
-
-    if (isAndroid10OrAbove) builder.setFullScreenIntent(OpenCallingScreen(), true)
+      .setAutoCancel(true)
 
     if (!not.isMainCall) {
       builder.setDefaults(NotificationCompat.DEFAULT_LIGHTS | NotificationCompat.DEFAULT_VIBRATE)
       builder.setSound(RingtoneUtils.getUriForRawId(cxt, R.raw.empty_sound))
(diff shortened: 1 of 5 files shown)


app/src/main/scala/com/waz/zclient/notifications/controllers/CallingNotificationsController.scala
```diff
@@ -225,15 +225,18 @@ object CallingNotificationsController {
...
       .setPriority(priority)
       .setOnlyAlertOnce(true)
-      .setOngoing(true)
+      .setAutoCancel(true)
 
     if (!not.isMainCall) {
       builder.setDefaults(NotificationCompat.DEFAULT_LIGHTS | NotificationCompat.DEFAULT_VIBRATE)
       builder.setSound(RingtoneUtils.getUriForRawId(cxt, R.raw.empty_sound))
+    } else {
+      if (isAndroid10OrAbove) {
```
gizemb: can it be `else if`?

BradleyWire: done

Issue #1134, opened 2017-09-20
Disable downloads when on mobile or even roaming


kroerig:
At the moment you can only limit the download of images to Wi-Fi.

Would be great to choose this for video, audio and documents as well.

someoneEsle:
The download of videos, audios and documents only starts when the user intentionally taps on them. :)

PR #2585, opened 2020-01-31
Edit Phone screen (Part 3) 

BradleyWire:
## What's new in this PR?

There is a [big hack in Part 2](https://github.com/wireapp/wire-android/pull/2580#discussion_r372977437) around starting dialogs from another dialog. So to overcome that we've removed the first dialog for editing the phone number and refactored that into a screen instead so we can launch the other two dialogs with ease.

#### APK
[Download build #1134](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1134/artifact/build/artifact/wire-dev-PR2585-1134.apk)
[Download build #1146](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1146/artifact/build/artifact/wire-dev-PR2585-1146.apk)


app/src/main/kotlin/com/waz/zclient/core/extension/Fragment.kt
```diff
@@ -20,3 +21,9 @@ fun Fragment.openUrl(url: String) =
 
 fun Fragment.startActivityWithAction(intentAction: String) =
     startActivity(Intent().apply { action = intentAction })
+
+fun Fragment.replaceFragment(frameId: Int, fragment: Fragment, addToBackStack: Boolean) =
+    (activity as AppCompatActivity).replaceFragment(frameId, fragment, addToBackStack)
+
+fun Fragment.showBackArrow() =
```
BradleyWire: Ideally we'd convert the toolbar into a custom composite tool that we can inject into the activity which will have this functionality anyway.

app/src/main/kotlin/com/waz/zclient/settings/main/SettingsMainFragment.kt
```diff
@@ -55,7 +54,7 @@ class SettingsMainFragment : Fragment(), OnItemClickListener {
     }
 
     private fun replaceFragment(fragment: Fragment) {
-        (activity as AppCompatActivity).replaceFragment(R.id.layout_container, fragment, true)
+        replaceFragment(R.id.layout_container, fragment, true)
```
mejdoo: Change `layout_container`  to align it with our coding style.

BradleyWire:
Tests will have full coverage in this PR: https://github.com/wireapp/wire-android/pull/2587

app/src/main/kotlin/com/waz/zclient/core/extension/Fragment.kt
android10: Maybe a name for this would be something like ```enableHomeNavigation()```

BradleyWire: i'll remove this for now and then we can work on a reusable toolbar component as a team in the next iteration?

Issue #1177, opened 2017-10-10
Archiving a Conversation does not work via Swipe menu in the conversations list.


tokariu:
## The problem

conversations can not be archived via the swipe menu in the conversations list.

## Environment

* Wire version that exhibits the issue: 2.41.359
* Last Wire version that did not exhibit the issue (if applicable): n\a
* Android OS version used to run Wire: 6.0
* Is this a custom firmware or a stock one: stock
* Mobile phone model/manufacturer: LG G3
* Mobile network type (EDGE/LTE/Wi-Fi/Offline): doesn't matter.

## Details

When you're in your conversations list, you swipe one conversation to the right and a menu shows up with the following options: archive, block, delete, mute, send pic, call

if you select archive from there, nothing happens. the selected conversation does not get archived and is still in the conversation list.

if you want it to work, you have to archive it by clicking on the conversation, open it, then click on the conversation/nickname at the top, then click on the "..." at the bottom right, (a menu opens) and then select "archive" from there. this will finally archive the conversion.

this is cumbersome and the swipe methode should work again.

makingthematrix:
Hi.
I'm going through old issues in order to figure out which ones can be closed.
I can't reproduce this bug. The functionality was changed a lot since 2017, so I guess it was fixed at some point. Please reopen this issue if you find it again.

PR #2587, opened 2020-02-03
Edit phone screen (Part 4) - Refactoring and test cases 

BradleyWire:
### Issues

In this part i've refactored to downsize the UserDataSource, update the test cases for phone number  changes and also added the confirmation behaviour to navigate to the new "verify phone" screen. 
#### APK
[Download build #1177](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1177/artifact/build/artifact/wire-dev-PR2587-1177.apk)


app/src/main/AndroidManifest.xml
```diff
@@ -227,7 +227,14 @@
             android:name=".conversation.folders.moveto.MoveToFolderActivity"
             android:screenOrientation="portrait" />
         <activity android:name=".settings.devices.detail.SettingsDeviceDetailActivity" />
-      
+
+        <activity
+            android:name="com.waz.zclient.settings.account.editphonenumber.EditPhoneNumberActivity"
+            android:hardwareAccelerated="true"
```
mejdoo: I'm just wondering why we need to specify explicitly `android:hardwareAccelerated="true"` here? You you have some heathy tasks ?

app/src/main/kotlin/com/waz/zclient/settings/account/editphonenumber/SettingsAccountPhoneNumberViewModel.kt
```diff
@@ -17,31 +18,30 @@ import com.waz.zclient.user.domain.usecase.phonenumber.PhoneNumberInvalid
...
-
 class SettingsAccountPhoneNumberViewModel(
     private val validatePhoneNumberUseCase: ValidatePhoneNumberUseCase,
     private val changePhoneNumberNumberUseCase: ChangePhoneNumberUseCase,
     private val countryCodeAndPhoneNumberUseCase: CountryCodeAndPhoneNumberUseCase,
     private val deletePhoneNumberUseCase: DeletePhoneNumberUseCase
 ) : ViewModel() {
 
-    private var _countryCodeErrorLiveData = MutableLiveData<CountryCodeErrorMessage>()
+    private var _countryCodeErrorLiveData = MutableLiveData<PhoneNumberErrorMessage>()
```
mejdoo: Is there any way to reduce the number of LiveData objects? It seems like we have a huge list in this viewModel.

app/src/main/kotlin/com/waz/zclient/user/data/phone/PhoneNumberDataSource.kt
```diff
@@ -0,0 +1,29 @@
...
+import com.waz.zclient.user.data.source.local.UsersLocalDataSource
+import com.waz.zclient.user.data.source.remote.UsersRemoteDataSource
+import kotlinx.coroutines.runBlocking
+
+class PhoneNumberDataSource(
+    private val usersRemoteDataSource: UsersRemoteDataSource,
+    private val usersLocalDataSource: UsersLocalDataSource
+) : PhoneNumberRepository {
+
+    override suspend fun changePhone(phone: String) = changePhoneRemotely(phone)
```
mejdoo: I guess you shouldn't use `usersRemoteDataSource` `usersLocalDataSource` here. Either you create `PhoneNumberRemoteDataSource` and `PhoneNumberLocalDataSource` or you keep all those methods in `UsersDataSource`

BradleyWire: The way I see it is: the `UserLocalDataSource` hit the Users table in the database and `UserRemoteDataSource` hits the Self/User API on the backend, which is what data sources are responsible for: requests. 

This repository just takes the responsibilities for the Phone away from the UserRepository as that repository is getting far too big. The repository doesn't need to be tied to phone data sources I don't believe.

app/src/main/AndroidManifest.xml
BradleyWire: Just copied it over from another activity. It isn't needed.

app/src/main/kotlin/com/waz/zclient/settings/account/editphonenumber/SettingsAccountPhoneNumberViewModel.kt
BradleyWire: The LiveData objects to me represent a different responsibility on the UI the ViewModel can trigger. These all have their own responsibilities based on logic in the ViewModel, I can't make the list smaller without passing some of the logic onto the view.

app/src/main/kotlin/com/waz/zclient/user/data/phone/PhoneNumberDataSource.kt
mejdoo: Using the same approach can you move responsibilities to some other repositories and get rid of UsersRepository? Why you decided to that only for the PhoneNumber?

BradleyWire: Because that's the scope of this PR

BradleyWire: Added a TODO (for us in the future) to fix this haha

PR #1229, opened 2017-10-30
Testing Gallery: broadcast calls result codes

red-avtovo:
provided proper intent calls result codes
#### APK
[Download build #9725](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9725/artifact/build/artifact/wire-dev-PR1229-9725.apk)

PR #2602, opened 2020-02-10
bugfix (M2 onboarding flow) : Flow should be restarted when I click on SSO link again 

mejdoo:
### Issues

There is a bug reported by QA : Flow should be restarted when I click on SSO link again https://wearezeta.atlassian.net/browse/AN-6645

### Solutions

Force sso webview to restart the flow when the app goes to the background

### Testing

- Open SSO deeplink. Example `wire://access/?config=https://aves.services.zinfra.io`
- Enter credentials on SSO page : email and password
- Minimise the application (Go to the background)
- Open same SSO deeplink
- Email and password fields in the webView should be empty

#### APK
[Download build #1229](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1229/artifact/build/artifact/wire-dev-PR2602-1229.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/appentry/SSOWebViewFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/appentry/SSOWebViewFragment.scala
@@ -100,7 +100,7 @@ class SSOWebViewFragment extends FragmentHelper {
     true
   }
 
-  override def onPause(): Unit ={
+  override def onPause(): Unit = {
     super.onPause()
     onBackPressed()
   }
--- a/app/src/main/scala/com/waz/zclient/appentry/SSOWebViewFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/appentry/SSOWebViewFragment.scala
@@ -17,14 +17,14 @@
  */
 package com.waz.zclient.appentry
 
-import android.app.FragmentManager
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/appentry/SSOWebViewFragment.scala
```diff
@@ -100,6 +100,11 @@ class SSOWebViewFragment extends FragmentHelper {
     true
   }
 
+  override def onPause(): Unit = {
+    super.onPause()
+    onBackPressed()
```
BradleyWire: If it's keeping state we're worrying about, would we not be better clearing the cache on the WebView when we leave the screen?

mejdoo: I discussed about that with Marco during the meeting and he suggested to do that instead of clearing the cache.

mejdoo: Changed the solution to clear cache after our discussion.

BradleyWire: nice

PR #1246, opened 2017-11-06
Reorganize the packages of the scala code

JPonte:
#### APK
[Download build #9745](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9745/artifact/build/artifact/wire-dev-PR1246-9745.apk)
[Download build #9750](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9750/artifact/build/artifact/wire-dev-PR1246-9750.apk)
[Download build #9905](http://192.168.10.18:8080/job/Pull%20Request%20Builder/9905/artifact/build/artifact/wire-dev-PR1246-9905.apk)

Issue #1245, opened 2017-11-04
When logging on another android phone the previous account is just not connecting anymore


Transisto:
There is not ever an error message coming up, just the top orange line loading endlessly on the original device.

Very disapointing behavior I was expecting to be able to use it like Telegram on two device.

someoneEsle:
Did you by any chance reset your password?

someoneEsle:
We don't have enough information to debug, so I'll close this for now. It should be fixed in the upcoming release anyway.

PR #2608, opened 2020-02-11
Refactor: Inflation logic + Settings back stack fix

BradleyWire:
## What's new in this PR?

Fix some issues with the back stack when you press back on some of the newer screens and refactored some of the inflation logic to use fragment constructors rather than having a single `onCreateView` method. 
#### APK
[Download build #1245](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1245/artifact/build/artifact/wire-dev-PR2608-1245.apk)
[Download build #1246](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1246/artifact/build/artifact/wire-dev-PR2608-1246.apk)


app/src/main/kotlin/com/waz/zclient/settings/account/editphonenumber/CountryCodePickerFragment.kt
```diff
@@ -32,6 +32,11 @@ class CountryCodePickerFragment : DialogFragment() {
 
     private var countryCodeListener: CountryCodePickerListener? = null
 
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setStyle(STYLE_NO_FRAME, R.style.Theme_Dark_Preferences)
```
gizemb: why do we force dark theme?🤔

mejdoo: I guess this is the actual implementation as well in the scala code

BradleyWire: This is how it's implemented in the current screen for the edit phone journey in Scala for other dialogs/screens. I'm sure we can look at changing it to be dynamic though.

gizemb: ok then

Issue #1260, opened 2017-11-09
Application stuck after password change


gapop:
## The problem

I have changed my password from the Web app. Now, the app cannot fetch new data from server, but it doesn't prompt for new password (doesn't log out).

When I open Wire on my Android tablet (Nexus 7, 2012), it opens to the last accessed conversation and shows an animated blue line at the top, as though it's trying to connect. It does not fetch new data, and there is no way to put in the new password. When I tap on my avatar to go to Account settings, the app jumps back to the conversation where it opened.

I guess that deleting the app data would force me to sign in again and would solve the issue, but I'm afraid that it would also lose all the device verification status information.

## Environment

* Wire version that exhibits the issue: 3.2.373
* Android OS version used to run Wire: 5.1.1
* Is this a custom firmware or a stock one: stock
* Mobile phone model/manufacturer: Asus Nexus 7 tablet, 2012
* Mobile network type (EDGE/LTE/Wi-Fi/Offline): Wi-Fi

## Details

Steps as it happened for me:

- Sign in to tablet, open conversation
- Close app, lock tablet
- Change password from different device (was Desktop app in my case)
- Open Wire on tablet

## Link to Debug logs

The bug prevents access to the app settings, cannot provide logs.

## How to send detailed debug report from Wire:

Cannot access app menu because of bug.

someoneEsle:
Hi, this is a known bug that we are currently addressing. Thanks for reporting!

someoneEsle:
Hey @gapop , are you a beta tester? If so, could you update to the latest version and let us know if it fixes your issue? Thanks!

gapop:
I enrolled as beta tester and updated to 3.3.374. At first, I saw the exact same behavior as before. Then I cleared the app's cache (not the "app data") and I got the login screen. After login, all is normal. Device verification information is unaffected.

Before reporting the bug, I did not try to clear the cache, not sure if that would've helped with the old version.

Thanks for the fix!

someoneEsle:
Interesting, thanks for reporting your experience!

PR #2609, opened 2020-02-11
Refactor assets to enable image rotation and avoid unnecessary creation of files

makingthematrix:
This commit is quite big but I felt there was no good way to break it into smaller ones.

With our new assets implementation, we try to treat assets of all types and coming from all sources in the same way as much as possible. In part it was achieved by turning them to input streams - no matter if originally they're files or byte arrays or any other kind of data. But in certain situations we need to know what we're working on. One such situation is when we send an image: we need to scale it down. Without knowing the path to the original file we were not able to get the EXIF information about the file (not sure why, ask Java) and sometimes the rotation of the scaled-down image came out wrong.

In order to fix it, I would have to carry the EXIF information through many layers of the asset functionality, only to use it if the asset was an image and needed to be scaled down at one point of the whole procedure. That wouldn't look nice.

On top of that, our original idea of treating everything as input streams meant that sometimes we converted files to input streams and then back to files, copying data, even when using the original file directly would be sufficient, risking performance issues.

So, I decided to refactor assets. Instead of `InputStream` we have now `AssetInput` - a trait which underneath hold the asset as either a real input stream, or a file (or a failure if something went wrong and we just carry information that sending/downloading failed). When we perform an operation on an "asset input" it may either perform a conversion between an input stream and a file, or just access them directly, if the conversion is not really necessary. In case of images from the gallery, which originally contain the EXIF information, the underlying instance is a file, so we can scale it down ad rotate it properly.

I will add more explanation of the whole sending/downloading procedure as comments in `AssetService`.

#### APK
[Download build #1248](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1248/artifact/build/artifact/wire-dev-PR2609-1248.apk)
[Download build #1250](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1250/artifact/build/artifact/wire-dev-PR2609-1250.apk)
[Download build #1260](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1260/artifact/build/artifact/wire-dev-PR2609-1260.apk)


wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/assets/AssetInput.scala
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+  override def toInputStream: Try[InputStream] = Try { new BufferedInputStream(new FileInputStream(file)) }
+
+  override def toBitmap(opts: BitmapFactory.Options): Try[Bitmap] =
+    Try {
+      val path            = file.getAbsolutePath
+      val bitmap          = BitmapFactory.decodeFile(path, opts)
+      val currentRotation = AssetInput.currentRotation(path)
+      if (currentRotation != 0) AssetInput.rotate(bitmap, currentRotation) else bitmap
```
gizemb: can we use `rotateIfNeeded`?

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/assets/AssetInput.scala
```diff
@@ -0,0 +1,156 @@
...
+      case Failure(throwable) =>
+        AssetFailure(throwable)
+    }
+
+  override def toInputStream: Try[InputStream] = Try { new BufferedInputStream(new FileInputStream(file)) }
+
+  override def toBitmap(opts: BitmapFactory.Options): Try[Bitmap] =
+    Try {
+      val path            = file.getAbsolutePath
+      val bitmap          = BitmapFactory.decodeFile(path, opts)
```
gizemb: isn't it prone to OutOfMemory errors? do we always check the image size before calling this method?

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/media/GoogleMapsMediaService.scala
```diff
@@ -45,6 +46,9 @@ class GoogleMapsMediaServiceImpl(googleMapsClient: GoogleMapsClient)(implicit ex
       case Right(result) => CancellableFuture.successful(result)
     }
   }
+
+  def loadMapPreview2(location: Location, dimensions: Dim2 = ImageDimensions): CancellableFuture[AssetInput] =
```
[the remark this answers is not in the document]
makingthematrix: ahh, sorry. that's an artifact from refactoring. removed.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/utils/crypto/AESUtils.scala
```diff
@@ -49,6 +50,9 @@ object AESUtils {
   def decrypt(key: AESKey, input: Array[Byte]): Array[Byte] =
     cipher(key, input.take(16), Cipher.DECRYPT_MODE).doFinal(input.drop(16))
 
+  def decrypt(key: AESKey2, input: Array[Byte]): Array[Byte] =
```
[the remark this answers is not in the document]
makingthematrix: This is something different from the original `AESKey`. I changed the name to `AESKeyBytes`.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/assets/AssetInput.scala
makingthematrix: yes

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/handler/MessagesSyncHandler.scala
```diff
@@ -253,6 +253,7 @@ class MessagesSyncHandler(selfUserId: UserId,
       }
 
     def sendWithV3(uploadAssetOriginal: UploadAsset): CancellableFuture[RemoteInstant] = {
+      verbose(l"sendWithV3($uploadAssetOriginal)")
```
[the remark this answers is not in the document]
makingthematrix: ok

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/assets/AssetInput.scala
makingthematrix: If there's an OOM error, we will get a failure from the `Try` clause instead of an exception. Then it will be propagated as `AssetFailure` and logged at some point.

gizemb: hmm yes, but rather than failing, it's better to resize the image to not to get an OOM exception. Anyway, I checked the code and seems like the size is always checked before calling this method, but adding a comment about this to this method would be nice imo.

makingthematrix:
build again

PR #2612, opened 2020-02-12
User Database migration bump

BradleyWire:
## What's new in this PR?

Bump start and end version for migration class


BradleyWire:
As discussed with @gizemb, there are too many unknowns to be confident about this being in production

PR #2614, opened 2020-02-13
Fix: Update room db version and fix migration versions

gizemb:
### Issues

The app was getting stuck when KOTLIN_SETTINGS_MIGRATION was enabled.

### Causes

The Room db and Scala side SQL db had the same versions (126). Moreover, both of them had a migration from version 125 to 126. In such a case where there are more than two migrations with the same version numbers, whichever added last overrides the others.

### Solutions

Bumped Room db version to SQLite db version + 1, along with the migration numbers.

### Testing

Update scenario tested:
Open app when flag is disabled and let some db transactions happen,
Install app again with flag enabled, and let migration happen.

Introduced unit tests for version checks.

## Notes

As a future work, we could come up with a way that doesn't require this manual update every time.

#### APK
[Download build #1257](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1257/artifact/build/artifact/wire-dev-PR2614-1257.apk)
[Download build #1267](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1267/artifact/build/artifact/wire-dev-PR2614-1267.apk)

### DIFF
--- a/storage/src/main/kotlin/com/waz/zclient/storage/db/UserDatabase.kt
+++ b/storage/src/main/kotlin/com/waz/zclient/storage/db/UserDatabase.kt
@@ -11,7 +11,7 @@ import com.waz.zclient.storage.db.users.service.UserPreferenceDao
 
 @Database(
     entities = [UserPreferenceEntity::class, UserEntity::class, ClientEntity::class],
-    version = 126,
+    version = 127,
     exportSchema = false
 )
 abstract class UserDatabase : RoomDatabase() {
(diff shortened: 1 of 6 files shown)


android10:
Should we somehow document this? I would say we need to find a better solution since every time we bump the database version, we need to remember to touch the other files, leading to errors and potential bugs. 

Also, is this part of planned story on JIRA?

storage/src/main/kotlin/com/waz/zclient/storage/db/UserDatabase.kt
```diff
@@ -11,7 +11,7 @@ import com.waz.zclient.storage.db.users.service.UserPreferenceDao
 
 @Database(
     entities = [UserPreferenceEntity::class, UserEntity::class, ClientEntity::class],
-    version = 126,
+    version = 127,
```
android10: Maybe write a unit test that asserts there is consistency between all these variables and versions? I would feel safer since the build breaks if they are not equivalent.

gizemb: ok I did that but I had some problems while doing so
1. I'd ideally would like to implement this test in storage module (where Room dbs are), but since it is independent of the zmessaging module (where Scala dbs are), I had to implement it in zmessaging, in order to access to legacy db versions.
2. I had to expose the Room databases' versions and migration list, to make them accessible from zmessaging.

Issue #1291, opened 2017-11-25
Wire can't handle big chats


juliss95:
The problem

When you open a big conversation, Wire takes 2-3 seconds to load the chat. Sending a message in a big chat takes 2-5 seconds (sometimes sending the message twice or more times). And same goes for when you're going into the media overview. I have the feeling that everything is 10x as slow as in small chats.
I think there's still a lot of room for optimization regarding the handling of larger chats!

Environment

* Wire version that exhibits the issue: 3.3.374 (and all of the versions before)
* Last Wire version that did not exhibit the issue (if applicable): -
* Android OS version used to run Wire: 5.0.2
* Is this a custom firmware or a stock one: MIUI 8
* Mobile phone model/manufacturer: Xiaomi Mi2S
* Mobile network type (EDGE/LTE/Wi-Fi/Offline): -

Details

The chat which has the loading times mentioned above has ~100.000 messages and ~2000 pictures. But generally, the larger a chat, the worse the loading times of it. 
Especially the times, that it needs to send messages can be a huge bummer...

CorvusCorax:
A workaround could be to erase the chat history. This might be useful in combination with the planned "export/backup" of chat history. I know some other messengers only keep the most recent messages (like a few hundred) around in the immediate history and the rest is read on demand from an (internal) archive if someone scrolls up or uses a search function. Wire might have to do something similar but make sure encryption is ensured.
Also archiving and self-destructing messages don't go well together. So whatever is archived should be older than the oldest self destructing message still hanging around.
Another issue is if someone (chat partner) deletes a message which is old and has already been archived or even backed up by their peers. Obviously theres limits to the extend to which deleting of already sent message can be guaranteed (after all people could always modify their client to dump a full clear text log to disk of anything they receive - self destructing or not)
But I think the solution to this problem is to write older messages into an internal archive instead of having them in memory as individual objects, and only open this archive when someone scrolls up or even presses a button (<load older messages>)

juliss95:
Yeah, those are changes that i was thinking about, cause in this state Wire is hardly usable as a primary messenger with close ones who you write a lot.
Because currently Wire is so overwhelmed in my biggest chat, that it takes sometimes really long with a stupid lag until a message is sent. And during that lag, the message often sends 2-5 times or stays in the message box even after sending. Or it even doesn't show up in the chat until you restart Wire.

So maybe a solution would be, to for example just load the last 200 messages and automatically load the previous 200 messages as you are scrolling up.
[...]

PR #2622, opened 2020-02-18
Scala/Room integration. Migrate SQLiteDatabase to SupportSQLiteDatabase

BradleyWire:
### Issues

When we decide to turn on the `KOTLIN_MIGRATION` flag and trigger the migration once we enter the settings screen. Once you put the application into the background - when you load the app up again you'll be met with a blank screen once we've entered the application

### Causes

This is due to the database migrations and initialisation relying on manual processes with a constant db version to restore data on login.

### Solutions

Make the old migrations compatible with the new Room migrations so we can bridge the gap somehow by migrating over SQLiteDatabase and the old SQLOpenHelper (DaoDB) to a new SupportSQLiteDatabase and SupportSQLOpenHelper which Room builds a foundation on to ensure there is some kind of compatibility between the two. 

## Notes

This isn't the complete issue resolved, with @gizemb's changes on top of this we'll be able to enable a dynamic migration of versions that are both compatible with the new room migrations and the old database migrations (which we still need the logic surrounding that in some use-cases) 

#### APK
[Download build #1290](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1290/artifact/build/artifact/wire-dev-PR2622-1290.apk)
[Download build #1291](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1291/artifact/build/artifact/wire-dev-PR2622-1291.apk)
[Download build #1314](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1314/artifact/build/artifact/wire-dev-PR2622-1314.apk)


wire-android-sync-engine/zmessaging/src/main/scala/com/waz/db/ZMessagingDB.scala
```diff
@@ -67,7 +66,7 @@ class ZMessagingDB(context: Context, dbName: String, tracking: TrackingService)
 }
 
 object ZMessagingDB {
-  val DbVersion = 126
+  val DbVersion = UserDatabase.VERSION
```
BradleyWire: This is temporary until we integrate @gizemb's fix into this.

BradleyWire:
Will update test cases.

app/src/androidTest/scala/com/waz/background/WorkManagerSyncRequestServiceSpec.scala
```diff
@@ -1,178 +0,0 @@
-/**
```
mejdoo: 👍  great ! But it would be better if done in a separate PR

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/db/package.scala
```diff
@@ -126,22 +126,19 @@ package db {
...
-        db.acquireReference()
-        try {
-          db.getThreadSession.beginTransaction()
-        }
-        finally db.releaseReference()
+      override def beginReadTransaction(db: DB): Unit = try reflectiveBegin(db) catch {
+        case _: Exception => db.beginTransactionNonExclusive()
       }
+
+      private def reflectiveBegin(db: DB): Unit = db.beginTransaction()
```
makingthematrix: You can inline this method, and then, in `beginReadTransaction` use `Try` instead of `try/catch`:
`Try(db.beginTransaction()).recover { case _ => db.beginTransactionNonExclusive() }`

mejdoo:
Good job. Hope this will fix the issue

app/src/androidTest/scala/com/waz/background/WorkManagerSyncRequestServiceSpec.scala
BradleyWire: I was debating that myself, but, there's no urgency for the tests to go into develop and it doesn't pollute this PR's scope too much as we are changing the tests.

BradleyWire:
> Good job. Hope this will fix the issue

It doesn't unfortunately but we are working on a fix as we've found the issue.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/model/package.scala
```diff
@@ -17,18 +17,14 @@
...
 import com.google.protobuf.nano.{CodedInputByteBufferNano, MessageNano}
 import com.waz.model.nano.Messages
 import com.waz.utils.crypto.AESUtils
 import com.waz.utils.{JsonDecoder, JsonEncoder, returning}
-import io.circe.{Decoder, Encoder}
 import org.json.JSONObject
 
-import scala.language.implicitConversions
 import scala.concurrent.duration.FiniteDuration
+import scala.language.implicitConversions
```
makingthematrix: I think we don't need implicitConversions anywhere here.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/db/package.scala
```diff
@@ -126,22 +126,19 @@ package db {
...
 
-      override def beginReadTransaction(db: DB): Unit = try reflectiveBegin(db) catch { case _: Exception => db.beginTransactionNonExclusive() }
-
-      private def reflectiveBegin(db: DB): Unit = {
-        db.acquireReference()
-        try {
-          db.getThreadSession.beginTransaction()
-        }
-        finally db.releaseReference()
+      override def beginReadTransaction(db: DB): Unit = try reflectiveBegin(db) catch {
```
gizemb: maybe we can get rid of reflectiveBegin method here. not that important though.

PR #2623, opened 2020-02-18
Scala/Room integration - Part 2 - Select Room/legacy database in runtime

gizemb:
## What's new in this PR?

**This solution requires the Room database to always be compatible with the Scala side.** 

### Issues

When we run the app with KOTLIN_SETTINGS_MIGRATION flag on, when we kill the app and reopen it, the app gets stuck in the launch screen.

### Causes

When the new Kotlin code runs and Room database is initialized, the database versions throughout the app increases (for user-specific database: 126 -> 127, for global database 24 -> 25). However, since Scala side uses the old databases (ZMessagingDB and ZGlobal) with old versions, the next time the app tries to make a database transaction with these legacy databases, an exception is thrown saying: "`android.database.sqlite.SQLiteException: Can't downgrade database from version 127 to 126`".

This exception is not thrown but also not propagated up to UI layer, since it is hidden inside a Signal call. In that case the app does not crash, but waits for a Signal that will never come because we got an exception while retrieving the signal, hence, the app gets stuck.

### Solutions

We now select the database in the runtime. If KOTLIN_SETTINGS_MIGRATION is on, the Scala side also operates on same SQLiteSupport database that Room wraps for us in Kotlin side. Otherwise, it continues to operate with the legacy database.

### Testing

Manual testing is done. A build with KOTLIN_SETTINGS_MIGRATION flag on should be given to QA to perform update and fresh installs with such a case.

WhatsApp Image 2020-02-18 at 18 46 13

#### APK
[Download build #1326](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1326/artifact/build/artifact/wire-dev-PR2623-1326.apk)
[Download build #1329](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1329/artifact/build/artifact/wire-dev-PR2623-1329.apk)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -130,7 +130,7 @@ android {
         buildConfigField 'boolean', 'WIPE_ON_COOKIE_INVALID',        "$buildtimeConfiguration.configuration.wipe_on_cookie_invalid"
         buildConfigField 'boolean', 'FORCE_PRIVATE_KEYBOARD',        "$buildtimeConfiguration.configuration.force_private_keyboard"
         buildConfigField 'Integer', 'PASSWORD_MAX_ATTEMPTS',         "$buildtimeConfiguration.configuration.password_max_attempts"
-        buildConfigField 'boolean', 'KOTLIN_SETTINGS_MIGRATION', "false"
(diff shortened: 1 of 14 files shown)


gizemb:
build again

PR #2625, opened 2020-02-19
Migrating Room entities from Scala part 1

BradleyWire:
### Issues

As apart of the migration from room to Scala integration we need to tell the room database that the tables in fact do exist so that the Scala code can read the database, but, to do that we need to create the entities that correspond to the Scala entities. 

### Causes

Room entities hadn't been migrated yet. 

### Solutions

Create all tables entities so that room can read it. 

### Testing

App should work as normal with the integration complete. 

#### APK
[Download build #1312](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1312/artifact/build/artifact/wire-dev-PR2625-1312.apk)
[Download build #1313](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1313/artifact/build/artifact/wire-dev-PR2625-1313.apk)
[Download build #1327](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1327/artifact/build/artifact/wire-dev-PR2625-1327.apk)


BradleyWire:
build again

PR #2640, opened 2020-02-25
refactor: Disable tracking on the TrackingService level

makingthematrix:
We don't track events right now, but we have a scaffold functionality left which can be used again if we decide to use another tracking system in the future. Right now, calling a track method on an event results only in a log message.
However, before that happens, in some cases `TrackingService` and `GlobalTrackingController` can perform quite a lot of operations, esp. `TrackingService.contribution` and `TrackingService.assetContribution`. I decided to move the code around a bit to prevent that. With this PR if the preference  `GlobalPreferences.AnalyticsEnabled` is set to `false` (which it is in production) we will skip performing these operations.

#### APK
[Download build #1372](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1372/artifact/build/artifact/wire-dev-PR2640-1372.apk)
[Download build #1491](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1491/artifact/build/artifact/wire-dev-PR2640-1491.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
+++ b/app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
@@ -68,10 +68,8 @@ class GlobalTrackingController(implicit inj: Injector, cxt: WireContext, eventCo
   tracking.events.foreach {
     case (zms, event: OpenedTeamRegistration) =>
       ZMessaging.currentAccounts.accountManagers.head.map {
-        _.size match {
-          case 0 => sendEvent(event, zms)
-          case _ => sendEvent(OpenedTeamRegistrationFromProfile(), zms)
-        }
+        case managers if managers.isEmpty => sendEvent(event, zms)
+        case _                            => sendEvent(OpenedTeamRegistrationFromProfile(), zms)
       }
     case (_, event: LoggedOutEvent) if event.reason == LoggedOutEvent.InvalidCredentials =>
       //This event type is trigged a lot, so disable for now
--- a/app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
+++ b/app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
@@ -21,17 +21,16 @@ package com.waz.zclient.tracking
 import android.content.Context
 import android.renderscript.RSRuntimeException
 import com.waz.api.impl.ErrorResponse
-import com.waz.content.Preferences.PrefKey
-import com.waz.content.{GlobalPreferences, UsersStorage}
+import com.waz.content.UsersStorage
 import com.waz.log.BasicLogging.LogTag
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
 import com.waz.model.ConversationData.ConversationType
 import com.waz.model.{UserId, _}
-import com.waz.service.{SearchQuery, ZMessaging}
+import com.waz.service.ZMessaging
 import com.waz.service.tracking.TrackingService.NoReporting
 import com.waz.service.tracking._
 import com.waz.threading.{SerialDispatchQueue, Threading}
-import com.waz.utils.events.{EventContext, Signal}
+import com.waz.utils.events.EventContext
 import com.waz.zclient._
 import com.waz.zclient.appentry.fragments.SignInFragment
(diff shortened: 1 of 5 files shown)


wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/tracking/TrackingService.scala
```diff
@@ -82,121 +88,156 @@ object TrackingService {
 
   trait NoReporting { self: Throwable => }
 
+  val analyticsPrefKey = BuildConfig.APPLICATION_ID match {
+    case "com.wire" | "com.wire.internal" => GlobalPreferences.AnalyticsEnabled
```
BradleyWire: Might be nicer checking the `BuildConfig.FLAVOR `here and checking for prod/internal

app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
```diff
@@ -73,47 +65,37 @@ class GlobalTrackingController(implicit inj: Injector, cxt: WireContext, eventCo
...
-            case true =>
-              throwable match {
-                case _: NoReporting =>
-                case _ => saveException(throwable, description)(e.tag)
-              }
-            case _ => //no action
+  tracking.events.foreach {
+    case (zms, event: OpenedTeamRegistration) =>
+      ZMessaging.currentAccounts.accountManagers.head.map {
+        _.size match {
```
BradleyWire: No way of checking `isEmpty` `true` or `false` here? would be more readable.

app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
```diff
@@ -73,47 +65,37 @@ class GlobalTrackingController(implicit inj: Injector, cxt: WireContext, eventCo
...
+    case (zms, event) => sendEvent(event, zms)
   }
 
   /**
     * @param eventArg the event to track
     * @param zmsArg a specific zms (account) to associate the event to. If none, we will try and use the current active one
+    *
+    * Right now it does nothing. The functionality stays here to be used when we implement a new tracking system
     */
   private def sendEvent(eventArg: TrackingEvent, zmsArg: Option[ZMessaging] = None) =
```
BradleyWire: If we're removing the code within this method, maybe it'd be good to clean up the params too? Unless we can guarantee any future solution will be dependent on `ZMessaging`?

makingthematrix: I think we can pretty much guarantee that. This is one `GlobalTrackingController`, collecting tracking events from all accounts. We need a reference to that account in order to get some additional data.

makingthematrix:
build again

app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
android10: I agree on the fact that if that is dead code doing nothing, should be removed. 
In case we decide to keep it, do we get any value out of logging the event?

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/tracking/TrackingService.scala
```diff
@@ -57,6 +60,8 @@ trait TrackingService {
   def historyRestored(isSuccess: Boolean): Future[Unit]
 
   def trackCallState(userId: UserId, callInfo: CallInfo): Future[Unit]
+
+  def isTrackingEnabled: Signal[Boolean]
 }
 
 class DummyTrackingService extends TrackingService {
```
android10: What is the purpose of this class? If it is for testing purpose then we have a design issue if we cannot mock it. 
Otherwise maybe rename to a more proper/understandable name?

app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
makingthematrix: The thing is, the whole tracking functionality is right now dead code. If I remove this methid then why not remove the method that calls it, and so on. 
We decided to keep it after removing Mixpanel becuase we thought that this way it will be easier to plug in a new tracking system. It's like an USB socket with no pendrive in it.

makingthematrix:
build again

PR #1378, opened 2018-01-15
[AN-5908] First time UX for team admins

JPonte:
#### APK
[Download build #10082](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10082/artifact/build/artifact/wire-dev-PR1378-10082.apk)
[Download build #10083](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10083/artifact/build/artifact/wire-dev-PR1378-10083.apk)
[Download build #10087](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10087/artifact/build/artifact/wire-dev-PR1378-10087.apk)
[Download build #10089](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10089/artifact/build/artifact/wire-dev-PR1378-10089.apk)
[Download build #10090](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10090/artifact/build/artifact/wire-dev-PR1378-10090.apk)
[Download build #10092](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10092/artifact/build/artifact/wire-dev-PR1378-10092.apk)
[Download build #10099](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10099/artifact/build/artifact/wire-dev-PR1378-10099.apk)
[Download build #10100](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10100/artifact/build/artifact/wire-dev-PR1378-10100.apk)
[Download build #10102](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10102/artifact/build/artifact/wire-dev-PR1378-10102.apk)
[Download build #10105](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10105/artifact/build/artifact/wire-dev-PR1378-10105.apk)
[Download build #10108](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10108/artifact/build/artifact/wire-dev-PR1378-10108.apk)
[Download build #10110](http://192.168.10.18:8080/job/Pull%20Request%20Builder/10110/artifact/build/artifact/wire-dev-PR1378-10110.apk)

PR #2641, opened 2020-02-25
perf: Optimize sharing many assets to many conversations

makingthematrix:
If the user goes to, for example, the gallery, and chooses a few photos and then share it to Wire to a few conversations at once, we handle that request in `SharingController`. We go through the collection of uris we got and for each uri we go through the collection of conversations, and for each pair we call a method which sends that one asset to that one conversation.
This PR inverts the procedure - first we go through conversations, and then through uris - and provides functionality for sending many assets at once to one conversation. Thanks to that, some operations, like setting the conversation's "last read time" or setting it to be the current conversation, are performed only once per sending all assets to that conversation, not each time for each asset. That makes the whole process much quicker, and also helps us avoid possible race conditions due to quick switching of the current conversation (still to be investigated).

Connected to: https://wearezeta.atlassian.net/browse/AN-5917
#### APK
[Download build #1378](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1378/artifact/build/artifact/wire-dev-PR2641-1378.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/common/controllers/SharingController.scala
+++ b/app/src/main/scala/com/waz/zclient/common/controllers/SharingController.scala
@@ -40,7 +40,6 @@ class SharingController(implicit injector: Injector, wContext: WireContext, even
   val sharableContent     = Signal(Option.empty[SharableContent])
   val targetConvs         = Signal(Seq.empty[ConvId])
   val ephemeralExpiration = Signal(Option.empty[FiniteDuration])
-  val conversationController = inject[ConversationController]
 
   def onContentShared(activity: Activity, convs: Seq[ConvId]): Unit = {
     targetConvs ! convs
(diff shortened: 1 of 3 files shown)


makingthematrix:
build again

app/src/main/scala/com/waz/zclient/common/controllers/SharingController.scala
```diff
@@ -50,14 +49,13 @@ class SharingController(implicit injector: Injector, wContext: WireContext, even
   def sendContent(activity: Activity): Future[Seq[ConvId]] = {
     def send(content: SharableContent, convs: Seq[ConvId], expiration: Option[FiniteDuration]) =
       content match {
-        case NewContent   =>
-          conversationController.switchConversation(convs.head)
+        case NewContent     =>
+          inject[ConversationController].switchConversation(convs.head)
         case TextContent(t) =>
-          conversationController.sendTextMessage(convs, t, Nil, None, Some(expiration))
+          inject[ConversationController].sendTextMessage(convs, t, Nil, None, Some(expiration))
```
BradleyWire: If we're injecting two instances wouldn't it be better to keep it as it was before as a global reference?

PR #1467, opened 2018-03-12
Replace old conversation options dialog with a bottom sheet

JPonte:
#### APK
[Download build #10524](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10524/artifact/build/artifact/wire-dev-PR1467-10524.apk)
[Download build #10529](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10529/artifact/build/artifact/wire-dev-PR1467-10529.apk)
[Download build #10533](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10533/artifact/build/artifact/wire-dev-PR1467-10533.apk)
[Download build #10544](http://192.168.120.33:8080/job/Pull%20Request%20Builder/10544/artifact/build/artifact/wire-dev-PR1467-10544.apk)

PR #1447, opened 2018-03-06
Fix exit animation on GuestOptionsFragment

makingthematrix:

PR #2659, opened 2020-03-03
Kotlin Spike: Navigation structure - Part 1: SettingsMainActivity

gizemb:
## What's new in this PR?

Jira ticket: https://wearezeta.atlassian.net/browse/AN-6658

### Issues

We need to find a way to abstract navigation implementation details (replaceFragment, startActivity, openUrl...) from the view. The current implementation highly couples activity & fragment via direct calls, id references etc., which is something we're trying to avoid in the new architecture.

### Solution

This PR introduces Android Navigation Components here for that purpose, but hides the implementation details behind an interface called Navigator. Hence, we can replace the implementation with some other library or our own code later, if we want to.

Also updated deprecated `Fragment#getFragmentManager` calls to `Fragment#getParentFragment`, since they caused build to fail after adding navigation library dependencies.

### Testing

Seems like there's a testing library for navigation components, but some research is need to be made since we've never tested ui before.

#### APK
[Download build #1447](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1447/artifact/build/artifact/wire-dev-PR2659-1447.apk)
[Download build #1467](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1467/artifact/build/artifact/wire-dev-PR2659-1467.apk)
[Download build #1471](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1471/artifact/build/artifact/wire-dev-PR2659-1471.apk)
[Download build #1537](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1537/artifact/build/artifact/wire-dev-PR2659-1537.apk)
[Download build #1539](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1539/artifact/build/artifact/wire-dev-PR2659-1539.apk)

### DIFF
--- a/app/src/main/kotlin/com/waz/zclient/core/navigation/NavigationComponentsNavigator.kt
+++ b/app/src/main/kotlin/com/waz/zclient/core/navigation/NavigationComponentsNavigator.kt
@@ -1,0 +1,13 @@
+package com.waz.zclient.core.navigation
+
+import androidx.fragment.app.Fragment
+import androidx.navigation.NavController
+import androidx.navigation.fragment.findNavController
+
+internal class NavigationComponentsNavigator(private val navController: NavController) : Navigator {
+    override fun navigateTo(navigationId: Int) {
+        navController.navigate(navigationId)
+    }
+}
+
+fun Fragment.navigator(): Navigator = NavigationComponentsNavigator(findNavController())
--- a/app/src/main/kotlin/com/waz/zclient/core/navigation/NavigationComponentsNavigator.kt
+++ b/app/src/main/kotlin/com/waz/zclient/core/navigation/NavigationComponentsNavigator.kt
@@ -10,18 +10,19 @@ import androidx.navigation.fragment.findNavController
 import java.lang.ref.WeakReference
 
 internal class NavigationComponentsNavigator(
-    private val navController: NavController,
+    navController: NavController,
     activity: Activity?
 ) : Navigator {
 
+    private val navController = WeakReference<NavController>(navController)
     private val activityWeakRef = WeakReference<Activity>(activity)
 
     override fun navigateTo(navigationId: Int) {
-        navController.navigate(navigationId)
+        navController.get()?.navigate(navigationId)
     }
 
     override fun navigateTo(navigationId: Int, bundle: Bundle) {
-        navController.navigate(navigationId, bundle)
+        navController.get()?.navigate(navigationId, bundle)
     }
 
     //TODO: do we need a custom navigator (for backstack, etc.)
--- a/app/src/main/kotlin/com/waz/zclient/core/navigation/NavigationComponentsNavigator.kt
+++ b/app/src/main/kotlin/com/waz/zclient/core/navigation/NavigationComponentsNavigator.kt
@@ -1,11 +1,21 @@
 package com.waz.zclient.core.navigation
 
+import android.app.Activity
+import android.content.Intent
+import android.net.Uri
 import android.os.Bundle
 import androidx.fragment.app.Fragment
 import androidx.navigation.NavController
 import androidx.navigation.fragment.findNavController
+import java.lang.ref.WeakReference
 
-internal class NavigationComponentsNavigator(private val navController: NavController) : Navigator {
+internal class NavigationComponentsNavigator(
+    private val navController: NavController,
+    activity: Activity?
+) : Navigator {
+
+    private val activityWeakRef = WeakReference<Activity>(activity)
+
     override fun navigateTo(navigationId: Int) {
         navController.navigate(navigationId)
     }
@@ -13,6 +23,11 @@ internal class NavigationComponentsNavigator(private val navController: NavContr
     override fun navigateTo(navigationId: Int, bundle: Bundle) {
         navController.navigate(navigationId, bundle)
     }
+
+    //TODO: do we need a custom navigator (for backstack, etc.)
+    override fun navigateTo(uri: Uri) {
+        activityWeakRef.get()?.startActivity(Intent(Intent.ACTION_VIEW, uri))
+    }
 }
 
-fun Fragment.navigator(): Navigator = NavigationComponentsNavigator(findNavController())
+fun Fragment.navigator(): Navigator = NavigationComponentsNavigator(findNavController(), activity)
--- a/app/src/main/kotlin/com/waz/zclient/core/navigation/NavigationComponentsNavigator.kt
+++ b/app/src/main/kotlin/com/waz/zclient/core/navigation/NavigationComponentsNavigator.kt
@@ -1,5 +1,6 @@
 package com.waz.zclient.core.navigation
 
+import android.os.Bundle
 import androidx.fragment.app.Fragment
 import androidx.navigation.NavController
 import androidx.navigation.fragment.findNavController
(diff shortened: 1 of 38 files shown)


app/src/main/res-migration/navigation/settings_nav_graph.xml
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="utf-8"?>
```
gizemb: this file is somewhat auto-generated. It is auto-created and then edited through drag&drop menu, which produces the xml code.

app/src/main/kotlin/com/waz/zclient/settings/account/SettingsAccountFragment.kt
```diff
@@ -106,7 +107,7 @@ class SettingsAccountFragment : Fragment(R.layout.fragment_settings_account) {
             settingsAccountViewModel.onResetPasswordClicked()
         }
         settingsAccountViewModel.resetPasswordUrlLiveData.observe(viewLifecycleOwner) {
-            openUrl(it)
+            navigator().navigateTo(Uri.parse(it))
```
mejdoo: Great 👍

app/src/main/kotlin/com/waz/zclient/settings/main/SettingsMainFragment.kt
```diff
@@ -31,23 +24,18 @@ class SettingsMainFragment : Fragment(R.layout.fragment_settings_main), OnItemCl
...
-    override fun onItemClicked(position: Int) {
+    override fun onItemClicked(position: Int) = with(navigator()) {
         when (position) {
-            ACCOUNT -> replaceFragment(SettingsAccountFragment.newInstance())
-            DEVICES -> replaceFragment(SettingsDeviceListFragment.newInstance())
-            OPTIONS -> replaceFragment(SettingsOptionsFragment.newInstance())
-            ADVANCED -> replaceFragment(SettingsAdvancedFragment.newInstance())
-            SUPPORT -> replaceFragment(SettingsSupportFragment.newInstance())
-            ABOUT -> replaceFragment(SettingsAboutFragment.newInstance())
+            ACCOUNT -> navigateTo(R.id.action_settingsMainFragment_to_settingsAccountFragment)
```
mejdoo: Looks nicer! 👍

app/src/main/res-migration/navigation/settings_nav_graph.xml
mejdoo: I understand that we need to handle the navigation in graphic mode here and then use the actions ID's in the code. Right?

mejdoo: Awesome!

app/src/main/kotlin/com/waz/zclient/core/navigation/NavigationComponentsNavigator.kt
```diff
@@ -0,0 +1,33 @@
...
+        navController.navigate(navigationId, bundle)
+    }
+
+    //TODO: do we need a custom navigator (for backstack, etc.)
+    override fun navigateTo(uri: Uri) {
+        activityWeakRef.get()?.startActivity(Intent(Intent.ACTION_VIEW, uri))
+    }
+}
+
+fun Fragment.navigator(): Navigator = NavigationComponentsNavigator(findNavController(), activity)
```
android10: I have mixed feelings about this approach, especially since this will create a static method retaining a reference to the activity, even though it is a weak one, I would like to see memory profiling for leaks. 

I would also like to know the reason why an extension function (seeing we are abusing a bit).

My approach would be to create an injectable component being injected where it corresponds (could be a ```@singleton```. In the end this is a collaborator for facilitating navigation and not an embedded component. 

A few years ago I did something similar without the graph part:
https://github.com/android10/Android-CleanArchitecture-Kotlin/blob/master/app/src/main/kotlin/com/fernandocejas/sample/core/navigation/Navigator.kt

app/src/main/kotlin/com/waz/zclient/features/settings/main/SettingsMainFragment.kt
```diff
@@ -31,23 +24,18 @@ class SettingsMainFragment : Fragment(R.layout.fragment_settings_main), OnItemCl
...
-    override fun onItemClicked(position: Int) {
+    override fun onItemClicked(position: Int) = with(navigator()) {
         when (position) {
-            ACCOUNT -> replaceFragment(SettingsAccountFragment.newInstance())
-            DEVICES -> replaceFragment(SettingsDeviceListFragment.newInstance())
-            OPTIONS -> replaceFragment(SettingsOptionsFragment.newInstance())
-            ADVANCED -> replaceFragment(SettingsAdvancedFragment.newInstance())
-            SUPPORT -> replaceFragment(SettingsSupportFragment.newInstance())
-            ABOUT -> replaceFragment(SettingsAboutFragment.newInstance())
+            ACCOUNT -> navigateTo(R.id.action_settingsMainFragment_to_settingsAccountFragment)
```
android10: Instead of passing all these (for me random) id, why not implement a method in each fragment that returns this id?
Thus, each component establishes how to be called, plus we favor readability. 
For Example:

```kotlin
ACCOUNT -> navigateTo(SettingsAcountFragment.getCallingActionId())
```

android10: and if you need to pass parameters, you can also pass them to this method and inside it you are encapsulating all the calling logic where it corresponds.

app/src/main/res-migration/navigation/settings_nav_graph.xml
gizemb: yes, you can reference another graph from a graph. I'll come up with an example as soon as I find some free time.

app/src/main/kotlin/com/waz/zclient/features/settings/main/SettingsMainFragment.kt
gizemb: hmm, we can change the namings but it is more convenient to keep the auto generated ones. 
What you're suggesting is to assign id's only to destination points, which is aligned with our previous static factory method approach, but it's not possible with the arch. components because it treats destinations as a pair of (start-end) points.
for instance if you have a screen C, that you can navigate from different screens A and B, you'd have 2 destinations:
from_A_to_C, and from_B_to_C.

I agree about your point on restricting the arguments to the destination fragment. I used bundle() methods to harden the constraints but they can of course be called without any arguments, which leaves some room for misuse.
Arch. components have a solution for that, called "safe args" but it makes code generation that I personally think is an overkill, so I used a manual approach here. we can discuss further.

app/src/main/kotlin/com/waz/zclient/core/navigation/NavigationComponentsNavigator.kt
gizemb: I don't think there shouldn't be any problem about memory leak here. This is like a getter method for the fragment, added for convenience. 
With this structure, each fragment has its own navigator component attached o the fragment's scope. to me, it's better than having a singleton that's scoped to the app's lifecycle, waiting to be called in case someone needs it.

I'm going to weaken NavController's reference also, just in case.

gizemb:
closing this since it's outdated. the PR can be a reference for the future, but for now it's too much work to rebase this.

PR #2663, opened 2020-03-05
AN-6714 && AN-6716: Conversation Search feature and about screen is broken 

BradleyWire:
### Issues

Conversation Search feature had broken after the migrated database code was implemented

### Causes

> E/SQLiteQuery: exception: unable to use function MATCH in the requested context (code 1 SQLITE_ERROR); query:

### Solutions

MessageContentIndex relies on [FTS3](https://www.sqlite.org/fts3.html) to perform faster searches. The new migrated room database created a normal table for this rather than a virtual one for the Scala code to query to. 

### Testing

Conversation searching feature should now behave as it did before

### Notes

This table is the only table that relies on FTS3/FTS4 so we shouldn't face the same problem again. 

#### APK
[Download build #1455](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1455/artifact/build/artifact/wire-dev-PR2663-1455.apk)

### DIFF
--- a/app/src/main/kotlin/com/waz/zclient/settings/about/SettingsAboutActivity.kt
+++ b/app/src/main/kotlin/com/waz/zclient/settings/about/SettingsAboutActivity.kt
@@ -5,12 +5,20 @@ import android.content.Intent
 import android.os.Bundle
 import androidx.appcompat.app.AppCompatActivity
 import com.waz.zclient.R
+import com.waz.zclient.core.extension.createScope
 import com.waz.zclient.core.extension.replaceFragment
+import com.waz.zclient.settings.di.SETTINGS_SCOPE
+import com.waz.zclient.settings.di.SETTINGS_SCOPE_ID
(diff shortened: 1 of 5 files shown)


app/src/main/kotlin/com/waz/zclient/settings/about/SettingsAboutActivity.kt
```diff
@@ -5,13 +5,21 @@ import android.content.Intent
...
 import com.waz.zclient.core.extension.replaceFragment
+import com.waz.zclient.settings.di.SETTINGS_SCOPE
+import com.waz.zclient.settings.di.SETTINGS_SCOPE_ID
 import kotlinx.android.synthetic.main.activity_settings_about.*
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.InternalCoroutinesApi
 
 class SettingsAboutActivity : AppCompatActivity() {
 
+    private val scope = createScope(
```
mejdoo: 👍

Issue #1569, opened 2018-05-16
Sent Pictures are no longer visible in Wire after being deleted from the Android filesystem


Larx:
## The problem

If I send a photo (possibly other stuff, too) via Android's share feature from the filesystem, they disappear from the local Wire instance as soon as I delete them from the device. 
They are, however, obviously still visible for the recipient of the message, and even in my own Wire account as soon as I access it with Wire instances installed on other devices.

## Environment

This and the other bullets are IMHO not applicable, it rather seems a design question that the local Wire instance does rely on sent items being present on the local file system.

## Proposal

If a sent item is no longer available on the local filesystem, Wire should resort to the item as stored for the communication and available for the other participients of the conversation.

someoneEsle:
duplicate #932

PR #2665, opened 2020-03-05
fix: Default config values

makingthematrix:
A few buildConfigFields and manifestPlaceholders  were missing from app/build.gradle. This usually is not a problem, but after removing Gradle's cache and doing `./gradlew clean`, I wasn't able to compile the app again without them.

#### APK
[Download build #1489](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1489/artifact/build/artifact/wire-dev-PR2665-1489.apk)
[Download build #1569](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1569/artifact/build/artifact/wire-dev-PR2665-1569.apk)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -100,8 +100,7 @@ android {
             customURLScheme   : buildtimeConfiguration.configuration.custom_url_scheme,
             applicationLabel  : "Wire Dev",
             applicationIcon   : "@mipmap/ic_launcher_wire_dev",
-            sharedUserId      : "",
-            internal_features : "true"
+            sharedUserId      : ""
         ]
 
         buildConfigField 'Integer', 'MAX_ACCOUNTS',                  "$buildtimeConfiguration.configuration.maxAccounts"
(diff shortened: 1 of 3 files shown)


makingthematrix:
build again

app/build.gradle
```diff
@@ -96,7 +96,13 @@ android {
         applicationId "com.waz.zclient"
         testInstrumentationRunner "com.waz.background.TestRunner"
 
-        manifestPlaceholders = [customURLScheme: buildtimeConfiguration.configuration.custom_url_scheme]
+        manifestPlaceholders = [
+            customURLScheme   : buildtimeConfiguration.configuration.custom_url_scheme,
+            applicationLabel  : "Wire Dev",
+            applicationIcon   : "@mipmap/ic_launcher_wire_dev",
+            sharedUserId      : "",
+            internal_features : "true"
```
android10: What is this flag doing? What is the difference between ```internal_features``` and ```developer_features_enabled```?

In any case I would be a bit defensive here and put this to ```false``` by default. Does that make sense? Or at least write a test checking this out in order to not accidentally ship any WIP or internal funcionality...

default.json
```diff
@@ -37,5 +37,8 @@
   "block_on_password_policy": false,
   "wipe_on_cookie_invalid": false,
   "force_private_keyboard": false,
-  "password_max_attempts": 0
+  "password_max_attempts": 0,
+  "developer_features_enabled": false,
```
android10: This is what I mean with the above comment.

app/build.gradle
makingthematrix: Good catch. I don't think it's used at all in practice. There is a line in `AndroidManifest.xml`:
```
        <!-- Internal features from SE -->
        <meta-data android:name="INTERNAL" android:value="${internal_features}"/>
```
but then we don't use that `INTERNAL` flag anywhere. I guess I can delete it.

android10: Go for it if you feel comfortable/safe. Is this something that maybe we are using for custom builds? It does not look like it doh....Just thinking loudly. ;)

makingthematrix:
build again

app/build.gradle
makingthematrix: done

default.json
makingthematrix: ok

android10:
@makingthematrix is this something we want to merge? It looks like it is 25 days old.

PR #2688, opened 2020-03-16
AN-6745: Migration between app versions freezes the app - Part 3

mejdoo:
## What's new in this PR?

Part 3 of #2679.

Added Test cases for: User tables entry migration from version 126 to 127

JIRA: https://wearezeta.atlassian.net/browse/AN-6745
#### APK
[Download build #1542](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1542/artifact/build/artifact/wire-dev-PR2688-1542.apk)
[Download build #1543](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1543/artifact/build/artifact/wire-dev-PR2688-1543.apk)
[Download build #1545](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1545/artifact/build/artifact/wire-dev-PR2688-1545.apk)
[Download build #1547](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1547/artifact/build/artifact/wire-dev-PR2688-1547.apk)

PR #2689, opened 2020-03-17
AN-6745: Migration between app versions freezes the app - Part 4

BradleyWire:
## What's new in this PR?

Part 4 of https://github.com/wireapp/wire-android/pull/2688

Started migrating tables without unit tests to see if any other issues arise and then will unit test prior to the release of this code (to ensure a smooth release). 

#### APK
[Download build #1551](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1551/artifact/build/artifact/wire-dev-PR2689-1551.apk)


storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabaseMigration.kt
```diff
@@ -32,88 +27,243 @@ val USER_DATABASE_MIGRATION_126_TO_127 = object : Migration(126, 127) {
     override fun migrate(database: SupportSQLiteDatabase) {
         if (BuildConfig.KOTLIN_CORE) {
             migrateClientTable(database)
```
mejdoo: Do we need to keep this only for dev mode? We should migrate this table with the old columns to fix the bug.

BradleyWire: Yes, we will migrate it eventually. But there are currently too many dependencies on the ClientsEntity for the old logic we used way back. So, we need to keep this as is until we refactor or remove the clients stuff.

storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabaseMigration.kt
```diff
@@ -32,88 +27,312 @@ val USER_DATABASE_MIGRATION_126_TO_127 = object : Migration(126, 127) {
...
-            migrateKeyValuesTable(database)
         }
 
-        //Needed in production
         migrateUserTable(database)
+        migrateKeyValuesTable(database)
```
android10: My only point is to keep in mind that this class will become a monster at some point. Totally fine for now doh.

BradleyWire: Yeah i'll have to designate this class purely for 126 - 127 migration.

BradleyWire: and tests will come in once every table has been migrated, so we can verify that we've actually solved the problem. But, they will come before the release that is 100%

BradleyWire: We could in theory think about having migrations for each table in their own class. Let's discuss it in the collective.

BradleyWire:
build again

PR #2691, opened 2020-03-17
AN-6745: Migration between app versions freezes the app - Part 6

mejdoo:
## What's new in this PR?

Part 6 of #2689

Fixed old migration for Users and KeyValues


storage/src/main/kotlin/com/waz/zclient/storage/db/users/model/UserEntity.kt
```diff
@@ -34,16 +34,16 @@ data class UserEntity(
...
     @ColumnInfo(name = "accent")
-    val accentId: Int?,
+    val accentId: Int,
 
     @ColumnInfo(name = "skey")
-    val sKey: String?,
+    val sKey: String,
 
     @ColumnInfo(name = "connection")
-    val connection: String?,
```
android10: oh yeahhh...all these ```null``` values gone :heart:

mejdoo: 👍

mejdoo:
build again

PR #2679, opened 2020-03-11
AN-6745: Migration between app versions freezes the app - Part 1.

BradleyWire:
## What's new in this PR?

JIRA: https://wearezeta.atlassian.net/browse/AN-6745

### Issues

When you update the application from 3.46 to 3.47 it'll freeze the application due to throwing an exception around unhandled migrations. 

### Causes

The Schema for the Scala database (SQLiteDatabase) doesn't explicitly define "NON NULL" modifiers to columns, so when the room creates a "room_master_table" for each of the tables it's mimicking it doesn't correspond to the mandatory NON NULL values that come with @PrimaryKey annotation (amongst other things).  

### Solutions

To migrate all tables in the database with the exact schema as it was before but with the explicit NON-NULL modifiers attached. Any breaking schema modifications will need to be done the next migration away from **127** and **25**.  

## Notes

All migrations will be Unit Tested for every version to ensure the data remains in tact. 
#### APK
[Download build #1518](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1518/artifact/build/artifact/wire-dev-PR2679-1518.apk)
[Download build #1519](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1519/artifact/build/artifact/wire-dev-PR2679-1519.apk)
[Download build #1522](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1522/artifact/build/artifact/wire-dev-PR2679-1522.apk)
[Download build #1526](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1526/artifact/build/artifact/wire-dev-PR2679-1526.apk)
[Download build #1527](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1527/artifact/build/artifact/wire-dev-PR2679-1527.apk)

Issue #1577, opened 2018-05-28
Markdown not rendered on Android


gr4v1t0n:
# The problem
The markdown described on https://support.wire.com/hc/en-us/articles/115002676029-How-can-I-use-markdown-to-format-text-on-Wire- is not working in the Android app. ("format some text inside **any** conversation on Wire" is unfortunately not working for all clients yet).
Would be nice to have it implemented on all clients.

## Environment
* Wire version that exhibits the issue: 3.10.442
* Android OS version used to run Wire: Android 7.1.2
* Is this a custom firmware or a stock one: LineageOS 
* Mobile phone model/manufacturer: Samsung Galaxy S4 Mini (GT-I9195)
* Mobile network type (EDGE/LTE/Wi-Fi/Offline): LTE/Wi-Fi

## Details
Markdown is rendered on the Windows client, but not on my phone (viewing the same conversation).

## Screenshots
![grafik](https://user-images.githubusercontent.com/39697677/40612385-913293f0-627a-11e8-9ddc-5578e8737cb6.png)

red-avtovo:
Hi @gr4v1t0n ! This feature was never developed on Android.

marcoconti83:
Hi @gr4v1t0n ,
this feature was ported to the Android app and is already in the public version - give it a try!

PR #2695, opened 2020-03-18
AN-6745: Migration between app versions freezes the app - Part 8.

BradleyWire:
## What's new in this PR?

Part 8 of #2679.

Migrated indexes over to room to correspond to the values given by Scala. 
#### APK
[Download build #1577](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1577/artifact/build/artifact/wire-dev-PR2695-1577.apk)
[Download build #1578](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1578/artifact/build/artifact/wire-dev-PR2695-1578.apk)


storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase126To127Migration.kt
```diff
@@ -1,10 +1,9 @@
-@file:Suppress("MagicNumber", "TooManyFunctions")
+@file:Suppress("MagicNumber", "TooManyFunctions", "LargeClass")
```
BradleyWire: We need to find a solution for this before the release.

storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase126To127Migration.kt
```diff
@@ -25,10 +24,6 @@ private const val NEW_CLIENT_TYPE_KEY = "type"
 
 val USER_DATABASE_MIGRATION_126_TO_127 = object : Migration(126, 127) {
     override fun migrate(database: SupportSQLiteDatabase) {
-        if (BuildConfig.KOTLIN_CORE) {
```
gizemb: why did we remove this check from this migration?

BradleyWire: When we update from version 3.46, Room expects everything to be migrated  (I assumed based on the lack of schema/database hash we have from a previous version). It'll throw a: `2020-03-18 16:12:38.499 29826-29911/com.wire.x E/Preference: Error while getting signal with preference key self_client (def: Unregistered). Exception is: java.lang.IllegalStateException: Migration didn't properly handle: Clients(com.waz.zclient.storage.db.userclients.UserClientsEntity).` otherwise.

BradleyWire: Okay, for some reason we have an overloaded `migrateClientsTable` method. I'll look at tested the correct one this time.

BradleyWire: Sorted it.

PR #2725, opened 2020-03-25
UserDatabase Migration: Remove duplicated client entity from user database schema 

mejdoo:
### Issues

`ClientEntity` is blocking the user database migration

### Causes

`ClientEntity` used by the Kotlin code in settings devices screen **(dev mode only)** does not have the same fields of the current `Clients` table. 

### Solutions

Removed ClientEntity and clientsDao temporary from user database schema to unblock the migration. (already replaced by `UserClientDao` and `UserClientsEntity`)

#### APK
[Download build #1662](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1662/artifact/build/artifact/wire-dev-PR2725-1662.apk)

### DIFF
--- a/app/src/main/kotlin/com/waz/zclient/clients/di/ClientsModule.kt
+++ b/app/src/main/kotlin/com/waz/zclient/clients/di/ClientsModule.kt
@@ -7,7 +7,6 @@ import com.waz.zclient.clients.datasources.remote.ClientsApi
 import com.waz.zclient.clients.datasources.remote.ClientsRemoteDataSource
 import com.waz.zclient.clients.mapper.ClientMapper
 import com.waz.zclient.core.network.NetworkClient
-import com.waz.zclient.storage.db.UserDatabase
 import org.koin.core.module.Module
 import org.koin.dsl.module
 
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/waz/zclient/clients/di/ClientsModule.kt
```diff
@@ -17,5 +17,5 @@ val clientsModule: Module = module {
     factory { get<NetworkClient>().create(ClientsApi::class.java) }
     factory { ClientsRemoteDataSource(get(), get()) }
     factory { ClientsLocalDataSource(get()) }
-    factory { get<UserDatabase>().clientsDao() }
+    //factory { get<UserDatabase>().clientsDao() }
```
BradleyWire: I guess we can remove this if we don't need it? or add a `TODO` regarding the intention of this change

mejdoo: I removed it

BradleyWire:
To highlight anything, me and Mejdoo have had a discussion around the impact this change will have on the Kotlin: Devices Screen and have added an item to the Kotlin code review to adapt that screen to the old Scala client table.

PR #2731, opened 2020-03-27
fix: Display websites' previews on Android 9-10

makingthematrix:
fixes https://wearezeta.atlassian.net/browse/AN-6685

Since version 9, Android by default disallows cleartext traffic and if the website uses that  we get an exception:
```
java.net.UnknownServiceException: CLEARTEXT communication to heise.de not permitted by network security policy
```
One way to fix it would be to disable this security check and allow cleartext again, but I don't want to do that. Instead, I decided to handle that specific exception and if it happens, retry the request with "https://" instead of "http://". In many cases that will  be enough. In the others, it simply means that the problem is on the website's side and we won't show the preview.

The weblink itself is not affected - if the user clicks on it, the webpage starting with "http://" will be opened.
#### APK
[Download build #1696](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1696/artifact/build/artifact/wire-dev-PR2731-1696.apk)

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/client/OpenGraphClient.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/client/OpenGraphClient.scala
@@ -23,9 +23,9 @@ import java.net.URL
 import com.waz.api.impl.ErrorResponse
 import com.waz.sync.client.OpenGraphClient.{OpenGraphData, OpenGraphImage}
 import com.waz.threading.{CancellableFuture, Threading}
-import com.waz.utils.wrappers.URI
+import com.waz.utils.wrappers.{AndroidURIUtil, URI}
 import com.waz.utils.{IoUtils, JsonDecoder, JsonEncoder}
-import com.waz.znet2.http.HttpClient.{ConnectionError, HttpClientError}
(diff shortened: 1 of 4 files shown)


wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/client/OpenGraphClient.scala
```diff
@@ -46,14 +46,21 @@ class OpenGraphClientImpl(implicit httpClient: HttpClient) extends OpenGraphClie
...
       .map(response => Right(response.data))
-      .recover {
-        case _: ConnectionError => Right(None)
-        case err: HttpClientError => Left(ErrorResponse.errorResponseConstructor.constructFrom(err))
+      .recoverWith {
+        case _: UnknownServiceError if url.getProtocol == "http" =>
+          loadMetadata(
+            AndroidURIUtil.parse(
+              uri.toString.toLowerCase.trim.replaceFirst("http", "https")
+            )
```
gizemb: can't we do sth like this: `uri.buildUpon().scheme("https").build()`
not very important though, I approved the PR anyway.

makingthematrix: I tried something like that but it would require changes to the whole `com.waz.utils.wrappers.URI`.

makingthematrix:
build again

PR #2747, opened 2020-04-03
fix: Dismiss associated notification when the conversation is opened

makingthematrix:
fixes https://wearezeta.atlassian.net/browse/AN-6822

While fixing the bug which processed the notifications in O(n^2), I accidentally removed the old way to dismiss notifications when the user opens the conversation associated with them. The old way was very convoluted: it tried to remove all displayed notifications and recreate them (and failed). Instead I made a simpler fix which just cancels notifications with matching ids. A notification id is created from the user id and the conversation id (no need for unique ids per message), plus there are "summary notifications" with ids based only on user ids). So if a conversation is opened by the user, we already know which notifications should be dismissed.
#### APK
[Download build #1770](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1770/artifact/build/artifact/wire-dev-PR2747-1770.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/notifications/controllers/MessageNotificationsController.scala
+++ b/app/src/main/scala/com/waz/zclient/notifications/controllers/MessageNotificationsController.scala
@@ -64,7 +64,7 @@ class MessageNotificationsController(bundleEnabled: Boolean = Build.VERSION.SDK_
 
   private lazy val notificationManager   = inject[NotificationManagerWrapper]
 
-  private lazy val selfId                = inject[Signal[UserId]]
+  private lazy val selfId                = inject[Signal[Option[UserId]]]
   private lazy val soundController       = inject[SoundController]
   private lazy val navigationController  = inject[NavigationController]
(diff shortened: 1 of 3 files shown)


app/src/main/scala/com/waz/zclient/notifications/controllers/MessageNotificationsController.scala
```diff
@@ -75,22 +75,36 @@ class MessageNotificationsController(bundleEnabled: Boolean = Build.VERSION.SDK_
...
+        (if (selfId != accId || !uiActive) Set.empty[ConvId]
+        else page match {
+          case Page.CONVERSATION_LIST => convs
+          case Page.MESSAGE_STREAM    => Set(convId).flatten
+          case _                      => Set.empty[ConvId]
+        })
+    }.toMap
+
+  /*
+  Clears notifications already displayed in the tray when the user opens the conversation associated
```
BradleyWire, line 95: Very minor but the formatting of this comment is off

PR #2748, opened 2020-04-03
fix : revert MessageContentIndex table to fix search content with special characters

mejdoo:
## What's new in this PR?

I reverted back the virtual FTS table `MessageContentIndex` to fix the issues related to content search with special characters.

https://wearezeta.atlassian.net/browse/AN-6804
https://wearezeta.atlassian.net/browse/AN-6805

### Testing

- Fresh Install the version 3.46
- Login and write some messages 
- Install the version 3.47 -> No crash/No stuck
- Search for content with special characters in your conversations i.e umlauts "aa" or "ää" or "oo" or "öö"

### DIFF
--- a/storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase126To127Migration.kt
+++ b/storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase126To127Migration.kt
@@ -754,6 +754,29 @@ val USER_DATABASE_MIGRATION_126_TO_127 = object : Migration(126, 127) {
         )
     }
 
+    private fun migrateMessageContentIndexTable(database: SupportSQLiteDatabase) {
+        val tempTableName = "MessageContentIndexTemp"
+        val originalTableName = "MessageContentIndex"
+        val messageId = "message_id"
+        val convId = "conv_id"
+        val content = "content"
+        val time = "time"
+        val createTempTable = """
+              CREATE VIRTUAL TABLE $tempTableName using fts4(
+              $messageId TEXT NOT NULL,
+              $convId TEXT NOT NULL,
+              $content TEXT NOT NULL,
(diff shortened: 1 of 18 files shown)


storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase126To127Migration.kt
```diff
@@ -787,6 +788,34 @@ val USER_DATABASE_MIGRATION_126_TO_127 = object : Migration(126, 127) {
         )
     }
 
+    private fun migrateMessageContentIndexTable(database: SupportSQLiteDatabase) {
```
gizemb, line 791: can you use `executeSimpleMigration` method?

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/content/ConvMessagesIndex.scala
```diff
@@ -33,15 +33,14 @@ import scala.collection.mutable
 import scala.concurrent.Future
 import scala.concurrent.duration._
 
-class ConvMessagesIndex(convId:      ConvId,
-                        messages:    MessagesStorage,
-                        selfUserId:  UserId,
-                        users:       UsersStorage,
-                        convs:       ConversationStorage,
+class ConvMessagesIndex(convId: ConvId, messages: MessagesStorageImpl,
+                        selfUserId: UserId, users: UsersStorage,
```
gizemb, line 37: can you reformat this?

storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase126To127Migration.kt
mejdoo: done

Issue #1808, opened 2018-09-27
Camera exposure


Da0012:
## The problem

The camera does not adjust the exposure well, many times it is necessary to use the smartphone camera,at night it is not possible to make a video call (the smartphone camera works well, whatsapp's and other applications too.).

## Environment

* Wire version that exhibits the issue:3.18.650
* Last Wire version that did not exhibit the issue (if applicable): any version before the previous month
* Android OS version used to run Wire: 7.1.1
* Is this a custom firmware or a stock one: stock
* Mobile phone model/manufacturer: zenfone 4 max
* Mobile network type (EDGE/LTE/Wi-Fi/Offline):

someoneEsle:
Sounds like a duplicate of https://github.com/wireapp/wire-android/issues/671

PR #2758, opened 2020-04-07
Migration: Anomalies 

BradleyWire:
### Issues

Migration is showing a black screen for some users using the beta version of 3.47

### Causes

Still unknown but this PR fixes some anomalies found within the migration after a line by line inspection of all the migration code. 

### Solutions

N/A right now. 

### Testing

There are numerous scenarios I tested here:

- Null preview In DownloadAssets (Voice message) = Still works / Wasn't an issue
- Stored BLOB reference for PushNotificationEvent (By sending Images) - Still works
- Multiple accounts (multiple database) migration  - Still works 
- Multiple accounts + one logged out users (which keeps the database stored so 3 user databases in total) - Still works 

## Notes

**Still unable to reproduce the black screen issue.** 

#### APK
[Download build #1808](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1808/artifact/build/artifact/wire-dev-PR2758-1808.apk)


storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase126To127Migration.kt
```diff
@@ -218,7 +218,7 @@ val USER_DATABASE_MIGRATION_126_TO_127 = object : Migration(126, 127) {
                 $convId TEXT NOT NULL,
                 msg_type TEXT NOT NULL, 
                 user_id TEXT NOT NULL,
-                content TEXT,
+                content TEXT NOT NULL,
```
BradleyWire, line 221: May need to move this change to a new migration

BradleyWire:
> Should we add the difference in "preview" field in DownloadAssets? Otherwise, we will have to create a migration to make in NOT NULL in the future (I guess).
> 
> I believe you should merge to release, not to develop. We're handling this like a hotfix, right?
> 
> Also, you should update migration tests, since some values were ints but they're booleans now. There should be some compile errors. Though they're not run on Jenkins.

Ah, yes, I'll take a look at the tests now.

gizemb:
we discussed with @BradleyWire and decided that DownloadAssets table can stay the same, as it never has a null value in database right now.

Issue #1869, opened 2018-11-17
404 to wiretranslations-android


suarezu:
## The problem

Hi there, the link in the readme.MD to wiretranslations-android seems broken. what's the up to date status of it?

marcoconti83:
Hi @suarezu ,
Thanks for the report, [I just fixed it](https://github.com/wireapp/wire-android/commit/2d2193a7738da01ca63a2256c88dfc0b080978e7).

PR #2768, opened 2020-04-14
Scala Migration tests - Part 2

gizemb:
## What's new in this PR?

Jira issue: https://wearezeta.atlassian.net/browse/AN-6839

### Issues

See #2763 

### Causes

See #2763 

### Solutions

This PR improves testing steps on #2763 by actually writing to and reading from the database.
The test steps are:

1. Create the ZMessaging database as it was in 3.46 version
2. Insert a value through scala daos and data models. Use null or empty values wherever possible. Close legacy ZMessaging db.
3. Create the ZMessaging db in the way that it's now in 3.47. This creates a RoomDaoDB, which wraps our Kotlin Room db.
4. When we create, the Room db executes the migration 126-127. 
5. Try to read the value written by using Room database, Room daos and Kotlin data models (Entities). See if there's any crash.

**The findings and fixes:**
- Checkout commit a431eba2b, where there are only tests and no fixes yet.
- Run ./gradlew :wire-android-sync-engine:zmessaging:testDebugUnitTest
- Results & fixes:
1. ContactHashesMigrationTest

> SQLiteException: [19] DB[4] step() [INSERT INTO ContactHashesTemp SELECT * FROM ContactHashes]DB[4][C] [ContactHashesTemp.hashes may not be NULL]

Meaning: The value read from Scala db is null. We're trying to insert a null value to a field with NOT NULL constraint.
Fix: Make `hashes` field nullable in `ContactHashesEntity`, and fix related migration code.

2. Assets2MigrationTest

> Parameter specified as non-null is null: method com.waz.zclient.storage.db.assets.AssetsEntity.<init>, parameter sha
> java.lang.IllegalArgumentException: Parameter specified as non-null is null: method com.waz.zclient.storage.db.assets.AssetsEntity.<init>, parameter sha
> 	at com.waz.zclient.storage.db.assets.AssetsEntity.<init>(AssetsEntity.kt)
> 	at com.waz.zclient.storage.db.assets.AssetsDao_Impl$3.call(AssetsDao_Impl.java:155)
> 	at com.waz.zclient.storage.db.assets.AssetsDao_Impl$3.call(AssetsDao_Impl.java:114)
> 	at androidx.room.CoroutinesRoom$Companion$execute$2.invokeSuspend(CoroutinesRoom.kt:54)
> 	(Coroutine boundary)
> 	at com.waz.KotlinMigrationHelper$assertAssetsEntity$1.invokeSuspend(KotlinMigrationHelper.kt:41)

Meaning: This is the line that we read from Room database: 
`val entity = roomDB.assetsDao().allAssets()[0]`
There's no problem during migration, but there is when reading.
Fix: Make `sha` field nullable in `AssetsEntity`

3. UploadAssetMigrationTest

Same problem as (2), but this time with `UploadAssetsEntity`.

4. ConversationsMigrationTest

> SQLiteException: [19] DB[18] step() [INSERT INTO ConversationsTemp SELECT * FROM Conversations]DB[18][C] [ConversationsTemp.access may not be NULL]

Meaning: Same as (1).
Fix: Make `access` field nullable in `ConversationsEntity`, and fix related migration code.
#### APK
[Download build #1869](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1869/artifact/build/artifact/wire-dev-PR2768-1869.apk)
[Download build #1879](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1879/artifact/build/artifact/wire-dev-PR2768-1879.apk)


BradleyWire:
Could you possibly add more to the description here? What were your findings from these changes? What was wrong before? Just for longevity really.

storage/src/main/kotlin/com/waz/zclient/storage/db/assets/UploadAssetsEntity.kt
```diff
@@ -54,4 +55,37 @@ data class UploadAssetsEntity(
 
     @ColumnInfo(name = "asset_id")
     val assetId: String?
-)
+) {
+    override fun equals(other: Any?): Boolean =
```
gizemb, line 59: `equals` and `hashcode` are overridden to be able to make equality checks in unit tests. The built-in equals method of Kotlin data classes does not play well here because we have `ByteArray`s, which should be checked by `contentEquals()`, rather than simple `==` check.
Actually, I think this should be done for all Entities that contains `ByteArray`s, but this is not a problem for us yet, since we only check for `null` `ByteArray`s for now, and `==` returns correctly in null values.

makingthematrix:
Great work!

I'm adding default values to the schema in another PR. I will also include the changes you made so it will be easier to merge the two later on.

gizemb:
build again

gizemb:
build again

gizemb:
build again

Issue #1970, opened 2019-01-31
Long audio pauses eventually resumed with no audible indication that session is still active


fortran77:
## The problem

During a Wire audio call between two parties, audio blanks out for long periods of time, as much as 1–2 minutes. Shortly after audio blanks out, a few fast beeps are heard, then only silence. While the audio is blanked out, the display shows the call still in progress. Most of the time, the audio will eventually resume. Some of the time the audio never resumes, but the connection still shows as active, so the user has to manually disconnect then redial.

## Environment

This has been observed on several Android devices at my end during the past 12 months or longer. At my end, it was most recently observed on (a) a stock Galaxy Tab S2 tablet running Android 7.0 and Wire 3.26.694 and (b) a stock Moto X4 running Android 9.0 and Wire 3.26.694. These devices were always connected to reliable Wi-Fi. I have been communicating with several different people using both Android and IOS devices whose exact models are unavailable (but probably not relevant).

## Details

There are two slightly different issues here.

First, the fact of the long audio gaps. At least in some of the cases, devices at both ends were connected to reliable Wi-Fi. So it's unlikely to be a network issue. More likely the Wire infrastructure itself is producing the long silences. A similar problem seems to have been reported in Issue #1405.

Second, during the long pauses, there was no indication to the user that audio will eventually resume. So I was not sure whether I should patiently wait or disconnect and redial.

The question of comfort noise was brought up in Issue #754, and the developers responded: 'Our approach to signal users that a call is up or has gone down is by playing audio notifications. Otherwise reliability should be just so high that the question "are you still there" is not needed.'

Both of these points appear to be not currently applicable. I.e., (a) no audio notifications are available that tell the user to keep holding rather than disconnecting and redialing, and (b) audio call reliability is not always high even over Wi-Fi.

I have not attempted to collect debugging logs. However, if the developers are not already aware of the reasons for this issue, I will be happy to create and provide debugging logs upon request.

marcoconti83:
Hi @fortran77 ,
this sounds like an issue somewhere. If the audio is not flowing because the connection failed, the app should play a "blip blip" sound to let you know that the audio is interrupted because of the connection.

The only cases in which you should hear silence are:
- the call is working fine and no one is talking
- the other side muted the microphone
- the other side has some issues with the microphone

If you can pull the logs for this situation, it would help shed light on the issue. 
Thanks!

fortran77:
Thanks, I will collect a debugging log as requested. (It may take some time.)

[...]

PR #2789, opened 2020-04-24
Create personal account - part 11

mejdoo:
## What's new in this PR?

Follow up of #2787 

- Implemented name validation logic 

https://wearezeta.atlassian.net/browse/AN-6854

#### APK
[Download build #1955](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1955/artifact/build/artifact/wire-dev-PR2789-1955.apk)
[Download build #1970](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1970/artifact/build/artifact/wire-dev-PR2789-1970.apk)

### DIFF
--- a/app/src/main/kotlin/com/waz/zclient/feature/auth/registration/personal/email/CreatePersonalAccountNameFragment.kt
+++ b/app/src/main/kotlin/com/waz/zclient/feature/auth/registration/personal/email/CreatePersonalAccountNameFragment.kt
@@ -2,24 +2,49 @@ package com.waz.zclient.feature.auth.registration.personal.email
 
 import android.os.Bundle
 import android.view.View
+import androidx.core.widget.doAfterTextChanged
 import androidx.fragment.app.Fragment
+import androidx.lifecycle.observe
 import com.waz.zclient.R
 import com.waz.zclient.core.extension.replaceFragment
 import com.waz.zclient.core.extension.sharedViewModel
+import com.waz.zclient.core.extension.viewModel
 import com.waz.zclient.feature.auth.registration.di.REGISTRATION_SCOPE_ID
 import kotlinx.android.synthetic.main.fragment_create_personal_account_name.*
 
 class CreatePersonalAccountNameFragment : Fragment(R.layout.fragment_create_personal_account_name) {
+
+    private val createPersonalAccountWithEmailViewModel: CreatePersonalAccountWithEmailViewModel
+        by viewModel(REGISTRATION_SCOPE_ID)
 
     private val emailCredentialsViewModel: EmailCredentialsViewModel
         by sharedViewModel(REGISTRATION_SCOPE_ID)
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         super.onViewCreated(view, savedInstanceState)
+        observeNameValidationData()
+        initNameChangedListener()
         initConfirmationButton()
     }
 
+    private fun observeNameValidationData() {
+        createPersonalAccountWithEmailViewModel.isValidNameLiveData.observe(viewLifecycleOwner) {
+            updateConfirmationButtonStatus(it)
+        }
+    }
+
+    private fun updateConfirmationButtonStatus(enabled: Boolean) {
+        createPersonalAccountNameConfirmationButton.isEnabled = enabled
+    }
+
+    private fun initNameChangedListener() {
+        createPersonalAccountNameEditText.doAfterTextChanged {
+            createPersonalAccountWithEmailViewModel.validateName(it.toString())
+        }
+    }
+
     private fun initConfirmationButton() {
+        updateConfirmationButtonStatus(false)
         createPersonalAccountNameConfirmationButton.setOnClickListener {
             saveName()
             showPasswordInputScreen()
--- a/app/src/main/kotlin/com/waz/zclient/feature/auth/registration/personal/email/CreatePersonalAccountWithEmailViewModel.kt
+++ b/app/src/main/kotlin/com/waz/zclient/feature/auth/registration/personal/email/CreatePersonalAccountWithEmailViewModel.kt
@@ -108,9 +108,9 @@ class CreatePersonalAccountWithEmailViewModel(
         }
     }
 
-    private fun updateNameValidationStatus(enabled: Boolean) {
-        _isValidNameLiveData.postValue(enabled)
-    }
+    private fun updateNameValidationStatus(status: Boolean) =
+        _isValidNameLiveData.postValue(status)
+
 
     fun register(name: String, email: String, password: String, activationCode: String) {
         registerPersonalAccountWithEmailUseCase(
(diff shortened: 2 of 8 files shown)


app/src/main/kotlin/com/waz/zclient/feature/auth/registration/personal/email/CreatePersonalAccountWithEmailViewModel.kt
```diff
@@ -90,6 +96,22 @@ class CreatePersonalAccountWithEmailViewModel(
         }
     }
 
+    fun validateName(name: String) {
+        validateNameUseCase(viewModelScope, ValidateNameParams(name), Dispatchers.Default) {
+            it.fold(::handleValidateNameFailure) { updateNameValidationStatus(true) }
```
android10, line 101: :heart:

app/src/main/kotlin/com/waz/zclient/feature/auth/registration/personal/email/CreatePersonalAccountWithEmailViewModel.kt
```diff
@@ -90,6 +96,22 @@ class CreatePersonalAccountWithEmailViewModel(
...
+        }
+    }
+
+    private fun handleValidateNameFailure(failure: Failure) {
+        if (failure is ValidateNameFailure) {
+            updateNameValidationStatus(false)
+        }
+    }
+
+    private fun updateNameValidationStatus(enabled: Boolean) {
```
android10, line 111: You can get rid of the parenthesis:

```kotlin
    private fun updateNameValidationStatus(enabled: Boolean) = 
        _isValidNameLiveData.postValue(enabled)
```

android10: Also the signature of this function is confusing. What do you mean by ```enabled```? IMO if this function is updating a status, the argument should be called ```status``` maybe?

android10:
Was there any part 10? since  #2783 is not even merged

mejdoo:
part 10 was approved. now it's merged with part 9. I will merge all the changes to develop after build success

app/src/main/kotlin/com/waz/zclient/feature/auth/registration/personal/email/CreatePersonalAccountWithEmailViewModel.kt
mejdoo: done

app/src/main/kotlin/com/waz/zclient/feature/auth/registration/personal/email/CreatePersonalAccountNameFragment.kt
```diff
@@ -2,24 +2,49 @@ package com.waz.zclient.feature.auth.registration.personal.email
...
 import com.waz.zclient.R
 import com.waz.zclient.core.extension.replaceFragment
 import com.waz.zclient.core.extension.sharedViewModel
+import com.waz.zclient.core.extension.viewModel
 import com.waz.zclient.feature.auth.registration.di.REGISTRATION_SCOPE_ID
 import kotlinx.android.synthetic.main.fragment_create_personal_account_name.*
 
 class CreatePersonalAccountNameFragment : Fragment(R.layout.fragment_create_personal_account_name) {
 
+    private val createPersonalAccountWithEmailViewModel: CreatePersonalAccountWithEmailViewModel
```
gizemb, line 17: this viewmodel is not a shared one, yet it contains lots of things not related to this fragment. isn't it better if we do this name validation stuff in a viewmodel dedicated to this fragment only?
or, if we're supposed to share this viewmodel, why do we have 2 different shared viewmodels for a single flow then?

app/src/test/kotlin/com/waz/zclient/shared/user/name/ValidateNameUseCaseTest.kt
```diff
@@ -0,0 +1,60 @@
...
+    fun setup() {
+        validationNameUseCase = ValidateNameUseCase()
+    }
+
+    @Test
+    fun `Given run is executed, when name length is smaller or equal to 1, then return failure`() {
+        runBlockingTest {
+            val name = "t"
+            `when`(validateNameParams.name).thenReturn(name)
+            validationNameUseCase.run(validateNameParams).isLeft shouldBe true
```
gizemb, line 33: maybe we can also check that the failure is NameTooShort, not any other failure

app/src/main/kotlin/com/waz/zclient/feature/auth/registration/personal/email/CreatePersonalAccountNameFragment.kt
mejdoo: It's not a shared viewModel. It's a viewModel managing the hole flow of creating a personal account from the first to the last screen. We can create a single viewModel per fragment but we gonna add more complexity to the implementation.

gizemb: I don’t understand how it is able to manage the whole flow, if it is not a shared viewModel. You're creating it via `by viewModel()`, not `by sharedViewModel()`. If you're using it like this in every fragment, then you really have distinct instances of the same viewmodel type (that's my understanding of Koin), so no state or whatsoever is actually shared between fragments. And this shows me that the operations have no real dependencies between them, so they can be naturally isolated.

As an example, the operations that is recently added to that viewmodel does not really interest any of the other screens in the same flow, as far as I can see. So, I think it wouldn't complicate the flow to move it into a separate viewmodel. Actually, every operation would be isolated in its own class, thus, creating a better separation of concerns, supporting the single responsibility principle.

I don’t mean to stick my nose into your business, but this usage seemed so different than what we usually do. We usually have 1-1 relationship between fragments and view models, and this usually enough for our tasks (also keeps view models simple & small!). I just want to understand what’s different here in this flow.

mejdoo: I started  in the past refactoring to 1 fragment - 1 viewModel approach, but then I decided to keep as it is now. There are some dependencies in the flow: `createPersonalAccountWithEmailViewModel.sendActivationCode()` is used by both `CreatePersonalAccountEmailFragment` and `CreatePersonalAccountPinCodeFragment`. we may have some duplicated code in two viewModels.  

If we keep the current approach, the fragment `CreatePersonalAccountWithEmailViewModel` shouldn't be declared as `sharedViewModel() `because this will generate some bugs when we go back and forth through the registration flow. In fact, the fragments will observe some old success/error status. 

If you don't totally agree with this approach, I will refactor to 1-1 relationship between fragments and view models but we need to keep in mind that will have some duplicated code in two different view models. It's a trade off here.

gizemb: I think executing one usecase in different viewmodels is not much of a duplication. I'm in favor of separating those viewmodels, tbh. but if you're not comfortable with it maybe we can have a third opinion from someone else?

gizemb: we decided to discuss this later when the feature is finalized. I'll approve the PR.

PR #2793, opened 2020-04-24
Fix false-positives in integration tests

gizemb:
### Issues

The current way we make the assertions in integration tests do not fail the test when the assertion is false.

For example:

```
class MyTest : IntegrationTest() {

    @Test
    fun asd() {
        assert(3 == 5)
    }
}
```

such a test results in success currently.

### Causes

The assertion method was from AssertionsJVM.kt, which was not intended for test case assertions.

### Solutions

Replace all `assert()` calls from AssertionsJVM with respective `org.junit.Assert` calls.

### Testing

Since our CI pipeline cannot run integration tests yet, I ran the tests locally. I also tried to fail them with false assertions, and they failed as expected.

## Notes

The `GlobalDatabase24to25MigrationTest` fails, but I didn't include the fix here, since it's not this PR's responsibility.

Another assertion library can be used in the future. I tried assertJ, but had some problems compiling the project with it.

#### APK
[Download build #1969](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1969/artifact/build/artifact/wire-dev-PR2793-1969.apk)

### DIFF
--- a/storage/src/androidTest/kotlin/com/waz/zclient/storage/userdatabase/assets/AssetTables126to128MigrationTests.kt
+++ b/storage/src/androidTest/kotlin/com/waz/zclient/storage/userdatabase/assets/AssetTables126to128MigrationTests.kt
@@ -4,6 +4,8 @@ import com.waz.zclient.storage.db.users.migration.USER_DATABASE_MIGRATION_126_TO
 import com.waz.zclient.storage.db.users.migration.USER_DATABASE_MIGRATION_127_TO_128
 import com.waz.zclient.storage.userdatabase.UserDatabaseMigrationTest
 import kotlinx.coroutines.runBlocking
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
 import org.junit.Test
 
 class AssetTables126to128MigrationTests : UserDatabaseMigrationTest(126, 128) {
@@ -25,9 +27,9 @@ class AssetTables126to128MigrationTests : UserDatabaseMigrationTest(126, 128) {
 
         runBlocking {
             with(getV1Assets()[0]) {
-                assert(this.id == assetId)
-                assert(this.assetType == assetType)
-                assert(this.data == assetData)
+                assertEquals(this.id, assetId)
+                assertEquals(this.assetType, assetType)
+                assertEquals(this.data, assetData)
             }
         }
     }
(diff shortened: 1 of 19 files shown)


storage/src/androidTest/kotlin/com/waz/zclient/storage/userdatabase/assets/AssetTables126to128MigrationTests.kt
```diff
@@ -108,14 +110,14 @@ class AssetTables126to128MigrationTests : UserDatabaseMigrationTest(126, 128) {
...
-                assert(this.size == size)
-                assert(this.name == assetName)
-                assert(this.preview == preview)
-                assert(this.details == details)
-                assert(this.status == status)
+                assertEquals(this.id, assetId)
+                assertEquals(this.mime, mime)
+                assertEquals(this.downloaded, downloaded)
+                assertEquals(this.size, size)
+                assertEquals(this.name, name)
```
gizemb, line 117: there was a wrong assertion here but it wasn't failing.

storage/src/androidTest/kotlin/com/waz/zclient/storage/userdatabase/assets/AssetTables126to128MigrationTests.kt
```diff
@@ -163,23 +165,23 @@ class AssetTables126to128MigrationTests : UserDatabaseMigrationTest(126, 128) {
...
-                assert(this.preview == preview)
-                assert(this.details == details)
-                assert(this.uploadStatus == uploadStatus)
-                assert(this.isPublic == isPublic)
-                assert(this.uploaded == uploaded)
-                assert(this.retention == retention)
-                assert(this.assetId == assetId)
-                assert(this.encryptionSalt == null)
+                assertEquals(this.id, uploadAssetId)
+                assertEquals(this.source, source)
```
gizemb, line 169: here also

storage/src/androidTest/kotlin/com/waz/zclient/storage/userdatabase/assets/AssetTables126to128MigrationTests.kt
mejdoo: Should we replace also `assert()`  by `assertEquals()` in unit tests?

PR #2804, opened 2020-04-29
fix: An error is JSON deserialization in team/[tid]/members

makingthematrix:
The name for the "has more" field should be `hasMore`, not `has_more`.
This led to a crash when syncing team members.

#### APK
[Download build #2006](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2006/artifact/build/artifact/wire-dev-PR2804-2006.apk)

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/handler/TeamsSyncHandler.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/handler/TeamsSyncHandler.scala
@@ -60,7 +60,7 @@ class TeamsSyncHandlerImpl(userId:    UserId,
         data        <- flatten(client.getTeamData(id))
         membersData <- flatten(client.getTeamMembers(id))
         roles       <- flatten(client.getTeamRoles(id))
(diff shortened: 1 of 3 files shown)


wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/handler/TeamsSyncHandler.scala
```diff
@@ -60,7 +60,7 @@ class TeamsSyncHandlerImpl(userId:    UserId,
         data        <- flatten(client.getTeamData(id))
         membersData <- flatten(client.getTeamMembers(id))
         roles       <- flatten(client.getTeamRoles(id))
-        members     =  if (membersData.has_more) Seq.empty[TeamMember] else membersData.members
+        members     =  if (membersData.hasMore) Seq.empty[TeamMember] else membersData.members
```
BradleyWire, line 63: Do we change the test case values too?

makingthematrix: Right. Good catch. One more small change in the unit test coming.

PR #2059, opened 2019-03-21
Revise onboarding copy to clarify audiences

infotexture:
## What's new in this PR?

Revised copy pr 2019-03-21 discussion w/ JG, SY & RH.

#### APK
[Download build #12484](http://192.168.120.33:8080/job/Pull%20Request%20Builder/12484/artifact/build/artifact/wire-dev-PR2059-12484.apk)

PR #2813, opened 2020-04-30
Feat: Logout functionality in Kotlin - Part 4

gizemb:
## What's new in this PR?

Jira issue: https://wearezeta.atlassian.net/browse/AN-6797

### Issues

This part handles navigation after a user logs out.

When user (User1) logs out,
- if there are no users left, direct to Login Screen
- if there is another user (User2), direct to Conversation List screen for User2

### Solutions

Since we cannot directly call Scala classes from Kotlin classes, I had to make a trick and trigger an implicit Intent. This intent will be resolved by the operating system.

### Testing

Manually tested and unit tests are also added. Whether the correct intent is resolved can also be detected by QA's integration tests.
#### APK
[Download build #2058](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2058/artifact/build/artifact/wire-dev-PR2813-2058.apk)
[Download build #2059](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2059/artifact/build/artifact/wire-dev-PR2813-2059.apk)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -153,6 +153,11 @@
                 <category android:name="android.intent.category.VIEW" />
             </intent-filter>
 
+            <intent-filter>
+                <action android:name="com.wire.ACTION_CURRENT_USER_CHANGED"/>
+                <category android:name="android.intent.category.DEFAULT"/>
+            </intent-filter>
+
         </activity>
 
         <activity
@@ -213,8 +218,12 @@
             android:launchMode="singleTask"
             android:theme="@style/Theme.Dark"
             android:windowSoftInputMode="stateHidden|adjustResize"
-            android:exported="false"
-            />
+            android:exported="false">
+            <intent-filter>
+                <action android:name="com.wire.ACTION_NO_USER_LEFT"/>
+                <category android:name="android.intent.category.DEFAULT"/>
+            </intent-filter>
+        </activity>
 
         <activity
             android:name=".ForceUpdateActivity"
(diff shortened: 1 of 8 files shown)


app/src/main/kotlin/com/waz/zclient/feature/settings/account/logout/LogoutUseCase.kt
```diff
@@ -25,21 +25,32 @@ class LogoutUseCase(
...
+    private suspend fun updateCurrentUserId(loggedOutUserId: String): Either<Failure, LogoutStatus> =
         accountsRepository.activeAccounts().fold({
             clearCurrentUserId()
-            Either.Right(Unit) //TODO should we log failure somehow?
+            Either.Right(CouldNotReadRemainingAccounts)
         }) {
             val remainingAccountId = it.firstOrNull { it.id != loggedOutUserId }?.id
-            remainingAccountId?.let { usersRepository.setCurrentUserId(it) } ?: clearCurrentUserId()
-            Either.Right(Unit)
+            val status = if (remainingAccountId == null) {
```
BradleyWire, line 40: Minor but we could use `isNull` here

app/src/main/AndroidManifest.xml
```diff
@@ -153,6 +153,11 @@
                 <category android:name="android.intent.category.VIEW" />
             </intent-filter>
 
+            <intent-filter>
```
mejdoo, line 156: I think you are missing `android:exported="false"`for `MainActivity`

gizemb:
build again

app/src/main/AndroidManifest.xml
gizemb: it wasn't there before either. I didn't put it on purpose because it would block the deep link in the first intent-filter. I didn't see a big risk in exposing this intent filter for now.

mejdoo: ok makes sense

gizemb:
build again

app/src/main/kotlin/com/waz/zclient/feature/settings/account/logout/LogoutUseCase.kt
gizemb: there's no resolved `isNull()` method. There is` isNullOrEmpty()` or `isNullOrBlank()` but I don't want the "or" in them

PR #2816, opened 2020-04-30
Create personal account - part 14

mejdoo:
## What's new in this PR?

Follow up of #2801 

- Handle no connection status
- Show different errors in AlertDialogs

https://wearezeta.atlassian.net/browse/AN-6888
#### APK
[Download build #2136](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2136/artifact/build/artifact/wire-dev-PR2816-2136.apk)


app/src/main/kotlin/com/waz/zclient/core/ui/dialog/Alert.kt
```diff
@@ -0,0 +1,24 @@
+package com.waz.zclient.core.ui.dialog
+
+import android.content.Context
+import androidx.appcompat.app.AlertDialog
+import com.waz.zclient.R
+
+object Alert {
+    fun showError(context: Context, message: String) {
```
BradleyWire, line 8: I'm dubious about making this super generic, what if we have a different name for the action button? or a different action for when the button is clicked? and if we need two buttons for the error dialog in some scenarios? We may end up with loads of external dialogs doing slightly different things. 

I think it might be good to abstract the `AlertDialog` logic into maybe our own Builder/Creational pattern so we can make it dynamic for the dialog we need without the need for repeating ourselves and creating the same `AlertDialog` over and over again or small variants of it.

So let's say we have a generic error dialog, it doesn't really need much to begin with:

```
WireDialog.Builder(context)
    .type(WireDialog.GENERIC_ERROR)
    .create() 
```
So now we have a generic one by defining a type, we can also add small tweaks: 

An example could be: 

```
WireDialog.Builder(context)
    .type(WireDialog.GENERIC_ERROR)
    .positiveButton(R.string.thanks) 
    .positiveAction(...)
    .create() 
```
By overriding the characteristics we want to change using our own implementation. 

This is just from the top of my head whilst writing this. I'm unsure of the implementation details as of yet.

BradleyWire, line 8: I'm dubious about making this super generic, what if we have a different name for the action button? or a different action for when the button is clicked? and if we need two buttons for the error dialog in some scenarios? We may end up with loads of external dialogs doing slightly different things. 

I think it might be good to abstract the `AlertDialog` logic into maybe our own Builder/Creational pattern so we can make it dynamic for the dialog we need without the need for repeating ourselves and creating the same `AlertDialog` over and over again or small variants of it.

So let's say we have a generic error dialog, it doesn't really need much to begin with:

```
WireDialog.Builder(context)
    .type(WireDialog.GENERIC_ERROR)
    .create() 
```
So now we have a generic one by defining a type, we can also add small tweaks: 

An example could be: 

```
WireDialog.Builder(context)
    .type(WireDialog.GENERIC_ERROR)
    .positiveButton(R.string.thanks) 
    .positiveAction(...)
    .create() 
```
By overriding the characteristics we want to change using our own implementation. 

This is just from the top of my head whilst writing this. I'm unsure of the implementation details as of yet.

app/src/test/kotlin/com/waz/zclient/feature/auth/registration/personal/CreatePersonalAccountWithEmailViewModelTest.kt
```diff
@@ -128,6 +129,18 @@ class CreatePersonalAccountWithEmailViewModelTest : UnitTest() {
             }
         }
 
+    @Test
+    fun `given sendActivationCode is called, when there is a network connection error then the activation code is not sent`() =
```
mejdoo, line 133: @android10 I added the test cases but the tests here are false positives. I need to update them as soon as we merge Gizem's PR #2784

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/WireDialog.kt
```diff
@@ -0,0 +1,78 @@
...
+            return this
+        }
+
+        fun negativeButton(text: String, listener: ((DialogInterface, Int) -> Unit)): Builder {
+            dialogBuilder.setNegativeButton(text, listener)
+            return this
+        }
+
+        fun show() {
+            dialogBuilder.create().show()
```
BradleyWire, line 70: if we use the Builder pattern, it'd be good to build something here, maybe build up the dialog or create an instance of the `WireDialog` with a `Builder` pass a param? I'd also wonder whether we should limit this functionality to `show()` and not `create() ` as we may need to access the `AlertDialog` public functionality outside of this builder.

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/WireDialog.kt
```diff
@@ -0,0 +1,78 @@
...
+
+class WireDialog {
+
+    class Builder(context: Context) {
+        private val dialogBuilder = AlertDialog.Builder(context)
+
+        fun type(value: Int): Builder {
+
+            when (value) {
+                GENERIC_ERROR -> {
```
BradleyWire, line 17: This seems like it'll become pretty big with the more types of dialogs we have, maybe we could use a `Factory` like we had previously to return an `AlertDialog.Builder` with predetermined characteristics? and then a regular builder with no conceived characteristics when no type is given here?

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/WireDialog.kt
```diff
@@ -0,0 +1,78 @@
...
+            return this
+        }
+
+        fun show() {
+            dialogBuilder.create().show()
+        }
+    }
+
+    companion object {
+        const val GENERIC_ERROR = 0
```
BradleyWire, line 75: An enum could be useful here

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/WireDialog.kt
BradleyWire: I would also add some documentation to the style guide for these changes.

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/WireDialog.kt
```diff
@@ -0,0 +1,78 @@
+package com.waz.zclient.core.ui.dialog
+
+import android.content.Context
+import android.content.DialogInterface
+import androidx.annotation.StringRes
+import androidx.appcompat.app.AlertDialog
+import com.waz.zclient.R
+
+class WireDialog {
```
gizemb, line 9: I'm not very confident about this implementation. It's basically wrapping the entire AlertDialog.Builder api just to add some convenience methods. I think it's ok to use AlertDialog.Builder directly. If we go this way, we need to write wrappers for every method that we want to use from AlertDialog api. Since AlertDialog is unlikely to change and pretty stable, I think that's not necessary.

I think instead we can create a different scope called DialogOwner or sth, and add delegate these "building" responsibilities there. I came up with sth like this quickly. Didn't try it out though:

```
//sealed classes instead of enums
sealed class DialogType(val block: AlertDialog.Builder.() -> Unit)
internal class GenericError(errorMessage: String) : DialogType({
    setTitle(errorMessage)
    positiveButton("OK")
})

//the real deal
interface DialogOwner {
    companion object {
        val NO_OP_LISTENER = DialogInterface.OnClickListener { _, _ -> /*no-op*/ }
    }

    fun showDialog(context: Context, block: AlertDialog.Builder.() -> Unit) {
        AlertDialog.Builder(context).apply(block).create().show()
    }

    fun showDialog(context: Context, type: DialogType) {
        AlertDialog.Builder(context).apply(type.block).create().show()
    }

    //we can also omit this, since it'll clutter this class as we add more 'DialogType's (pretty much what Bradley said). It's pretty convenient though...
    fun showErrorDialog(context: Context, errorMessage: String) =
        showDialog(context, GenericError(errorMessage).block)
}

//Some extensions (optional)
fun AlertDialog.Builder.positiveButton(text: String) = setPositiveButton(text, DialogOwner.NO_OP_LISTENER)
```

And in Fragment:

```
class MyFragment : DialogOwner {
    fun onError() = showErrorDialog(requireContext(), "ERROR!!!")
    
    //OR like this
    fun onError() = showDialog(requireContext(), GenericError("ERROR!!!"))
    
    //You can also DIY
    fun onError() = showDialog(requireContext()) {
        title("ERROR!!")
        positiveButton("OK")
    }
}
```

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/WireDialog.kt
```diff
@@ -0,0 +1,78 @@
...
+        fun type(value: Int): Builder {
+
+            when (value) {
+                GENERIC_ERROR -> {
+                    positiveButton(android.R.string.ok) { dialog, _ -> dialog.dismiss() }
+                }
+                NETWORK_CONNECTION_ERROR -> {
+                    title(R.string.no_internet_connection_title)
+                    message(R.string.no_internet_connection_message)
+                    positiveButton(android.R.string.ok) { dialog, _ -> dialog.dismiss() }
```
gizemb, line 23: As far as I remember, DialogInterface already dismisses the dialog if you don't do anything here. You don't need to call dialog.dismiss()

I remember this because I once struggled to make one that doesn't dismiss the dialog automatically :D

app/src/main/kotlin/com/waz/zclient/feature/auth/registration/personal/email/CreatePersonalAccountWithEmailViewModel.kt
```diff
@@ -83,11 +86,13 @@ class CreatePersonalAccountWithEmailViewModel(
     }
 
     private fun sendActivationCodeFailure(failure: Failure) {
-        when (failure) {
-            is EmailBlacklisted -> _sendActivationCodeErrorLiveData.value =
-                ErrorMessage(R.string.create_personal_account_with_email_email_blacklisted_error)
-            is EmailInUse -> _sendActivationCodeErrorLiveData.value =
-                ErrorMessage(R.string.create_personal_account_with_email_email_in_use_error)
+        if (!isNetworkConnectionFailure(failure)) {
```
gizemb, line 89: I think we should still show a generic error for other failures

app/src/test/kotlin/com/waz/zclient/feature/auth/registration/personal/CreatePersonalAccountWithEmailViewModelTest.kt
gizemb: will do that today 👍

gizemb, line 133: we can't really conclude that the "activation is not sent" from this test right?
maybe we can change the naming, or add a verification like `verifyNoInteractions(successLiveData)`

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/WireDialog.kt
mejdoo: I think we can bring this discussion out of this PR and everyone is welcome to provide a solution in a separate PR  (like you did @gizemb for false positive tests). There are different approaches to handle the problem. For the scope of this PR, I'm gonna use directly `AlertDialog.Builder` to show the dialogs. Once we agree on something, I'm happy to refactor later.

BradleyWire: i'm happy for you to investigate this solution over mine. We can't really merge this PR without an accepted solution anyway. But, this ticks all the boxes and feels much more functional than the one I proposed.

gizemb: Yes, I'm OK with moving this dialog stuff into another PR with its tests and so on..
We can switch back to native AlertDialog.Builder for the scope of this PR i guess.

app/src/main/kotlin/com/waz/zclient/core/ui/dialog/WireDialog.kt
mejdoo: good catch 👍

PR #2829, opened 2020-05-06
chore: Remove unused "last accent color" preference

makingthematrix:
For a long time now we keep the accent color used by the current user in the storage.
In the past we used a preference for this and a bit of the code was left in the codebase:
the preference was set, but never read. I removed that code.


makingthematrix:
build again

PR #2833, opened 2020-05-09
fix: Too many prekeys requests

makingthematrix:
With the Large Teams feature we changed the way we specify to what users should a message be sent. That led to error responses from the backend in some cases which in consequence trigger new prekeys requests. There are way too many prekeys requests coming from Android devices to the backend right now - basically at least one for every message received by a device with the new version if the user also has Wire on another device.

The message types which are affected are: last reads, delivery confirmations, and read receipts. In standard cases we want to send the message to all users in the conversation - and that works well - but in these three cases (*) we want them to be sent only to other self devices and to devices of the sender of the original message. So, we put those user ids on the list of
explicit recipients and later we put that list in the `report_missing` field of the OTR message. And here's something unclear starts to happen. In those three cases, sending the message to the conversation (`/conversations/[conv id]/otr/messages` endpoint) results in the error 412, Clients Mismatch with the data that a client is missing. In consequence, the Android app requests new prekeys for that client.

Before Large Teams, the error didn't happen, because the list of recipients wasn't put in the `report_missing` field at all (which means that the message is send to everyone in the conversation). So, it looks like the error happens when `report_missing` specifies the list of recipients, but the backend decides for some reason that the list is wrong.

**EDIT**:
It turned out the real reason was that Android uses an obsolete version of backend-api-proto. The new version was compiled and put on Bintray, and the version number in Android was bumped. In the new version we have the reportMissing field in OtrNewMessage. Our report_missing from OtrMessage has to be copied there.

I also made some indentation and other small changes to improve readability. It helps me think.

*) There are also button confirmations - but they go only ot the poll's owner, not to the other self devices, and there are ephemeral expirations, but they seemed to work well.

#### APK
[Download build #2071](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2071/artifact/build/artifact/wire-dev-PR2833-2071.apk)
[Download build #2072](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2072/artifact/build/artifact/wire-dev-PR2833-2072.apk)
[Download build #2075](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2075/artifact/build/artifact/wire-dev-PR2833-2075.apk)

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/handler/LastReadSyncHandler.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/handler/LastReadSyncHandler.scala
@@ -38,19 +38,16 @@ class LastReadSyncHandler(selfUserId: UserId,
 
   import com.waz.threading.Threading.Implicits.Background
 
-  def postLastRead(convId: ConvId, time: RemoteInstant): Future[SyncResult] = {
-    verbose(l"postLastRead($convId, $time)")
-
+  def postLastRead(convId: ConvId, time: RemoteInstant): Future[SyncResult] =
     convs.get(convId).flatMap {
       case Some(conv) if conv.lastRead.isAfter(time) => // no need to send this msg as lastRead was already advanced
         Future.successful(Success)
       case Some(conv) =>
         val msg = GenericMessage(Uid(), LastRead(conv.remoteId, time))
         otrSync
-          .postOtrMessage(ConvId(selfUserId.str), msg)
+          .postOtrMessage(ConvId(selfUserId.str), msg, recipients = Some(Set(selfUserId)))
           .map(SyncResult(_))
       case None =>
         Future.successful(Failure(s"No conversation found for id: $convId"))
     }
-  }
 }
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/handler/MessagesSyncHandler.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/handler/MessagesSyncHandler.scala
@@ -97,14 +97,14 @@ class MessagesSyncHandler(selfUserId: UserId,
   def postReceipt(convId: ConvId, msgs: Seq[MessageId], userId: UserId, tpe: ReceiptType): Future[SyncResult] =
     convs.convById(convId).flatMap {
       case Some(conv) =>
-        val msg = tpe match {
-          case ReceiptType.Delivery         => GenericMessage(msgs.head.uid, Proto.DeliveryReceipt(msgs))(Proto.DeliveryReceipt)
-          case ReceiptType.Read             => GenericMessage(msgs.head.uid, Proto.ReadReceipt(msgs))(Proto.ReadReceipt)
-          case ReceiptType.EphemeralExpired => GenericMessage(msgs.head.uid, Proto.MsgRecall(msgs.head))
+        val (msg, recipients) = tpe match {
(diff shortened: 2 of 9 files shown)


wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/handler/LastReadSyncHandler.scala
```diff
@@ -38,19 +38,16 @@ class LastReadSyncHandler(selfUserId: UserId,
...
-
+  def postLastRead(convId: ConvId, time: RemoteInstant): Future[SyncResult] =
     convs.get(convId).flatMap {
       case Some(conv) if conv.lastRead.isAfter(time) => // no need to send this msg as lastRead was already advanced
         Future.successful(Success)
       case Some(conv) =>
         val msg = GenericMessage(Uid(), LastRead(conv.remoteId, time))
         otrSync
-          .postOtrMessage(ConvId(selfUserId.str), msg)
+          .postOtrMessage(ConvId(selfUserId.str), msg, recipients = Some(Set(selfUserId)))
```
makingthematrix, line 48: last_read recipients list set to the self user.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/handler/MessagesSyncHandler.scala
```diff
@@ -95,16 +95,16 @@ class MessagesSyncHandler(selfUserId: UserId,
...
-          case ReceiptType.EphemeralExpired => (GenericMessage(msgs.head.uid, Proto.MsgRecall(msgs.head)), Set(selfUserId, userId))
+        val msg = tpe match {
+          case ReceiptType.Delivery         => GenericMessage(msgs.head.uid, Proto.DeliveryReceipt(msgs))(Proto.DeliveryReceipt)
+          case ReceiptType.Read             => GenericMessage(msgs.head.uid, Proto.ReadReceipt(msgs))(Proto.ReadReceipt)
+          case ReceiptType.EphemeralExpired => GenericMessage(msgs.head.uid, Proto.MsgRecall(msgs.head))
         }
 
         otrSync
-          .postOtrMessage(conv.id, msg, Some(recipients), nativePush = false)
+          .postOtrMessage(conv.id, msg, recipients = Some(Set(selfUserId, userId)), nativePush = false)
```
makingthematrix, line 107: delivery confirmations and read receipts recipients set to the self user and the sender of the original message (was only the original sender before). Ephemeral expirations stay the same.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/sync/otr/OtrSyncHandler.scala
```diff
@@ -93,7 +93,7 @@ class OtrSyncHandlerImpl(teamId:             Option[TeamId],
         resp       <- if (content.estimatedSize < MaxContentSize)
                         msgClient.postMessage(
                           conv.remoteId,
-                          OtrMessage(selfClientId, content, external, nativePush, recipients),
+                          OtrMessage(selfClientId, content, external, nativePush),
```
makingthematrix, line 96: The list of recipients is **not** set as the `report_missing` field, just as it was before Large Teams.
The list of recipients is still used to encrypt messages.
The list of recipients is used as `report_missing` for broadcasting availability status changes.

android10:
Maybe I'm missing something but there are requests involved here right? Does this affect the networking framework that @gizemb and I wrote for the migration?

makingthematrix:
@android10 :
It's not connected to that. The problem arose from changes made for the Large Teams feature. The main code is in `OtrSyncHandler` in the `postOtrMessage` method. We get an error: `MessageResponse.Failure(ClientMismatch(_, missing, _, _))` and in the response we ask for prekeys for all the `missing` clients.

PR #2845, opened 2020-05-19
fix: Re-add hanfling of team.member-join events

makingthematrix:
fixes https://wearezeta.atlassian.net/browse/AN-6896

At one point in the Large Teams feature implementation we removed the code for handling incoming events of new members joining the team. Later design changes reverted that decision but the code was not reintroduced, resulting in that, for example, new team members are initially invisible in SearchUI.

This PR reintroduces a few lines which sync new team members if the event comes.

#### APK
[Download build #2108](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2108/artifact/build/artifact/wire-dev-PR2845-2108.apk)

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
@@ -92,12 +92,16 @@ class TeamsServiceImpl(selfUser:           UserId,
     verbose(l"Handling events: $events")
     import TeamEvent._
 
+    val membersJoined  = events.collect { case MemberJoin(_, u)    => u }.toSet
     val membersLeft    = events.collect { case MemberLeave(_, u)   => u }.toSet
     val membersUpdated = events.collect { case MemberUpdate(_, u)  => u }.toSet
 
     for {
-      _ <- RichFuture.traverseSequential(events.collect { case e:Update => e}) { case Update(id, name, icon) => onTeamUpdated(id, name, icon) }
-      _ <- onMembersLeft(membersLeft)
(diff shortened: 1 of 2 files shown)


wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
```diff
@@ -203,6 +207,15 @@ class TeamsServiceImpl(selfUser:           UserId,
     )
   }
 
+  private def onMembersJoined(members: Set[UserId]) = {
+    verbose(l"onMembersJoined: members: $members")
+    for {
+      _ <- sync.syncUsers(members).flatMap(syncRequestService.await)
+      _ <- sync.syncTeam().flatMap(syncRequestService.await)
+      _ <- userStorage.updateAll2(members, _.copy(teamId = teamId, deleted = false))
```
mejdoo, line 215: Why we have two functions for the same functionality: `updateAll` and `updateAll2` ? . I saw in `CachedStorageImpl` that `updateAll` is calling `updateAll2`.

makingthematrix: @mejdoo : `updateAll` is very generic: it takes a map of ids of elements to functions, and those functions can be different for every id. In most cases we don't need that - it's enough to have one and the same function doing the update on all elements. `updateAll2` does that. I guess someone who wrote it (long time ago) first came up with the more generic version and then made the other one to simplify usage in some places. Now in most Scala code we use the second one.

mejdoo: I would say you don't have to refactor that now. Maybe you can change function names later to something making more sense than `updateAll`/ `updateAll2`

Issue #2128, opened 2019-05-10
Sharing external content crashes app


kroerig:
**Describe the bug**
Sharing an external content i.e. browser URL crashes Wire.

**To Reproduce**
1. Fire up your browser (i.e Chrome),
2. open an URL
3. choose 'share' from the menu.
4. select Wire

=> Android reports "This app keeps crashing"

**Expected behavior**
Wire opens an presents the contact list.

**Screenshots**
![Screenshot_20190510-102446](https://user-images.githubusercontent.com/5148795/57513218-0610a000-730e-11e9-9113-98bed573263d.png)

**Smartphone (please complete the following information):**
- Wire version that exhibits the issue: 3.32.781
- Last Wire version that did not exhibit the issue (if applicable): N/A
- Android OS version used to run Wire: 9
- Is this a custom firmware or a stock one: Sony stock
- Mobile phone model/manufacturer: Sony Xperia ZX2 compact
- Mobile network type (EDGE/LTE/Wi-Fi/Offline): any

**Additional context**
Workaround: 
Fire up Wire, switch back to the browser, share your content.

madranet:
Ah!  I'm glad it's not just me then. I've just got a new phone and thought I'd accidentally broken Wire while setting it up, as it had been working fine on my previous phone. But this has been happening to me, every time I try and share a photo to Wire from within the Google Photos app. 



---
* Wire version that exhibits the issue: 3,32,781
* Last Wire version that did not exhibit the issue (if applicable): ???
* Android OS version used to run Wire: 8,0,1
* Is this a custom firmware or a stock one: Rooted but otherwise stock Samsung TouchWiz
* Mobile phone model/manufacturer: Samsung Galaxy S7
* Mobile network type (EDGE/LTE/Wi-Fi/Offline): WiFi [not tested on mobile]
---

kroerig:
Does the workaround work for you?

3.32.781 is a public version. First thought it could be a beta-problem only.

madranet:
Since I've had the problem when attaching images from Google Photos, my workaround has been to just use the attachments function instead:

![Screenshot_20190510-160519](https://user-images.githubusercontent.com/669658/57537441-2e17f780-733e-11e9-873c-f42d21db05ea.jpg)

IronJunkie:
Suspect this issue may be related to #2127 where notifications are not received when app is a cached background process?

Can confirm issue present with
    Wire version that exhibits the issue: 3.32.781
[...]

Issue #2121, opened 2019-05-06
Cannot see reply when one message used


Kreloc:
**Describe the bug**

The timestamp for read receipt came through at one time and then when I opened the Wire app next for one text message in a one to one conversation, it had the timestamp for the current time as I opened it as the sent time. Now when this message is selected and reply is used with it, neither of us can see the reply and get a message stating 'You cannot see this image'

The other person sent it first, then did a different message that did work and then when I tried to send my own message as a reply, they got the same message as above where the reply would normally show up.

**To Reproduce**
After I sent that message, I did immediately close the wire app after sending it.

1. Go to 'a chat with another person that has read receipts enabled'
2. Click on 'text input field, type message. immediately close app after sending'
3. Have other participant read message before next step.
4. Open app later and see if time stamps are mixed
5. Select message and try to use reply.
6. See if get error message

**Expected behavior**
That the timestamps would be correct and that the reply message would be visible.

**Smartphone (please complete the following information):**
- Wire version that exhibits the issue: 3.31.777
- Last Wire version that did not exhibit the issue (if applicable): NA
- Android OS version used to run Wire: 8.0.0
- Is this a custom firmware or a stock one: stock
- Mobile phone model/manufacturer: Samsung Galaxy S8 Activ
- Mobile network type (EDGE/LTE/Wi-Fi/Offline): LTE

**Additional context**
Noticed the mixed up timestamps and brought it to the other person's attention, then we noticed that this broke being able to use that message in a reply. Thought you all might like to know that it happened.

Thanks!

marcoconti83:
Hi @Kreloc ,

thanks for opening the issue. I'm interested in the issue with the reply resulting in an error message, any chance you can send us a screenshot of the issue (with the text blurred, if sensitive)?

marcoconti83:
Might be related to https://github.com/wireapp/wire-android/issues/2129

Kreloc:
Hi @marcoconti83 ,

Please see the two attached screenshots. One shows the message itself with the incorrect timestamps and the other shows that message being used in a reply with the error message. Thank you for looking into this!

![Wire 2019-07-09 at 9_32 AM](https://user-images.githubusercontent.com/11083519/60910330-fc7fb780-a22c-11e9-968c-807eae30847d.jpeg)
[...]

PR #2852, opened 2020-05-20
Fix flaky unit tests & polish some

gizemb:
### Issues

Part 2 of #2784 

This PR removes lenient() calls, introduces LiveData.awaitValue() for awaiting the LiveData change, adds InstantTaskExecutorRule for issues with main methods and android framework, and polishes up some code.

It also bumps coroutines version to 1.3.7.

### Causes & Solutions

**lenient():** This shouldn't be needed at all in a robust test case, as it indicates that we mock something that we actually don't need 

**LiveData.awaitValue():** We often try to observe a change in the value of the liveData by observeOnce{}, or inside flow.collect{}, but those observer methods are sometimes executed too early and gives us the previous value of the liveData. This method suspends the coroutine and doesn't execute anything before we get the result of liveData.

**InstantTaskExecutorRule:** mutableLiveData.postValue() method was causing problems in test environment:
> Suppressed: java.lang.UnsatisfiedLinkError: android.util.Log.println_native(IILjava/lang/String;Ljava/lang/String;)I
> (called from postValue())

I copied the solution from here: https://stackoverflow.com/questions/45988310/setvalue-and-postvalue-on-mutablelivedata-in-unittest

### Testing

Run ./gradlew testDevDebugUnitTest. 
I ran the tests ~10 times in a row on my locale, and didn't see any flakiness or errors.

## Notes

I couldn't fix the errors in one class, because the requirements in the test cases were a lot different than the production code. I couldn't make sure which one is the source of truth. Those tests are ignored for now, and I opened an issue to fix either the tests, or the class itself: https://wearezeta.atlassian.net/browse/AN-6906

#### APK
[Download build #2121](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2121/artifact/build/artifact/wire-dev-PR2852-2121.apk)
[Download build #2128](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2128/artifact/build/artifact/wire-dev-PR2852-2128.apk)

### DIFF
--- a/app/src/main/kotlin/com/waz/zclient/feature/settings/devices/detail/SettingsDeviceDetailViewModel.kt
+++ b/app/src/main/kotlin/com/waz/zclient/feature/settings/devices/detail/SettingsDeviceDetailViewModel.kt
@@ -1,6 +1,5 @@
 package com.waz.zclient.feature.settings.devices.detail
 
-import android.util.Log
 import androidx.lifecycle.LiveData
 import androidx.lifecycle.MutableLiveData
 import androidx.lifecycle.ViewModel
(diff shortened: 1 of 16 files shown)


gizemb:
build again

PR #2193, opened 2019-06-24
fix: It shouldn't be possible to create multiple group conversations 

makingthematrix:
fixes: https://wearezeta.atlassian.net/browse/AN-6235

### What's new in this PR?

A fix preventing the user from creating multiple conversations by clicking the "create" button many times quickly.

### Issues

It is technically possible to create many conversations with the same participants: fake 1-to-1, and real group convs. This possibilty leads to a bug when the user clicks the "Create conversation" button many times quick enough that if the app experiences a delay it is possible to create many conversations at once. 

### Solutions

To prevent that, clicks on all the buttons used for creating conversation - in the "Create group" UI, "Create guestroom", and fake 1-to-1 conversations with other team members, are now ignored while the creation of the conversation is not finished after the first click.

### Testing

The reliable way to test it requires introduction of an artificial delay in SE, in `ConversationsUiService.getOrCreateOneToOneConversation`. Without it, in most cases the first click is processed quick enough to prevent the bug. 
There are three "buttons" in the UI which are affected:
* If you're a team member, go to SearchUI and look for an *uncontacted* team member. Clicking on its name should immediately open the conversation with that person without any confirmation screen. If you click on the name many times fast, it might result in creation of many fake 1-to-1 conversations.
* SearchUI, the "Create guestroom" button. Clicking on it quickly many times may create many guestrooms.
* SearchUI, the "Create group" button. It takes the user to the Create Group UI. Provide the name of the group and choose participants. Then try to click "Done" many times quickly.

PR #2878, opened 2020-06-08
fix: Race condition on conversation list click while reloading the list

makingthematrix:
fixes a bug reported in Google Play Console
https://play.google.com/apps/publish/?account=7098984309886892484#AndroidMetricsErrorsPlace:p=com.wire&appid=4973241010395499500&appVersion=PRODUCTION&clusterName=apps/com.wire/clusters/9e255feb&detailsAppVersion=PRODUCTION&detailsSpan=7

We use a mutable list for items and in rare cases it may lead to a situation when a click is executed while the list is refreshing. Rare, but it results in a crash. The ideal solution to this would be to change the items into an immutable list (but held in a `var`
instead of `val`) but that messes up with how we put items in folders and I don't want to spend time on this right now. Instead, I wrapped the `onClick` methods in `Try`. This way the worst what will happen is that the user will have to click again.
#### APK
[Download build #2193](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2193/artifact/build/artifact/wire-dev-PR2878-2193.apk)


makingthematrix:
build again

PR #2201, opened 2019-07-01
fix: Show audio and video assets in the collection view

makingthematrix:
## What's new in this PR?

Audio and video assets are now displayed in the collection view as files.
fixes: https://wearezeta.atlassian.net/browse/AN-6282

### Issues

Audio and video assets have their own types, but the collection view tried to treat them as if their type was the same as for all other files. In result, they were filtered out.

### Solutions

In every place the collection view queries for `ANY_ASSET` I extended it to "one of `ANY_ASSET`, `AUDIO_ASSET`, or `VIDEO_ASSET`".

### Testing

1. Make an audio or video recording in a conversation.
2. Go to the collection (the "magnifying lens" icon).
3. You should see the recording under "files".

## Notes

I made the fix with the least amount of changes to the collection code. Apart from a few minor things, the code was not touched since January 2017. I think it's a good candidate for rewriting in Kotlin.

PR #2880, opened 2020-06-08
fix: The "Done" button on Create Group Conversation screen unresponsive for large teams

makingthematrix:
fixes https://wearezeta.atlassian.net/browse/AN-6912 (?)

It's more of a performance improvement than a fix.

I wasn't able to reproduce this behaviour but from reading the code I guess the performance problem might lie in that we try to "track" how many of added users are guests. This information is then ignored because we don't use a tracking system anymore. I removed it from the code.

That being said, creating a new conversation is a complex operation and if the device is already under heavy load because of processing a big number of team members, it can take some considerable time.

#### APK
[Download build #2201](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2201/artifact/build/artifact/wire-dev-PR2880-2201.apk)


makingthematrix:
build again

PR #2897, opened 2020-06-26
fix: Remove file restrictions on file pickers

makingthematrix:
When we open a file picker from inside Wire, and try to force it to hide some of the files basing on their extensions,
some file pickers end up restricking access to whole folders (e.g. Huawei phones, Android 8.0). Until we find out
how to handle it better, we decided not to restrict the file picker. The user will be allowed to choose any file,
and only then Wire will prevent the file from being sent. The code that adds the list of allowed MIME types to the intent
will stay commented out in case we change this decision.

I also moved the dialog with the error message a bit to make sure it is displayed (there was a report that it doesn't
appear on one device). And I fixed a bug when the error system message was displayed with empty text when the extension is unknown (rare, but possible).

fixes https://wearezeta.atlassian.net/browse/AN-6941
#### APK
[Download build #2263](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2263/artifact/build/artifact/wire-dev-PR2897-2263.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/pages/main/conversation/AssetIntentsManager.java
+++ b/app/src/main/java/com/waz/zclient/pages/main/conversation/AssetIntentsManager.java
@@ -26,12 +26,10 @@ import android.content.Intent;
 import android.content.pm.PackageManager;
 import android.os.Build;
 import android.provider.MediaStore;
-import android.webkit.MimeTypeMap;
 
 import androidx.annotation.Nullable;
 import androidx.core.content.FileProvider;
 
-import com.waz.service.ZMessaging;
 import com.waz.utils.IoUtils;
 import com.waz.utils.wrappers.AndroidURI;
 import com.waz.utils.wrappers.AndroidURIUtil;
@@ -39,7 +37,6 @@ import com.waz.utils.wrappers.URI;
 import com.waz.zclient.BuildConfig;
 import com.waz.zclient.Intents;
 import com.waz.zclient.core.logging.Logger;
-import com.waz.service.assets.FileRestrictionList;
 
 import java.io.File;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/waz/zclient/pages/main/conversation/AssetIntentsManager.java
```diff
@@ -106,7 +106,10 @@ private void openDocument(Set<String> mimeTypes, IntentType tpe, boolean allowMu
 
     public void openFileSharing() {
         final Set<String> mimeTypes = new HashSet<>();
-        FileRestrictionList fileRestrictions = ZMessaging.currentGlobal().fileRestrictionList();
+        // FIXME: For now we let the user choose from all files and we check restrictions before sending
+        // The reason is, restricting access to files via intent make some file pickers restrict also
+        // access to certain folders, while other file pickers ignore the restrictions altogether.
+        /*FileRestrictionList fileRestrictions = ZMessaging.currentGlobal().fileRestrictionList();
         if (!fileRestrictions.getEnabled()) {
```
mejdoo, line 113: Maybe we can remove this dead code, if we will keep this behaviour for a long time.

makingthematrix: @mejdoo : I'd like to keep it like that for now. It's possible that this decision will be overturn.

makingthematrix:
build again

app/src/main/java/com/waz/zclient/pages/main/conversation/AssetIntentsManager.java
mejdoo: ok

PR #2896, opened 2020-06-24
Make request to SFT server (conference part 2)

johnxnguyen:
## What's new in this PR?

With this PR the app is able to contact the SFT server to connect to a conference call.

When starting or joining a conference call, AVS will request the client to make a request to the SFT server on their behalf. They do this by invoking the `SFTRequestHandler`, passing the url of the server and a payload to post. The client then schedules the request and when it receives the response, it passes along the body of the response back to AVS via the `wcall_sft_resp` function.

## Testing

1. Receive an incoming conference call
2. Accept
3. See the call connects
#### APK
[Download build #2252](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2252/artifact/build/artifact/wire-dev-PR2896-2252.apk)
[Download build #2253](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2253/artifact/build/artifact/wire-dev-PR2896-2253.apk)

PR #2900, opened 2020-06-29
[Feature] Join a conference call

johnxnguyen:
## What's new in this PR?

This is a feature branch for the basic implementation of conference calling. With this PR, the client will be able to join a conference call. The client however, will **not** be able to start a conference call. All outgoing calls are still normal group calls.

## Notes

This PR consists of the following merged PRs:

- [x] https://github.com/wireapp/wire-android/pull/2892
- [x] https://github.com/wireapp/wire-android/pull/2895
- [x] https://github.com/wireapp/wire-android/pull/2896

#### APK
[Download build #2265](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2265/artifact/build/artifact/wire-dev-PR2900-2265.apk)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -14,7 +14,7 @@ ext {
     avsGroup = 'com.wire'
 
     // proprietary avs artifact configuration
-    customAvsVersion = '6.0.36@aar'
+    customAvsVersion = '6.0.39@aar'
     customAvsInternalVersion = customAvsVersion
     customAvsName = 'avs'
     customAvsGroup = 'com.wearezeta.avs'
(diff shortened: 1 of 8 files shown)


johnxnguyen:
> Are we going to keep this code behind a feature flag? I don't see anything related to that in this PR

@mejdoo yes we will, but currently all of these changes can't and don't need to be behind a feature flag. The reason is because this PR doesn't change the client behavior in regards to how we receive calls. It does add new capabilities
to satisfy AVS 6.

The next changes for conference calling will likely require the feature flag, for example when we start a call, the flag would determine whether we start a legacy call or conference call.

PR #2907, opened 2020-07-01
fix: Team data gets updated but new team data is not stored

makingthematrix:
fixes https://wearezeta.atlassian.net/browse/AN-6857

I introduced this bug a few weeks ago when I was working on team data updates - I removed the possibility to receive a completely new team data. This PR fixes that. The side-effect, which led to discovering the bug, was that the team name didn't appear on the profile view for new team members.

#### APK
[Download build #2280](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2280/artifact/build/artifact/wire-dev-PR2907-2280.apk)
[Download build #2281](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2281/artifact/build/artifact/wire-dev-PR2907-2281.apk)

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
@@ -47,7 +47,7 @@ trait TeamsService {
 
   val selfTeam: Signal[Option[TeamData]]
 
-  def onTeamUpdated(id: TeamId, name: Option[Name], icon: AssetId): Future[Unit]
+  def onTeamUpdated(team: TeamData): Future[Unit]
 
   def onTeamSynced(team: TeamData, members: Seq[TeamMember], roles: Set[ConversationRole]): Future[Unit]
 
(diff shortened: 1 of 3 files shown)


wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/teams/TeamsService.scala
```diff
@@ -204,12 +207,9 @@ class TeamsServiceImpl(selfUser:           UserId,
...
 
-  override def onTeamUpdated(id: TeamId, name: Option[Name], icon: AssetId): Future[Unit] = {
-    verbose(l"onTeamUpdated: $id, name: $name, icon: $icon")
-    teamStorage.update(id, team => team.copy (
-      name    = name.getOrElse(team.name),
-      icon    = icon)
-    ).flatMap(_ => lastTeamUpdate := Instant.now())
+  override def onTeamUpdated(team: TeamData): Future[Unit] = {
+    verbose(l"onTeamUpdated: ${team.id}, name: ${team.name}, icon: ${team.icon}")
+    teamStorage.insert(team).flatMap(_ => lastTeamUpdate := Instant.now())
```
gizemb, line 212: I assume `insert` updates the team data if it already exists, right?

makingthematrix: @gizemb : Yes. It's replacing the whole old entry. That's why in this one situation when we get an update event, which has only the name and the avatar id, we have to first create the entry. Actually, that one could use the `update` method instead, but then there's this log statement, and this `flatMap` at the end, and I didn't want to duplicate this code...

PR #2918, opened 2020-07-07
[Fix] Restore calling tests

johnxnguyen:
### Issues

`CallingServiceSpec` is ignored. When it is not ignored, almost all tests failed.

### Causes

It appears there were some issues with the test with the sync engine was migrated to the ui project. Over time, some changes to the code base have made the tests fail in various ways, such as new methods are being called that aren't being mocked, accessing the global module and not calling participant changed handler.

### Solutions

Mock all unmocked methods. Inject dependencies rather than access the global module. Invoke the handler.

#### APK
[Download build #2311](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2311/artifact/build/artifact/wire-dev-PR2918-2311.apk)


johnxnguyen:
build again

PR #2308, opened 2019-09-25
feat: Delete conversation (#AN-6378)

gizemb:
### Issues

Added support for responding to [conversation delete](https://wearezeta.atlassian.net/browse/AN-6378) events.

### Testing

Unit tests are provided for the feature. Notifications should also be tested due to refactoring.

PR #2920, opened 2020-07-08
Fix : Wrong welcome screen shown while being connected to a custom backend

mejdoo:
### Issues

A customer reported that his users are able to create an account or a new team on their custom backend.

https://wearezeta.atlassian.net/browse/AN-6988

### Causes

Clicking on the back button in OKTA webview always redirects the user to the standard welcome screen. When the user is already connected to a custom backend, he should be redirected to Enterprise welcome screen. 

### Solutions

Checking if the user is connected to a custom backend and show the right welcome screen in `SSOWebViewFragment`.

### Testing

- I click on Enterprise login and I enter en email address ending with @qa-demo.com. 
- I'm redirected to the custom backend.
- I click the back arrow in the top left corner. 
- I should see the Enterprise Welcome screen, not the standard one.
#### APK
[Download build #2308](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2308/artifact/build/artifact/wire-dev-PR2920-2308.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/appentry/SSOWebViewFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/appentry/SSOWebViewFragment.scala
@@ -27,21 +27,22 @@ import androidx.appcompat.widget.Toolbar
 import androidx.fragment.app.FragmentManager
 import com.waz.service.{AccountsService, ZMessaging}
 import com.waz.threading.Threading
+import com.waz.threading.Threading._
 import com.waz.utils._
 import com.waz.utils.wrappers.URI
 import com.waz.zclient.appentry.DialogErrorMessage.EmailError
 import com.waz.zclient.appentry.SSOWebViewWrapper.SSOResponse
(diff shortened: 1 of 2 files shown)


app/src/main/scala/com/waz/zclient/appentry/SSOWebViewFragment.scala
```diff
@@ -87,7 +88,12 @@ class SSOWebViewFragment extends FragmentHelper {
   }
 
   override def onBackPressed(): Boolean = {
-    getFragmentManager.popBackStack(SSOWebViewFragment.Tag, FragmentManager.POP_BACK_STACK_INCLUSIVE)
+    activity.getSupportFragmentManager.popBackStackImmediate(SSOWebViewFragment.Tag, FragmentManager.POP_BACK_STACK_INCLUSIVE)
+    if (backendController.hasCustomBackend) {
```
BradleyWire, line 92: Can we move this logic to a private method for readability? Also, we could optimise the showing of a fragment by having it's own private method to avoid repeating ourselves too.

PR #2925, opened 2020-07-13
New backups (1): The data sources, DAOs, etc. - classes needed for the functionality

makingthematrix:
This is the initial structure of classes used in serialization and deserialization (so it also performs a part of a functionality from a future ticket). I ran into a problem with customized serialization of byte arrays and while learning about it realized that we will probably have similar problems in the future with other fields. I decided to write all those intermediate classes so that no matter if we decide to switch to another JSON implementation or if the underlying database classes change, the JSON layer will be abstract enough to work with this. 

I needed to create `ButtonsDao`. It was not needed before because the Buttons table was created only after the bug migration and buttons were not touched at all in Kotlin until now. I believe it does not require a new migration.

I tried to use Base64 for byte array encoding and failed. I'm not exactly sure of the reason, but the unit test for comparing serialized and deserialized array with the origin was failing. I used Android's `Base64Utils`. There is another implementation of Base64, but it needs API 26 minimum. Currently we have to support APIs below that. In the end, I decided to convert bytes to ints and store the arrays as regular arrays of numbers in JSON. 

**EDIT**: The custom byte array conversions were removed in https://github.com/wireapp/wire-android/pull/2928


app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSource.kt
```diff
@@ -0,0 +1,29 @@
...
+
+import com.waz.model.Handle
+import com.waz.model.UserId
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.shared.backup.BackupRepository
+import com.waz.zclient.shared.backup.Password
+import com.waz.zclient.shared.backup.datasources.local.*
+import java.io.File
+
+class BackupDataSource(
```
android10, line 11: This is for me already a code smell: This class is doing too many things by just looking at the number of collaborators.

makingthematrix: @android10 : Any ideas how to do it differently? We need to gather together data coming from all 13 data sources. We will need a way to access them all, one after another, in a few methods of this class.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -0,0 +1,17 @@
...
+abstract class BackupLocalDataSource<T> {
+    protected val json by lazy { Json(JsonConfiguration.Stable.copy(isLenient = true, ignoreUnknownKeys = true)) }
+
+    abstract suspend fun getAll(): List<T>
+    abstract fun serialize(entity: T): String
+    abstract fun deserialize(jsonStr: String): T
+
+    companion object {
+        fun toIntArray(bytes: ByteArray?): IntArray? = bytes?.map { it.toInt() }?.toIntArray()
+        fun toByteArray(ints: IntArray?): ByteArray? = ints?.map { it.toByte() }?.toByteArray()
```
gizemb, line 15: i might be asking this for the 2nd time, but why did we need this conversion exactly?

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSource.kt
android10: Well, for example I see 4 DataSources starting with ```conversation...```. Maybe group them in one generic ```ConversationDataSource``` (or something like this). Let's be more granular. 

Otherwise this will end up in an anti-pattern called **God Class** which will bring up a maintenance nightmare

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSource.kt
```diff
@@ -0,0 +1,29 @@
...
+        private val conversationLocalDataSource: ConversationsLocalDataSource,
+        private val conversationMemberLocalDataSource: ConversationMemberLocalDataSource,
+        private val conversationRoleActionLocalDataSource: ConversationRoleActionLocalDataSource,
+        private val foldersLocalDataSource: FoldersLocalDataSource,
+        private val keyValueLocalDataSource: KeyValuesLocalDataSource,
+        private val likesLocalDataSource: LikesLocalDataSource,
+        private val messagesLocalDataSource: MessagesLocalDataSource,
+        private val propertiesLocalDataSource: PropertiesLocalDataSource,
+        private val readReceiptsLocalDataSource: ReadReceiptsLocalDataSource,
+        private val userLocalDataSource: UserLocalDataSource
```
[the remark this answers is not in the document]
android10: It is a good point! If we have another repository containing and grouping other DataSources, let's access through the repository. That might also fix my comment above. Good catch @gizemb

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
```diff
@@ -0,0 +1,75 @@
...
+        val size: Int = 0,
+        val source: String? = null,
+        val preview: String? = null,
+        val details: String = "",
+        val conversationId: String? = null
+) {
+    override fun hashCode(): Int =
+        id.hashCode() + token.hashCode() + name.hashCode() + encryption.hashCode() +
+                mime.hashCode() + size.hashCode() + source.hashCode() + preview.hashCode() +
+                details.hashCode() + conversationId.hashCode() //+ (sha?.size ?: 0)
```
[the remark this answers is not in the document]
android10: Let's remove this to overridden methods. That is the purpose of a data class. This is very prone to introduce bugs and issues. 

**Data class:**
The compiler automatically derives the following members from all properties declared in the primary constructor:

```kotlin
    equals()/hashCode() pair;
    toString() of the form "User(name=John, age=42)";
    componentN() functions corresponding to the properties in their order of declaration;
    copy() function (see below).
```

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    override fun equals(other: Any?): Boolean =
+        other != null && other is AssetsJSONEntity && other.id == id && other.token == token &&
+                other.name == name && other.encryption == encryption && other.mime == mime &&
+                other.size == size && other.source == source && other.preview == preview &&
+                other.details == details && other.conversationId == conversationId &&
+                ((other.sha == null && sha == null) || other.sha != null && sha !== null &&
+                    other.sha.zip(sha).all { it.first == it.second }
+                )
```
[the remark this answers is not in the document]
android10: Check my previous comment. In java that was needed. In Kotlin it is not required, especially by declaring this as a data class.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
```diff
@@ -0,0 +1,75 @@
...
+
+    fun toEntity(): AssetsEntity = AssetsEntity(
```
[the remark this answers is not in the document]
android10: AFAIK the compiler will not complain if you get rid of the explicit type.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ButtonLocalDataSource.kt
```diff
@@ -0,0 +1,41 @@
...
+    override suspend fun getAll(): List<ButtonEntity> = buttonDao.allButtons()
+
+    override fun serialize(entity: ButtonEntity): String =
+        json.stringify(ButtonJSONEntity.serializer(), ButtonJSONEntity.from(entity))
+    override fun deserialize(jsonStr: String): ButtonEntity =
+        json.parse(ButtonJSONEntity.serializer(), jsonStr).toEntity()
+}
+
+@Serializable
+data class ButtonJSONEntity(
```
android10, line 17: Format here as discussed.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ButtonLocalDataSource.kt
```diff
@@ -0,0 +1,41 @@
+package com.waz.zclient.shared.backup.datasources.local
+
+import com.waz.zclient.storage.db.buttons.ButtonDao
+import com.waz.zclient.storage.db.buttons.ButtonEntity
+import kotlinx.serialization.Serializable
+
+class ButtonLocalDataSource(private val buttonDao: ButtonDao): BackupLocalDataSource<ButtonEntity>() {
+    override suspend fun getAll(): List<ButtonEntity> = buttonDao.allButtons()
+
+    override fun serialize(entity: ButtonEntity): String =
```
android10, line 10: All these explicit types could be removed.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ButtonLocalDataSource.kt
```diff
@@ -0,0 +1,41 @@
...
+        json.stringify(ButtonJSONEntity.serializer(), ButtonJSONEntity.from(entity))
+    override fun deserialize(jsonStr: String): ButtonEntity =
```
android10, line 12: Same here

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ButtonLocalDataSource.kt
android10: Check the rest of the code, you will see many places where you can get rid of this. Especially coming from ```Scala``` this is something familiar for you :)

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ConversationFoldersLocalDataSource.kt
```diff
@@ -0,0 +1,32 @@
...
+    override suspend fun getAll(): List<ConversationFoldersEntity> = conversationFoldersDao.allConversationFolders()
+
+    override fun serialize(entity: ConversationFoldersEntity): String =
+        json.stringify(ConversationFoldersJSONEntity.serializer(), ConversationFoldersJSONEntity.from(entity))
+    override fun deserialize(jsonStr: String): ConversationFoldersEntity =
+        json.parse(ConversationFoldersJSONEntity.serializer(), jsonStr).toEntity()
+}
+
+@Serializable
+data class ConversationFoldersJSONEntity(
```
android10, line 17: Format here

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/MessagesLocalDataSource.kt
```diff
@@ -0,0 +1,121 @@
...
+    val ephemeral: Int? = null,
+    val expiryTime: Int? = null,
+    val expired: Boolean = false,
+    val duration: Int? = null,
+    val quote: String? = null,
+    val quoteValidity: Int = 0,
+    val forceReadReceipts: Int? = null,
+    val assetId: String? = null
+) {
+    override fun hashCode(): Int =
```
android10, line 45: Check again my comment. These methods are not necessary. The compiler will do the magic here.

android10:
Something else I would like to point out is the fact that this PR is so hard to follow due to its size. We agreed as a team to keep them not so big. So please next time, try to split them up into smaller ones.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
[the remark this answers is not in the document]
gizemb: yes, you're right, data class generates the equals and hashCode methods, but you're also missing the fact that if the class contains an `Array` field, it results in a false negative.

The generated `equals` method of the `data class` compares the fields with their default "equals" method, but for arrays, `equals` method checks the _reference equality_, not the content. So, for this code block:

```
    val a1 = intArrayOf(1,2,3)
    val a2 = intArrayOf(1,2,3)
    
    println("equal? " + (a1 == a2))
    
    println("contentEqual? " + a1.contentEquals(a2))
```

The output is: 
```
equal? false
contentEqual? true
```

That's why it is always recommended to override `equals` and `hashCode` in data classes, if you have array fields. I think we also care about the content equality here rather than the reference equality, so i say that Maciek's approach is correct.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSource.kt
makingthematrix: @android10 : It's just similar names. They are all different data sources, connected to different daos. I created one abstract class for them all, but that's for common functionality. And I understand that the standard way to inject data sources is through a constructor. On the Scala side, in SE, services and handlers also look like that exactly because the DI works through constructors. On the other hand, this pattern is easy to understand. I don't think we ever had maintenance issues with it.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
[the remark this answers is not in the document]
android10: Fair enough!

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSource.kt
android10: This is clearly an anti-pattern, this class is doing too many things. If we never had any maintenance problem was mainly because of luck. This class is doing too many things. 

You have too many collaborators. If its difficult to test which means design could be improved. It is just a code smell:
 - https://refactoring.guru/smells/long-parameter-list
 - https://www.codingrevolution.com/long-parameter-list/

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
makingthematrix: We don't. I removed it in https://github.com/wireapp/wire-android/pull/2928

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/MessagesLocalDataSource.kt
```diff
@@ -0,0 +1,121 @@
...
+}
+
+@Serializable
+data class MessagesJSONEntity(
+    val id: String,
+    val conversationId: String = "",
+    val messageType: String = "",
+    val userId: String = "",
+    val content: String? = null,
+    val protos: IntArray? = null,
```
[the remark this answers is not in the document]
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -0,0 +1,17 @@
+package com.waz.zclient.shared.backup.datasources.local
+
+import kotlinx.serialization.json.Json
+import kotlinx.serialization.json.JsonConfiguration
+
+abstract class BackupLocalDataSource<T> {
+    protected val json by lazy { Json(JsonConfiguration.Stable.copy(isLenient = true, ignoreUnknownKeys = true)) }
```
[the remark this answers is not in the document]
makingthematrix: In the end, I moved this line to the companion object and I give a JSON's "serializer" to the data source as an argument.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSource.kt
makingthematrix: Changed in https://github.com/wireapp/wire-android/pull/2929

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSource.kt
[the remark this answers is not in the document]
makingthematrix: Changed in https://github.com/wireapp/wire-android/pull/2929

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ButtonLocalDataSource.kt
makingthematrix: Done in https://github.com/wireapp/wire-android/pull/2929

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ButtonLocalDataSource.kt
makingthematrix: https://github.com/wireapp/wire-android/pull/2929

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ButtonLocalDataSource.kt
makingthematrix: Done in https://github.com/wireapp/wire-android/pull/2929

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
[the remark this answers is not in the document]
makingthematrix: Done in https://github.com/wireapp/wire-android/pull/2929

PR #2927, opened 2020-07-14
New backups(2): Reading entities from db in batches

makingthematrix:
The most important method here is `suspend fun next(): List<T>`. It will allow us to read batches of entries from the given table. When the result is an empty list it means we have read everything.


app/src/main/kotlin/com/waz/zclient/core/functional/Utils.kt
```diff
@@ -0,0 +1,8 @@
+package com.waz.zclient.core.functional
+
+object Utils {
```
android10, line 3: When using names like: Utils or Managers as part of something's name, means we are not clear about the intention of this. Even what does ```Util``` mean? And what does ```returning``` mean?

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
```diff
@@ -8,6 +8,8 @@ import kotlinx.serialization.Serializable
 
 class AssetLocalDataSource(private val assetsDao: AssetsDao): BackupLocalDataSource<AssetsEntity>() {
     override suspend fun getAll(): List<AssetsEntity> = assetsDao.allAssets()
+    override suspend fun getInBatch(batchSize: Int, offset: Int): List<AssetsEntity> =
```
android10, line 11: :+1:

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -1,17 +1,32 @@
...
 
+    private var currentOffset: Int = 0
+
     abstract suspend fun getAll(): List<T>
+    protected abstract suspend fun getInBatch(batchSize: Int, offset: Int): List<T>
+
     abstract fun serialize(entity: T): String
     abstract fun deserialize(jsonStr: String): T
 
+    suspend fun next(): List<T> = returning(getInBatch(batchSize, currentOffset)) {
```
android10, line 18: Maybe rename this to something more meaningful? This could mean many things, even the first thing that came to my mind when I saw it, it is the ```next()``` method of a RxJava monadic type: ```Observable```. 

I think in this case, this means something completely different.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -1,17 +1,32 @@
...
+        currentOffset += this.batchSize
+    }
+
+    fun reset(): Unit {
```
android10, line 22: What is here being reset?

app/src/main/kotlin/com/waz/zclient/core/functional/Utils.kt
makingthematrix: It's a copy of a similar method in Scala. It's going to help me write expressions. 
It can be used like this: `fun foo(): T = returning(T()){ t -> ... }`. This means that I want to create and return a new instance of `T`, but before that happens I first want to do something with it. Without `returning` I would have to write:
```
fun foo(): T {
  val t = T()
  ... // doing something with t
  return t
}
```

It can also be used in bigger blocks of code as a way to initialize local values. With `returning` the code that deals only with setting up that new value is scoped more precisely, so it's more readable and there's less risk that it will be moved around during refactoring.
```
...
val t = returning(T()) { t -> 
  ... /// what is declared here will be valid only in this block
}
...
```

makingthematrix: As for `Utils`: I just need to to put `returning` somewhere. A utils object looked like a usual way to do it.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
makingthematrix: Changed to `suspend fun nextJSONArrayAsString(): String?`

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
makingthematrix: Changed to `fun resetCurrentOffset()`

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -15,27 +14,33 @@ abstract class BackupLocalDataSource<EntityType, JSONType>(
...
 
-    suspend fun nextJSONString(): String? {
-        val list = next()
-        return if (list.isNotEmpty()) serializeList(list) else null
-    }
+    fun getCurrentOffset(): Int = currentOffset
 
-    fun reset(): Unit {
-        currentOffset = 0
+    suspend fun nextJSONArrayAsString(): String? {
```
android10, line 36: Do we want to return a ```null``` here? Maybe an empty string?

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -15,27 +14,33 @@ abstract class BackupLocalDataSource<EntityType, JSONType>(
     private var currentOffset: Int = 0
 
     protected abstract suspend fun getInBatch(batchSize: Int, offset: Int): List<EntityType>
-    protected abstract fun toJSONType(entity: EntityType): JSONType
-    protected abstract fun toEntityType(json: JSONType): EntityType
+    protected abstract fun toJSON(entity: EntityType): JSONType
+    protected abstract fun toEntity(json: JSONType): EntityType
+
+    fun serialize(entity: EntityType): String = json.stringify(serializer, toJSON(entity))
```
android10, line 20: I do not think you need the type here. Should be implicit.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -15,27 +14,33 @@ abstract class BackupLocalDataSource<EntityType, JSONType>(
...
+
+    fun deserialize(jsonStr: String): EntityType = toEntity(json.parse(serializer, jsonStr))
```
android10, line 22: Same here: type should be implicit.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/FoldersLocalDataSource.kt
```diff
@@ -4,13 +4,15 @@ import com.waz.zclient.storage.db.folders.FoldersDao
 import com.waz.zclient.storage.db.folders.FoldersEntity
 import kotlinx.serialization.Serializable
 
-class FoldersLocalDataSource(private val foldersDao: FoldersDao):
-    BackupLocalDataSource<FoldersEntity, FoldersJSONEntity>(FoldersJSONEntity.serializer()) {
+class FoldersLocalDataSource(
```
android10, line 7: Format as discussed looks a bit weird here. Is ```detekt``` also complaining about it?

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/KeyValuesLocalDataSource.kt
```diff
@@ -4,13 +4,15 @@ import com.waz.zclient.storage.db.property.KeyValuesDao
 import com.waz.zclient.storage.db.property.KeyValuesEntity
 import kotlinx.serialization.Serializable
 
-class KeyValuesLocalDataSource(private val keyValuesDao: KeyValuesDao):
-    BackupLocalDataSource<KeyValuesEntity, KeyValuesJSONEntity>(KeyValuesJSONEntity.serializer()) {
+class KeyValuesLocalDataSource(
+        private val keyValuesDao: KeyValuesDao,
+        batchSize: Int = BatchSize
+): BackupLocalDataSource<KeyValuesEntity, KeyValuesJSONEntity>(KeyValuesJSONEntity.serializer(), batchSize) {
```
android10, line 10: Same point as above with the code format.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/MessagesLocalDataSource.kt
```diff
@@ -6,15 +6,18 @@ import com.waz.zclient.storage.db.messages.MessagesDao
 import com.waz.zclient.storage.db.messages.MessagesEntity
 import kotlinx.serialization.Serializable
 
-class MessagesLocalDataSource(private val messagesDao: MessagesDao):
-    BackupLocalDataSource<MessagesEntity, MessagesJSONEntity>(MessagesJSONEntity.serializer()) {
+class MessagesLocalDataSource(
+    private val messagesDao: MessagesDao,
+    batchSize: Int = BatchSize
+): BackupLocalDataSource<MessagesEntity, MessagesJSONEntity>(MessagesJSONEntity.serializer(), batchSize) {
```
android10, line 12: Same. Not strong about all this formatting if we agree on something and we maintain it across the entire codebase.

app/src/test/kotlin/com/waz/zclient/shared/backup/datasources/BackupLocalDataSourceTest.kt
```diff
@@ -0,0 +1,63 @@
...
+
+    @Mock
+    private lateinit var keyValuesDao: KeyValuesDao
+
+    @Test
+    fun `read all data as a json array string`(): Unit = runBlocking {
+        val dataSource = KeyValuesLocalDataSource(keyValuesDao)
+        val keyValuesEntity = KeyValuesEntity("key", "value")
+
+        Mockito.`when`(keyValuesDao.getKeyValuesInBatch(ArgumentMatchers.anyInt(), ArgumentMatchers.anyInt())).thenReturn(listOf(keyValuesEntity))
```
android10, line 26: You can get rid of the ```Mockito.``` static call and add the import.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
makingthematrix: @android10 : I need a good way to inform that there is nothing more to read. I think `null` is better for that than an empty string. An empty string might me mistaken for something that was actually read.

android10: Just wanted to be clear on the meaning of ```null``` since it could mean anything and always tends to confuse. 
Fair enough!

app/src/test/kotlin/com/waz/zclient/shared/backup/datasources/BackupLocalDataSourceTest.kt
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/core/functional/Utils.kt
gizemb: Kotlin already has built-in functions for that purpose. take a look at `let`, `apply`, etc. you can do the same with 
`val t = T().let{ ... }` for example
https://kotlinlang.org/docs/reference/scope-functions.html

makingthematrix: Great :)

makingthematrix: done in https://github.com/wireapp/wire-android/pull/2929

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/FoldersLocalDataSource.kt
makingthematrix: Fixed in https://github.com/wireapp/wire-android/pull/2928

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/KeyValuesLocalDataSource.kt
makingthematrix: Fixed in https://github.com/wireapp/wire-android/pull/2928

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/MessagesLocalDataSource.kt
makingthematrix: Fixed in https://github.com/wireapp/wire-android/pull/2928

PR #2330, opened 2019-10-03
Chore: Failing CI tests

marcoconti83:
## What's new in this PR?

Fix CI tests constantly failing.

### Issues

The main causes of failures were:
- Global timeouts (15 seconds to finish the test). This can occasionally be reached on the CI machine due to network timeouts, GC-delay due to previous tests, etc.
- A specific test, `com.waz.service.EventSchedulerSpec`, was failing constantly on CI. This test was creating 42.000 events and scheduling them. I assume this hits some memory ceiling that constantly triggers GC and slows down the test to a halt.

### Solutions

I increased the timeout from 15 to 30 seconds, and reduce the number of events created in `com.waz.service.EventSchedulerSpec` from 42.000 to 1.000.

### Testing

The nightly run with these changes ran successfully on CI at least once.

#### APK
[Download build #108](http://10.10.124.11:8080/job/Pull%20Request%20Builder/108/artifact/build/artifact/wire-dev-PR2330-108.apk)

PR #2928, opened 2020-07-15
New backups (3): Changes requested in the code reviews and by detekt

makingthematrix:
#### APK
[Download build #2330](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2330/artifact/build/artifact/wire-dev-PR2928-2330.apk)
[Download build #2331](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2331/artifact/build/artifact/wire-dev-PR2928-2331.apk)

### DIFF
--- a/app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
+++ b/app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
@@ -1,7 +1,5 @@
 package com.waz.zclient.shared.backup.datasources.local
 
-import com.waz.zclient.shared.backup.datasources.local.BackupLocalDataSource.Companion.toByteArray
-import com.waz.zclient.shared.backup.datasources.local.BackupLocalDataSource.Companion.toIntArray
 import com.waz.zclient.storage.db.assets.AssetsDao
 import com.waz.zclient.storage.db.assets.AssetsEntity
 import kotlinx.serialization.Serializable
@@ -9,7 +7,7 @@ import kotlinx.serialization.Serializable
 class AssetLocalDataSource(
     private val assetsDao: AssetsDao,
     batchSize: Int = BatchSize
-): BackupLocalDataSource<AssetsEntity, AssetsJSONEntity>(AssetsJSONEntity.serializer(), batchSize) {
+) : BackupLocalDataSource<AssetsEntity, AssetsJSONEntity>(AssetsJSONEntity.serializer(), batchSize) {
     override suspend fun getInBatch(batchSize: Int, offset: Int): List<AssetsEntity> =
         assetsDao.getAssetsInBatch(batchSize, offset)
 
@@ -17,20 +15,20 @@ class AssetLocalDataSource(
     override fun toEntity(json: AssetsJSONEntity): AssetsEntity = json.toEntity()
 }
 
-@SuppressWarnings("ComplexMethod")
+@SuppressWarnings("ComplexMethod", "ParameterListWrapping")
 @Serializable
 data class AssetsJSONEntity(
-        val id: String,
-        val token: String? = null,
-        val name: String = "",
-        val encryption: String = "",
-        val mime: String = "",
-        val sha: IntArray? = null,
-        val size: Int = 0,
-        val source: String? = null,
-        val preview: String? = null,
-        val details: String = "",
-        val conversationId: String? = null
+    val id: String,
+    val token: String? = null,
+    val name: String = "",
+    val encryption: String = "",
+    val mime: String = "",
+    val sha: ByteArray? = null,
+    val size: Int = 0,
+    val source: String? = null,
+    val preview: String? = null,
+    val details: String = "",
+    val conversationId: String? = null
 ) {
     override fun hashCode(): Int =
         id.hashCode() + token.hashCode() + name.hashCode() + encryption.hashCode() +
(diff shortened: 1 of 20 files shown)


app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
```diff
@@ -20,39 +17,37 @@ class AssetLocalDataSource(
...
 
     override fun equals(other: Any?): Boolean =
         other != null && other is AssetsJSONEntity && other.id == id && other.token == token &&
         other.name == name && other.encryption == encryption && other.mime == mime &&
         other.size == size && other.source == source && other.preview == preview &&
         other.details == details && other.conversationId == conversationId &&
-        ((other.sha == null && sha == null) || other.sha != null && sha !== null &&
-            other.sha.zip(sha).all { it.first == it.second }
-        )
+        ((other.sha == null && sha == null) || other.sha != null && sha !== null && other.sha.contentEquals(sha))
```
gizemb, line 42: `(other.sha == null && sha == null)` can be shortened as `(other.sha == sha)`. this returns true if both are null or they have the same reference, so in that case we avoid unnecessary contentEquals check.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -44,8 +42,7 @@ abstract class BackupLocalDataSource<EntityType, JSONType>(
     }
 
     companion object {
-        fun toIntArray(bytes: ByteArray?): IntArray? = bytes?.map { it.toInt() }?.toIntArray()
-        fun toByteArray(ints: IntArray?): ByteArray? = ints?.map { it.toByte() }?.toByteArray()
+        protected val json by lazy { Json(JsonConfiguration.Stable.copy(isLenient = true, ignoreUnknownKeys = true)) }
```
gizemb, line 45: it's still doesn't look very test friendly 🤔 how did you test this part?

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ReadReceiptsLocalDataSource.kt
```diff
@@ -4,10 +4,8 @@ import com.waz.zclient.storage.db.receipts.ReadReceiptsDao
 import com.waz.zclient.storage.db.receipts.ReadReceiptsEntity
 import kotlinx.serialization.Serializable
 
-class ReadReceiptsLocalDataSource(
-    private val readReceiptsDao: ReadReceiptsDao,
-    batchSize: Int = BatchSize
-): BackupLocalDataSource<ReadReceiptsEntity, ReadReceiptsJSONEntity>(ReadReceiptsJSONEntity.serializer(), batchSize) {
+class ReadReceiptsLocalDataSource(private val readReceiptsDao: ReadReceiptsDao, batchSize: Int = BatchSize) :
```
gizemb, line 7: isn't there a better way to avoid passing the same BatchSize in every extending class?

app/src/test/kotlin/com/waz/zclient/shared/backup/datasources/AssetLocalDataSourceTest.kt
```diff
@@ -67,7 +51,7 @@ class AssetLocalDataSourceTest : UnitTest() {
         val assetsJSONEntity = AssetsJSONEntity.from(assetsEntity)
         val result: AssetsEntity = assetsJSONEntity.toEntity()
 
-        result.id shouldEqual assetsEntity.id
+        result shouldEqual assetsEntity
```
gizemb, line 54: i think we use assertJ instead of kluent, but i'm not sure :)

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
android10: Good catch! Agree!

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
makingthematrix: Right. Thanks.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
makingthematrix: @gizemb , @android10 :
I test that conversion between entities and their corresponding `*JSONEntity` data classes work, and I test that I can both make json strings from them and recreate them from json strings. I think it covers all cases. I don't need to mock kotlinx json implementation.

app/src/test/kotlin/com/waz/zclient/shared/backup/datasources/AssetLocalDataSourceTest.kt
makingthematrix: I see kluent everywhere in unit tests...

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ReadReceiptsLocalDataSource.kt
makingthematrix: I don't know. Actually, this one here is important for unit tests. The batch size needs to be set to a low number to test if data is read and written in many batches.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/AssetLocalDataSource.kt
makingthematrix: @gizemb : Done, although Android Studio says it's a "dangerous comparison".

PR #2344, opened 2019-10-09
[Feature] Favourites folder ui

johnxnguyen:
## What's new in this PR?

This PR contains changes to show the favorites folder. Additionally, a bug where a favorites group or one to one conversation was only shown in the favorites folder has been fixed.

#### APK
[Download build #172](http://10.10.124.11:8080/job/Pull%20Request%20Builder/172/artifact/build/artifact/wire-dev-PR2344-172.apk)
[Download build #181](http://10.10.124.11:8080/job/Pull%20Request%20Builder/181/artifact/build/artifact/wire-dev-PR2344-181.apk)
[Download build #182](http://10.10.124.11:8080/job/Pull%20Request%20Builder/182/artifact/build/artifact/wire-dev-PR2344-182.apk)
[Download build #184](http://10.10.124.11:8080/job/Pull%20Request%20Builder/184/artifact/build/artifact/wire-dev-PR2344-184.apk)
[Download build #185](http://10.10.124.11:8080/job/Pull%20Request%20Builder/185/artifact/build/artifact/wire-dev-PR2344-185.apk)
[Download build #186](http://10.10.124.11:8080/job/Pull%20Request%20Builder/186/artifact/build/artifact/wire-dev-PR2344-186.apk)

PR #2339, opened 2019-10-06
Implement add to/remove from Favorites folder menu for a conversation

gizemb:
### Issues

Add "Add to Favorites" and "Remove from Favorites" menu items for a conversation & call related service methods.

#### APK
[Download build #133](http://10.10.124.11:8080/job/Pull%20Request%20Builder/133/artifact/build/artifact/wire-dev-PR2339-133.apk)
[Download build #147](http://10.10.124.11:8080/job/Pull%20Request%20Builder/147/artifact/build/artifact/wire-dev-PR2339-147.apk)
[Download build #148](http://10.10.124.11:8080/job/Pull%20Request%20Builder/148/artifact/build/artifact/wire-dev-PR2339-148.apk)
[Download build #149](http://10.10.124.11:8080/job/Pull%20Request%20Builder/149/artifact/build/artifact/wire-dev-PR2339-149.apk)
[Download build #150](http://10.10.124.11:8080/job/Pull%20Request%20Builder/150/artifact/build/artifact/wire-dev-PR2339-150.apk)
[Download build #152](http://10.10.124.11:8080/job/Pull%20Request%20Builder/152/artifact/build/artifact/wire-dev-PR2339-152.apk)
[Download build #153](http://10.10.124.11:8080/job/Pull%20Request%20Builder/153/artifact/build/artifact/wire-dev-PR2339-153.apk)

PR #2929, opened 2020-07-17
New backups (4): Writing JSON strings to files

makingthematrix:
Functionality for writing json to text files.

#### APK
[Download build #2338](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2338/artifact/build/artifact/wire-dev-PR2929-2338.apk)
[Download build #2339](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2339/artifact/build/artifact/wire-dev-PR2929-2339.apk)
[Download build #2344](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2344/artifact/build/artifact/wire-dev-PR2929-2344.apk)
[Download build #2347](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2347/artifact/build/artifact/wire-dev-PR2929-2347.apk)
[Download build #2361](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2361/artifact/build/artifact/wire-dev-PR2929-2361.apk)


app/src/main/kotlin/com/waz/zclient/core/utilities/SuspendIterable.kt
```diff
@@ -0,0 +1,57 @@
...
+    val it = this.iterator()
+    var finished = false
+
+    do {
+        val value = runBlocking { it.next() }
+        if (value != null) fn(value) else finished = true
+    } while (!finished)
+}
+
+fun <T, S> SuspendIterable<T>.map(fn: (T) -> S): List<S> {
```
makingthematrix, line 25: I tried to find an easy way to have a `map` method which returns a `SuspendIterable<S>` but even the Kotlin collections return `List<S>` in such case, so I gave up :) I think for our purposes this is fine.

app/src/main/kotlin/com/waz/zclient/core/utilities/SuspendIterable.kt
```diff
@@ -0,0 +1,57 @@
...
+
+    do {
+        val value = runBlocking { it.next() }
+        if (value != null) results += fn(value) else finished = true
+    } while (!finished)
+
+    return results.toList()
+}
+
+fun <T, S> SuspendIterable<T>.mapOrFail(fn: (T) -> Either<Failure, S>): Either<Failure, List<S>> {
```
makingthematrix, line 38: The `map` method above goes through the whole collection, no matter what the consecutive calls return. But in many cases we want to stop at the first failure.

app/src/main/kotlin/com/waz/zclient/core/utilities/UiUtils.kt
```diff
@@ -0,0 +1,22 @@
...
+import java.io.File
+import java.io.IOException
+
+object UiUtils {
+    data class IOFailure(val exception: IOException) : FeatureFailure()
+
+    fun writeTextToFile(targetDir: File, fileName: String, text: () -> String): Either<Failure, File> =
+        try {
+            val file = File(targetDir, fileName)
+            if (file.exists()) file.delete()
```
makingthematrix, line 15: If the file exists we replace it. Adding new text at the end would corrupt the data in many cases (e.g. json)

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -41,6 +43,10 @@ abstract class BackupLocalDataSource<EntityType, JSONType>(
         }
     }
 
+    override fun iterator(): SuspendIterator<String> = object : SuspendIterator<String> {
```
makingthematrix, line 46: A local data source can now give us an iterator that will return data in batches, every batch already converted to a json string, ready to be written to a file.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSource.kt
```diff
@@ -21,21 +28,57 @@ import com.waz.zclient.shared.backup.datasources.local.UserLocalDataSource
...
+            acc.flatMap { files ->
+                writeToFiles(targetDir, pair.first, pair.second).map { listOf(files, it).flatten() }
+            }
+        }
+
+    private fun <Entity, JSON> writeToFiles(
+        targetDir: File,
+        fileNamePrefix: String,
+        dataSource: BackupLocalDataSource<Entity, JSON>
+    ): Either<Failure, List<File>> {
```
makingthematrix, line 77: For the given data source, we read data in batches and for every batch create a new text file. In the end, we return either the collection of all created files, or the failure reason. If there was a failure, we stop the process at that moment. We are interested only in having the whole backup. A partial backup is of no use.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSource.kt
```diff
@@ -21,21 +28,57 @@ import com.waz.zclient.shared.backup.datasources.local.UserLocalDataSource
...
+        "folders" to foldersLocalDataSource,
+        "keyValues" to keyValueLocalDataSource,
+        "likes" to likesLocalDataSource,
+        "messages" to messagesLocalDataSource,
+        "properties" to propertiesLocalDataSource,
+        "readReceipts" to readReceiptsLocalDataSource,
+        "users" to usersLocalDataSource
+    )
+
+    private fun writeAllToFiles(targetDir: File) =
```
makingthematrix, line 65: For every data source, we read data in batches, create files for batches and write data in them, and in the end return the list of all files containing data from all data sources.

The `fold` method, in contrast to the custom `mapOrFail`, will always go through the whole list of data sources. But the `Either.flatMap` used inside it ensures that if a failure happens, we will just go through all subsequent data sources doing nothing - `writeToFiles` for the new data source is called only if all the previous ones were successful.

app/src/main/kotlin/com/waz/zclient/core/utilities/SuspendIterable.kt
```diff
@@ -0,0 +1,57 @@
...
+import com.waz.zclient.core.exception.Failure
+import com.waz.zclient.core.functional.Either
+import kotlinx.coroutines.runBlocking
+
+interface SuspendIterable<T> {
+    fun iterator(): SuspendIterator<T>
+}
+
+interface SuspendIterator<T> {
+    suspend fun next(): T?
```
gizemb, line 12: i don't understand why we needed an Iterable type of our own just to make the `next()` method suspendable. we're calling `runBlocking` in the extension methods anyway. instead, we could just move `runBlocking{}` to the `next()` method of our Iterator.

```
abstract class BackupLocalDataSource<EntityType, JSONType>(..) : Iterable<String?> {

    override fun iterator(): Iterator<String?> = object : Iterator<String?> {
        override fun hasNext(): Boolean // this actually is next() != null, but we need to avoid fetching it twice here.
        override fun next(): String? = runBlocking { nextJSONArrayAsString() }
    }
}
```

something that i also noticed that with this approach, we have to duplicate all the nice Iterable methods that Kotlin gives us. For instance, if we go with the code block above, i believe we could simplify this logic:

```
 var index = 0

return dataSource.mapOrFail {
      returning(writeTextToFile(targetDir, "${fileNamePrefix}_$index.json") { it }) { ++index }
}
```

with this one:

```
dataSource.forEachIndexed { index, s -> 
    writeTextToFile(targetDir, "${fileNamePrefix}_$index.json") { s }
}
```
(here i assumed that forEachIndexed doesn't return a value when `hasNext()` returns false, but we better double check)


there are also many nice methods such as `takeWhile{ }`, which can be useful in "stop at first failure" situations.

app/src/main/kotlin/com/waz/zclient/core/utilities/SuspendIterable.kt
```diff
@@ -0,0 +1,57 @@
...
+        val value = runBlocking { it.next() }
+        if (value != null) {
+            when (val res = fn(value)) {
+                is Either.Right -> results += res.b
+                is Either.Left -> failure = res.a
+            }
+        } else {
+            finished = true
+        }
+    } while (!finished || failure != null)
```
makingthematrix, line 54: ok, I already see that this should be `!finished && failure == null`

app/src/main/kotlin/com/waz/zclient/core/utilities/SuspendIterable.kt
makingthematrix: @gizemb : Okay. I'll try to figure out how to write the standard `Iterator` and then use `takeWhile`. The thing is, without having everyone in the office, I'm forced to either spam you all with questions, which I don't want to do too often,  or go through docs and try to figure out how I can make it work the way I want. And sometimes I come up with too complex ideas. Thanks :)

gizemb: no problem :) that's a big feature, so it's normal to make patches more than once.

app/src/main/kotlin/com/waz/zclient/core/utilities/SuspendIterable.kt
makingthematrix: Changed to a standard Iterator.

app/src/main/kotlin/com/waz/zclient/core/utilities/SuspendIterable.kt
makingthematrix: Changed to using a standard Iterator and named it `mapRight`.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
makingthematrix: Changed to a standard Iterator

app/src/main/kotlin/com/waz/zclient/core/utilities/SuspendIterable.kt
makingthematrix: done

app/src/main/kotlin/com/waz/zclient/shared/backup/BackupRepository.kt
```diff
@@ -2,9 +2,12 @@ package com.waz.zclient.shared.backup
...
 import com.waz.model.Handle
 import com.waz.model.UserId
+import com.waz.zclient.core.exception.Failure
 import com.waz.zclient.core.functional.Either
 import java.io.File
 
 interface BackupRepository {
     suspend fun exportDatabase(userId: UserId, userHandle: Handle, backupPassword: String, targetDir: File): Either<String, File>
+
+    fun writeAllToFiles(targetDir: File): Either<Failure, List<File>>
```
android10, line 12: What does mean **All** here?

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -1,43 +1,47 @@
...
+        override fun hasNext(): Boolean = currentOffset < listSize
 
-    suspend fun nextJSONArrayAsString(): String? {
-        val list = getInBatch(batchSize, currentOffset)
-        return if (list.isEmpty()) {
-            null
-        } else {
-            currentOffset += list.size
-            serializeList(list)
+        override fun next(): String = runBlocking {
```
android10, line 37: why this ```runBlocking``` block here?

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ButtonLocalDataSource.kt
```diff
@@ -21,7 +18,7 @@ data class ButtonJSONEntity(
     val ordinal: Int = 0,
     val state: Int = 0
 ) {
-    fun toEntity(): ButtonEntity = ButtonEntity(
+    fun toEntity() = ButtonEntity(
```
android10, line 21: This format is a bit weird, it can fit in the same line right? We already discussed this to have 140 chars per line. Is ```Detekt``` swallowing this?

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ButtonLocalDataSource.kt
```diff
@@ -30,7 +27,7 @@ data class ButtonJSONEntity(
     )
 
     companion object {
-        fun from(entity: ButtonEntity): ButtonJSONEntity = ButtonJSONEntity(
+        fun from(entity: ButtonEntity) = ButtonJSONEntity(
```
android10, line 30: same here with the format.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ConversationFoldersLocalDataSource.kt
```diff
@@ -4,27 +4,25 @@ import com.waz.zclient.storage.db.conversations.ConversationFoldersDao
...
-    override fun toEntity(json: ConversationFoldersJSONEntity): ConversationFoldersEntity = json.toEntity()
+class ConversationFoldersLocalDataSource(dao: ConversationFoldersDao, batchSize: Int = BatchSize) :
+BackupLocalDataSource<ConversationFoldersEntity, ConversationFoldersJSONEntity>
+    ("conversationFolders", dao, batchSize, ConversationFoldersJSONEntity.serializer()) {
+    override fun toJSON(entity: ConversationFoldersEntity) = ConversationFoldersJSONEntity.from(entity)
+    override fun toEntity(json: ConversationFoldersJSONEntity) = json.toEntity()
 }
 
 @Serializable
 data class ConversationFoldersJSONEntity(
```
android10, line 15: Format again. This should look like this:
```kotlin
data class ConversationFoldersJSONEntity(val convId: String = "", val folderId: String = "") {
...
}
```

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ConversationMembersLocalDataSource.kt
```diff
@@ -4,25 +4,23 @@ import com.waz.zclient.storage.db.conversations.ConversationMembersDao
...
+BackupLocalDataSource<ConversationMembersEntity, ConversationMembersJSONEntity>
+    ("conversationMembers", dao, batchSize, ConversationMembersJSONEntity.serializer()) {
+    override fun toJSON(entity: ConversationMembersEntity) = ConversationMembersJSONEntity.from(entity)
+    override fun toEntity(json: ConversationMembersJSONEntity) = json.toEntity()
 }
 
 @Serializable
 data class ConversationMembersJSONEntity(val userId: String = "", val conversationId: String = "", val role: String = "") {
-    fun toEntity(): ConversationMembersEntity = ConversationMembersEntity(
+    fun toEntity() = ConversationMembersEntity(
```
android10, line 16: Format

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ConversationMembersLocalDataSource.kt
```diff
@@ -4,25 +4,23 @@ import com.waz.zclient.storage.db.conversations.ConversationMembersDao
...
         userId = userId,
         conversationId = conversationId,
         role = role
     )
 
     companion object {
-        fun from(entity: ConversationMembersEntity): ConversationMembersJSONEntity = ConversationMembersJSONEntity(
+        fun from(entity: ConversationMembersEntity) = ConversationMembersJSONEntity(
```
android10, line 23: Please check all this code format

app/src/main/kotlin/com/waz/zclient/core/functional/Either.kt
```diff
@@ -119,3 +119,28 @@ fun <L, R> Either<L, R>.getOrElse(value: R): R =
...
     }
+
+/**
+ * Maps over an iterable of T where the mapping function returns an Either<F, S>.
+ * If the mapping function returns Left<F>, signifying an error, the whole mapRight is interrupted and
+ * the error is returned. If the mapping function returns Right for all elements, the result
+ * is Right<List<S>>. You can think of this method as a functional version of a map wrapped in
+ * a try/catch because the mapping function can throw an exception.
+ */
+fun <T, S, F> Iterable<T>.mapRight(fn: (T) -> Either<F, S>): Either<F, List<S>> {
```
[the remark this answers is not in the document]
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ButtonLocalDataSource.kt
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ButtonLocalDataSource.kt
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ConversationFoldersLocalDataSource.kt
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ConversationMembersLocalDataSource.kt
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/core/utilities/UiUtils.kt
```diff
@@ -0,0 +1,22 @@
+package com.waz.zclient.core.utilities
+
+import com.waz.zclient.core.exception.Failure
+import com.waz.zclient.core.exception.FeatureFailure
+import com.waz.zclient.core.functional.Either
+import java.io.File
+import java.io.IOException
+
+object UiUtils {
```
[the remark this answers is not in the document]
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/ConversationMembersLocalDataSource.kt
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -1,43 +1,47 @@
...
-    private val batchSize: Int = BatchSize
-) {
-    private var currentOffset: Int = 0
-
-    protected abstract suspend fun getInBatch(batchSize: Int, offset: Int): List<EntityType>
+    val name: String,
+    protected val dao: BatchReader<EntityType>,
+    protected val batchSize: Int,
+    private val serializer: KSerializer<JSONType>
+) : Iterable<String> {
```
gizemb, line 16: I feel like this class knows too much. we usually structure the repository class as a user of other smaller components. Take a look at other repositories in the project. usually, the repository has some smaller components and it just combines them to get the result. a classical example would be:

1. fetch from remote (`val remoteModels = remoteDataSource.fetch()`)
2. convert remote response to local entity (`val entities = mapper.from(remoteModels)`)
3. save local entity to database (`localDataSource.save(entities)`)

------

I also see the back up functionality like that, but instead of remote and local data sources, we have one data source that fetches from the backup file, and another data source that saves to the database -or the other way around-:

**creating a back up file:**
1. fetch from database (`val entities = databaseDataSource.fetch()`)
2. convert entity to back up model (`val models = mapper.fromEntity(entities)`)
3. save model to back up file (`backUpDataSource.save(models)`)

**restoring from a back up file:**
1. fetch from back up file (`val backUpModels = backUpDataSource.fetch()`)
2. convert back up model to db entity (`val entities = mapper.toEntity(backUpModels)`)
3. save entities to database (`databaseDataSource.save(entities)`)

```
 // I named it BackUp..., because it's the feature name, but i'm sure there's a better naming
interface BackUpDataSource<T> {   
    suspend fun write(models: List<T>): Either<Failure, Unit>
    suspend fun read(): Either<Failure, List<T>>
}


abstract class BackUpRepository<T, E> {
    abstract val backUpDataSource: BackUpDataSource<T>    //for backup model, like ButtonJSONEntity
    abstract val databaseDataSource: BackUpDataSource<E>   //for database model, like ButtonEntity
    abstract val converter: BackUpDataConverter<T, E>

    suspend fun backUp() =
        databaseDataSource.read()
            .map { converter.fromEntity(it) }
            .flatMap { backUpDataSource.write(it) }

    suspend fun restore() =
        backUpDataSource.read()
            .map { converter.toEntity(it) }
            .flatMap { databaseDataSource.write(it) }
}
```
-----

and all the other logic in this class could go to somewhere else. for instance:
- private val serializer: KSerializer<JSONType>
- serialization logic & serialize/deserialize methods -> these are implementation details about the library that we use. it's also only required for File IO, so it can be a dependency of the File reader/writer. I'd like to see them to be encapsulated in a JsonConverter class. And then the FileReader can use this JsonConverter
```
class JsonConverter<T>(private val serializer: KSerializer<T>) {
    private val json = ..

    fun toJson(models: List<T>): String = json.stringify(serializer.list, models)
    fun fromJson(str: String): List<T> = json.parse(serializer.list, str)
}
```

- batch size -> implementation logic about the reader/writer
- name parameter -> just concerns the File reader/writer
- iterator -> i don't know if it's possible but i'd like to see that as a part of writer too

The `backUpDataSource` and `databaseDataSource` that the Repository uses can make use of such writer for their read()/write() operations: 
```
class BatchedFileIOHandler<T>(
    private val fileName: String,
    private val jsonConverter: JsonConverter<T>,
    private val batchSize: Int
)  {
    fun writeToFile(models: List<T>): Either<Failure, Unit> {
        val json = jsonConverter.toJson(models)
        //write json to file
    }

    fun readFromFile(): Either<Failure, List<T>> {
        val jsonStr = //read file in batches
        val models = jsonConverter.toJson(jsonStr)
        //return in Either
    }
}
```

- toJson/toEntity conversions -> this could be extracted into a Converter class as I mentioned above

```
interface BackUpDataConverter<T, E> {
    fun fromEntity(entity: E): T
    fun toEntity(model: T): E
}
```

makingthematrix: I think this is very important, but this is a lot. So, I'd like to suggest the following:
1. Merge these four PRs as they are now.
2. Add one more with zipping and encrypting the backup files. This one will also move the logic around so that we will have more smaller classes, for unit testing.
3. And then, on top of the new PR, we can come back to this. 
I still don't really "feel" this approach. I think I'd like to talk more about this, after I come back from vacations.

gizemb: ok let's merge it for now so you don't get blocked. and then we can have a call etc about it and maybe others want to participate in the discussion too.

app/src/test/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSourceTest.kt
```diff
@@ -0,0 +1,101 @@
...
+    @Before
+    fun setup() {
+        keyValuesLocalDataSource = KeyValuesLocalDataSource(keyValuesDao, 3)
+        propertiesLocalDataSource = PropertiesLocalDataSource(propertiesDao, 2)
+        val dataSources = listOf(keyValuesLocalDataSource, propertiesLocalDataSource)
+        backupDataSource = BackupDataSource(dataSources)
+    }
+
+    @Test
+    fun `create files and write down json arrays to them`() = runBlocking {
```
gizemb, line 56: we try to use given/when/then naming structure in tests.

`given KeyValueEntities, when writeAllToFiles is called, then creates Properties files from them`

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
```diff
@@ -1,43 +1,47 @@
...
-    suspend fun nextJSONArrayAsString(): String? {
-        val list = getInBatch(batchSize, currentOffset)
-        return if (list.isEmpty()) {
-            null
-        } else {
-            currentOffset += list.size
-            serializeList(list)
+        override fun next(): String = runBlocking {
+            val list = dao.getBatch(batchSize, currentOffset)
+            if (list == null || list.isEmpty()) {
```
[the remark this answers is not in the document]
makingthematrix: ok

app/src/test/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSourceTest.kt
```diff
@@ -0,0 +1,101 @@
...
+import kotlinx.coroutines.runBlocking
+import org.amshove.kluent.shouldEqual
+import org.junit.Before
+import org.junit.Test
+import org.mockito.Mock
+import org.mockito.Mockito.`when`
+import java.nio.file.Files
+
+@ExperimentalCoroutinesApi
+class BackupDataSourceTest : UnitTest() {
```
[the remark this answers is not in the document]
makingthematrix: @gizemb : It tests the `writeAllToFiles` method which is the only public method in `BackupDataSource`.

app/src/test/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSourceTest.kt
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
android10: I'm with @gizemb on this one. I think the code is not consistent with the structure we defined at architecture level, plus it does not respect many of the good practices. 

I also do not think it makes sense to merge it if we are not happy about it.  
Is there any good reason for this? I would say it is always better to fix it. 

Can we have a meeting to discuss all this? My points to treat:
 - Architecture consistency
 - Smaller PR with only a tiny backup that does not include everything (divide and conquer approach)
 - Kotlin Programming Style (This should not be a port from any other Programming language)
 - Tests: I see there is room for improvement since these are more integration than unit tests. 

Maybe @makingthematrix you can schedule a meeting? It is really hard and time consuming to review such a big PR.

app/src/test/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSourceTest.kt
android10: :+1: that is also the type of consistency I commented above.

makingthematrix:
build again

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
BradleyWire: After a discussion with Gizem, we're going to show an example of the propose solution in a separate PR and then we can fine - tune it from there. But from a high level perspective this solution has everything we need for it to be consistent with what we have in the codebase already.

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
BradleyWire: AFAIK it was to appease a linting issue

app/src/test/kotlin/com/waz/zclient/shared/backup/datasources/BackupDataSourceTest.kt
```diff
@@ -0,0 +1,101 @@
...
+        PropertiesEntity("p1", "A"),
+        PropertiesEntity("p2", "B"),
+        PropertiesEntity("p3", "C"),
+        PropertiesEntity("p4", "D"),
+        PropertiesEntity("p5", "E")
+    )
+
+    @Before
+    fun setup() {
+        keyValuesLocalDataSource = KeyValuesLocalDataSource(keyValuesDao, 3)
```
BradleyWire, line 49: Would be good to split these tests into their own classes as this test class could get huge

app/src/main/kotlin/com/waz/zclient/shared/backup/datasources/local/BackupLocalDataSource.kt
mejdoo: I agree with @android10 but we should not forget that this is the first trial for @makingthematrix with clean architecture in Kotlin.  I would prefer to see @makingthematrix carrying on backup stuff after the end of his vacations with the help of either @gizemb or @BradleyWire.

PR #2355, opened 2019-10-11
Fix: Convert remote conversation IDs in folders payload to lowercase

marcoconti83:
## What's new in this PR?

Fixing cross-platform compatibility with iOS folders metadata

### Causes

iOS uses uppercase UUIDs, Android expects lowercase UUIDs.

iOS generated payload
```json
{
  "labels" : [
    {
      "id" : "C06641F4-62E6-48F8-8E09-A3EBEA09058B",
      "type" : 1,
      "conversations" : [
        "5AE4CA03-2FDD-4CD5-8F62-F96FF6DA4F22",
        "3653D8CF-85F7-4B95-9A14-F03EC2F44F7D"
      ]
    }
  ]
}
```

Android generated payload
```{
  "labels" : [
    {
      "id" : "C06641F4-62E6-48F8-8E09-A3EBEA09058B",
      "name" : "",
      "type" : 1,
      "conversations" : [
        "3653d8cf-85f7-4b95-9a14-f03ec2f44f7d"
      ]
    }
  ]
}
```

### Solutions

Lowercase all conversation IDs when parsing them from the payload

### Testing

Covered by unit tests, and tested on real device against an iOS device

#### APK
[Download build #206](http://10.10.124.11:8080/job/Pull%20Request%20Builder/206/artifact/build/artifact/wire-dev-PR2355-206.apk)

Issue #2354, opened 2019-10-10
URL get broken when containing %20


Transisto:
%20 get converted to spaces.

marcoconti83:
Hi @Transisto ,
thanks for the report. Can you specify what broken means? Does it not highlight it as a URL?

Transisto:
You're right, It just look very different but the link actually work.

PR #2939, opened 2020-07-24
[Feature] Targeted calling messages

johnxnguyen:
## What's new in this PR?

This PR introduces changes to support targeting calling messages. 

### Context

AVS needs to send encrypted messages, which it must do through the client code. It achieves this by invoking the `SendHandler`, passing in the message payload and conversation id (amongst other things). The client code then proceeds to post the message in the conversation as it would normally do.

The `SendHandler` already has two parameters (`userIdDest` and `clientIdDest`) which, if not `null`, specify a single target recipient. In AVS 6.2, the semantics of these two parameters have changed, such that `userIdDest` now contains a json string specifying a set of target recipients. `clientIdDest` will now always be `null`. This re-purposing was done on the AVS side to avoid introducing new API changes.

### Changes

- Rename the `userIdDest` parameter in the `SendHandler` to `targetRecipientsJson`
- Check the `userIdSelf` and clientIdSelf` parameters to ensure they match the current self user
- Decode the `targetRecipientsJson` to `AvsClientList`, use it to send a targeted message.

#### APK
[Download build #2351](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2351/artifact/build/artifact/wire-dev-PR2939-2351.apk)
[Download build #2354](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2354/artifact/build/artifact/wire-dev-PR2939-2354.apk)
[Download build #2355](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2355/artifact/build/artifact/wire-dev-PR2939-2355.apk)
[Download build #2365](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2365/artifact/build/artifact/wire-dev-PR2939-2365.apk)

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/call/CallingServiceImpl.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/call/CallingServiceImpl.scala
@@ -215,13 +215,13 @@ class CallingServiceImpl(val accountId:       UserId,
   )
 
   def onSend(ctx: Pointer, msg: String, convId: RConvId, targetRecipients: Option[AvsClientList]): Future[Unit] =
-    withConv(convId) { (_, conv) =>
+    withConv(convId) { (wCall, conv) =>
       val recipients = targetRecipients match {
         case Some(clientList) => TargetRecipients.SpecificClients(clientsMap(clientList.clients.toSet))
         case None             => TargetRecipients.ConversationParticipants
       }
 
-      sendCallMessage(conv.id, GenericMessage(Uid(), GenericContent.Calling(msg)), recipients, ctx)
+      sendCallMessage(wCall, conv.id, GenericMessage(Uid(), GenericContent.Calling(msg)), recipients, ctx)
     }
 
   private def clientsMap(avsClients: Set[AvsClient]): Map[UserId, Set[ClientId]] = {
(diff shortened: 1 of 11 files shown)


wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/call/CallingServiceImpl.scala
```diff
@@ -484,30 +497,35 @@ class CallingServiceImpl(val accountId:       UserId,
...
+
+  private def sendCallMessage(wCall: WCall, convId: ConvId, msg: GenericMessage, targetRecipients: TargetRecipients, ctx: Pointer): Unit = {
+    verbose(l"Sending msg on behalf of avs: convId: $convId")
+    otrSyncHandler.postOtrMessage(convId, msg, targetRecipients).map {
+      case Right(_) =>
+        updateActiveCall(_.copy(outstandingMsg = None))("sendCallMessage/verified")
+        avs.onHttpResponse(wCall, 200, "", ctx)
+      case Left(ErrorResponse.Unverified) =>
+        warn(l"Conversation degraded, delay sending message on behalf of AVS")
+        //TODO need to handle degrading of conversation during a call
```
franziskuskiefer, line 520: Please link an issue for this TODO.

wire-android-sync-engine/zmessaging/src/test/scala/com/waz/service/call/CallingServiceSpec.scala
```diff
@@ -1051,6 +1055,59 @@ class CallingServiceSpec extends AndroidFreeSpec with DerivedLogTag {
...
+      (avs.answerCall _).expects(*, *, *, *).once().onCall { (_, _, _, _) =>
+        service.onEstablishedCall(groupConv.remoteId, otherUserId)
+        service.onParticipantsChanged(groupConv.remoteId, Set(otherUser, otherUser2))
+      }
+
+      service.startCall(groupConv.id)
+      awaitCP(checkpoint1)
+    }
+
+    scenario("Messages are targeted if target recipients are specified") {
```
franziskuskiefer, line 1083: What happens if the targeted recipient doesn't exist or is not supposed to get the message? One or two negative test cases would be good here.

johnxnguyen: The scope of these tests is to assert that when AVS tells us to send a message to specific clients, then we invoke the `postOtrMessage` method with specific arguments.

The negative test case for a non existent client should be asserted on a deeper layer, probably where we encrypt the payloads. I'll add one there.

I'm not sure the second case makes sense here. By definition if they're a target then they should receive the message. Whether they're the right target should, I think, be asserted in AVS since the client has no knowledge of the contents of the message payload.

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/call/CallingServiceImpl.scala
johnxnguyen: Honestly I'm not sure what this TODO is referring to, but I'll try to see if they call really does "just time out".

franziskuskiefer: :D we can also remove he TODO :)

wire-android-sync-engine/zmessaging/src/test/scala/com/waz/service/call/CallingServiceSpec.scala
franziskuskiefer: 👍 ok, sounds good

johnxnguyen:
build again

wire-android-sync-engine/zmessaging/src/test/scala/com/waz/service/call/CallingServiceSpec.scala
johnxnguyen: If it's fine with you, I would add the tests in a separate PR.

franziskuskiefer: wfm

johnxnguyen:
build again

Issue #2373, opened 2019-10-17
Not possible to make linebreaks in messages


n3eo:
**Describe the bug**
When writing a message to a contact it is not possible to create a linebreak in the message any more.

**To Reproduce**
Setting set to show additional send button in UI

1. Write message to any contact
2. Write anything
3. Try to make line break
    I use the Huawei Swipe keyboard in German -> Button on bottom left says "Weiter" instead of the usual line break symbol
4. Focuses out of keyboard, no linebreak

**Expected behavior**
Show linebreake button and make a line break in message.

**Screenshots**
1. Writing the message
![Screenshot_20191017-103933](https://user-images.githubusercontent.com/36729513/66992737-98da1580-f0ca-11e9-8208-e3f49613248a.png)
2. Clicking on "Weiter"
![Screenshot_20191017-103938](https://user-images.githubusercontent.com/36729513/66992771-abece580-f0ca-11e9-8fa3-52a38921cab0.png)

**Smartphone:**
- Wire version that exhibits the issue: 3.41.848
- Last Wire version that did not exhibit the issue (if applicable): 3.41.844 (not sure)
- Android OS version used to run Wire: 7.0
- Is this a custom firmware or a stock one: stock
- Mobile phone model/manufacturer: Huawei P9
- Mobile network type (EDGE/LTE/Wi-Fi/Offline): makes no difference

**Additional context**
When changing the setting to send by pressing enter, it is possible to make linebreaks, but you can't send anymore.

marcoconti83:
Hi @n3eo ,
thanks for the report, we are looking into a fix

Transisto:
You should have pushed a fix to you newly added bug a few days after.

Last 50 reviews are almost all 1 Star, I suggest you get your shit together.

marcoconti83:
Hi all,
A fix for this has been prepared [here](https://github.com/wireapp/wire-android/pull/2384) and should be released soon.

marcoconti83:
Hi all,
This fix has just been released to public with version 3.42. I'll close the issue, feel free to reopen if this is still a problem.

PR #2942, opened 2020-07-28
[Feature] Conference calling beta

johnxnguyen:
## What's new in this PR?

This PR introduces the beta program toggle for conference calling (replacing the previous feature flag in `BuildConfig`). The setting stored per account. When enabled, all outgoing group calls will be conference calls and group video restrictions are removed. By default, the beta program is disabled.

## Notes

The client is still able participate in conference calls even when it is not participating in the beta program. This is possible when it receives a conference call from another user.

#### APK
[Download build #2373](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2373/artifact/build/artifact/wire-dev-PR2942-2373.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/calling/views/ControlsView.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/views/ControlsView.scala
@@ -22,6 +22,7 @@ import android.content.Context
 import android.util.AttributeSet
 import android.view.View
 import android.widget.GridLayout
+import com.waz.content.UserPreferences
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
 import com.waz.permissions.PermissionsService
 import com.waz.service.call.Avs.VideoState._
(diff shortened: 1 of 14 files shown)


app/src/main/scala/com/waz/zclient/calling/views/ControlsView.scala
```diff
@@ -78,14 +80,23 @@ class ControlsView(val context: Context, val attrs: AttributeSet, val defStyleAt
...
+      conv             <- controller.conversation
+      isGroup          <- zms.conversations.groupConversation(conv.id)
+      isTeam           =  zms.teamId.isDefined
+      established      <- controller.isCallEstablished
+      showVideo        <- controller.isVideoCall
+      members          <- controller.conversationMembers.map(_.size)
+      isConferenceCall <- controller.isConferenceCall
+    } yield {
+      if (isGroup && isConferenceCall) {
+        established
```
gizemb, line 93: doesn't that already fall into the `else` case?

johnxnguyen: Yes good point, I'll fix this.

johnxnguyen: Fixed

PR #2617, opened 2020-02-17
Bump internal avs version to 5.5.6

gizemb:
https://wearezeta.atlassian.net/browse/AN-6667

## What's new in this PR?

Updates AVS library to version 5.5.6

#### APK
[Download build #1273](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1273/artifact/build/artifact/wire-dev-PR2617-1273.apk)

PR #2401, opened 2019-11-06
Chore: Close database cursors

gizemb:
## What's new in this PR?

Closes open database cursors to release their resources.
#### APK
[Download build #310](http://10.10.124.11:8080/job/Pull%20Request%20Builder/310/artifact/build/artifact/wire-dev-PR2401-310.apk)
[Download build #319](http://10.10.124.11:8080/job/Pull%20Request%20Builder/319/artifact/build/artifact/wire-dev-PR2401-319.apk)

PR #2945, opened 2020-07-29
Feat: Backup

gizemb:
## What's new in this PR?

This PR adds some basic components for back up functionality.

**Structure:** 
Each table will have a dedicated `BackUpRepository` (or, `BackUpDataSource` as a concrete class) which is responsible from back up/restore operations of that particular table.

`CreateBackUpUseCase`: use case that creates a complete back up by invoking all repositories
`BackUpRepository`: Repository interface with back up/restore operations
`BackUpIOHandler`: Base interface for i/o operations. This can be implemented by a file reader, or a Dao class in the future.
`BackUpDataMapper`: Base interface for conversion between entity (database) models and back up models.
`BackUpDataSource`: Class that combines database and back up file i/o collaborators. Reads from one source and writes to another. Each table will implement their own `BackUpDataSource` with specific `BackUpIOHandlers`. 

### Testing

Unit tests are provided for concrete classes/methods.

## Notes

BackUpDataSource will return Either instead of Unit as a result of backUp() & restore() operations.

https://github.com/wireapp/wire-android/pull/2946 was also merged into this PR. 
https://github.com/wireapp/wire-android/pull/2947 was also merged into this PR. 
https://github.com/wireapp/wire-android/pull/2948 was also merged into this PR.

#### APK
[Download build #2385](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2385/artifact/build/artifact/wire-dev-PR2945-2385.apk)
[Download build #2401](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2401/artifact/build/artifact/wire-dev-PR2945-2401.apk)
[Download build #2404](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2404/artifact/build/artifact/wire-dev-PR2945-2404.apk)
[Download build #2414](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2414/artifact/build/artifact/wire-dev-PR2945-2414.apk)
[Download build #2422](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2422/artifact/build/artifact/wire-dev-PR2945-2422.apk)
[Download build #2440](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2440/artifact/build/artifact/wire-dev-PR2945-2440.apk)
[Download build #2462](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2462/artifact/build/artifact/wire-dev-PR2945-2462.apk)
[Download build #2617](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2617/artifact/build/artifact/wire-dev-PR2945-2617.apk)
[Download build #2645](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2645/artifact/build/artifact/wire-dev-PR2945-2645.apk)
[Download build #2659](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2659/artifact/build/artifact/wire-dev-PR2945-2659.apk)
[Download build #2716](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2716/artifact/build/artifact/wire-dev-PR2945-2716.apk)
[Download build #2727](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2727/artifact/build/artifact/wire-dev-PR2945-2727.apk)


app/src/main/kotlin/com/waz/zclient/feature/backup/usecase/CreateBackUpUseCase.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.usecase.UseCase
+import com.waz.zclient.feature.backup.BackUpRepository
+
+class CreateBackUpUseCase(private val backUpRepositories: List<BackUpRepository>) : UseCase<Unit, Unit> {
+
+    override suspend fun run(params: Unit): Either<Failure, Unit> {
+        var failure = false
+        //TODO we can speed it up by async + await maybe?
+        for (repo in backUpRepositories) {
```
BradleyWire, line 14: You could do something: 

```
coroutineScope.launch {
    failure = backUpRepositories.map {
        async(Dispatchers.IO) {
            it.backUp()
        }
    }.awaitAll().any { it.isLeft }
}
```

You'll probably need to inject the scope for testing purposes. In some situations people use `GlobalScope` (but it [affects performance](https://medium.com/@elizarov/the-reason-to-avoid-globalscope-835337445abc))

app/src/main/kotlin/com/waz/zclient/feature/backup/BackUpDataSource.kt
```diff
@@ -0,0 +1,28 @@
...
+
+//TODO: implement BackUpRepository
+@Suppress("IteratorNotThrowingNoSuchElementException")
+//TODO handle NoSuchElement when signature changes
+abstract class BackUpDataSource<T, E> {
+    abstract val databaseLocalDataSource: BackUpIOHandler<E>
+    abstract val backUpLocalDataSource: BackUpIOHandler<T>
+    abstract val mapper: BackUpDataMapper<T, E>
+
+    fun backUp() {
```
mejdoo, line 11: I think the name `backUp()` is a little bit confusing. Can we use `save()` and `restore()`?

BradleyWire: I'd prefer we still kept backup, so i'll rename to `saveBackup` & `restoreBackup`. This will be in part 9.

makingthematrix:
build again

app/src/main/kotlin/com/waz/zclient/feature/backup/EncryptionHandler.kt
```diff
@@ -0,0 +1,300 @@
...
+import com.waz.zclient.core.logging.Logger.Companion.warn
+import com.waz.zclient.core.logging.Logger.Companion.verbose
+import org.libsodium.jni.NaCl
+import org.libsodium.jni.Sodium
+import java.io.File
+import java.io.IOException
+import java.nio.ByteBuffer
+import java.security.SecureRandom
+
+class EncryptionHandler {
```
mejdoo, line 23: This class looks very complex and it should be optimized to respect Kotlin idiomatic stlyle.
I'm seeing:
-  a lot of `if else `blocks.
- hardcoded strings i.e "_encrypted", "wire_backup"...
- strange variable name `ret`, `s`...
- 300 lines of code very hard to understand.

Please revert the last commit and reopen the PR #2970  to simply the review for everyone.

gizemb: I agree. I would also extract all `Sodium` lib things to another place like `Cryptographer` to reduce complexity here.

app/src/main/kotlin/com/waz/zclient/feature/backup/EncryptionHandler.kt
```diff
@@ -0,0 +1,300 @@
...
+    fun encrypt(backup: File, password: String, userId: UserId): Either<Failure, File> =
+        try {
+            loadLibrary
+
+            val salt = generateSalt()
+
+            getMetaDataBytes(salt, userId).flatMap { meta ->
+                val backupBytes = backup.readBytes()
```
gizemb, line 31: is `readBytes()` safe for huge files?

app/src/main/kotlin/com/waz/zclient/feature/backup/buttons/ButtonsBackupDataSource.kt
```diff
@@ -0,0 +1,42 @@
...
+import kotlinx.serialization.Serializable
+import java.io.File
+
+@Serializable
+data class ButtonsBackUpModel(
+    val messageId: String = String.empty(),
+    val buttonId: String = String.empty(),
+    val title: String = String.empty(),
+    val ordinal: Int = 0,
+    val state: Int = 0
```
gizemb, line 17: i think we shouldn't be giving default values here. we should force the values if they surely exist in db entity.

app/src/main/kotlin/com/waz/zclient/feature/backup/io/BatchReader.kt
```diff
@@ -0,0 +1,55 @@
...
+    /**
+     * Reads next item from source and returns the result.
+     *
+     * @return Either.Left(Failure) if there is an error,
+     * Either.Right(t) if next item is read successfully,
+     * Either.Right(null) if there are no more items to read.
+     */
+    suspend fun readNext(): Either<Failure, T?>
+
+    suspend fun hasNext(): Boolean
```
gizemb, line 17: I think `hasNext()` method here is not aligned with the contract of the `readNext() `method. What I originally had in mind was to have no hasNext() method. if readNext() returns null, we'd know that the input was consumed. Of course that implies that the iterator cannot have a nullable type but that was OK for the back up use case. Did we take this into attention while modifying this class? Now when I look at the methods, I see that the items can be nullable and I should always use the hasNext() before using readNext(), like Java's iterator. But the documentation says something else. I prefer updating the documentation if it doesn't represent the logic anymore. or updating the logic itself.

app/src/main/kotlin/com/waz/zclient/feature/backup/io/BatchReader.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    suspend fun performAction(value: T) =
+        when (val res = action(value)) {
+            is Either.Right -> rightValues += res.b
+            is Either.Left -> failure = res.a
+        }
+
+    while (hasNext() && failure == null) {
+        when (val next = readNext()) {
+            is Either.Right -> next.b?.let { performAction(it) }
```
gizemb, line 49: now that readNext() can return nullable, this line is problematic imo. what if the `action` wants to operate on null inputs as well? perhaps an action that receives String wants to convert null Strings to ""s. the method name is very generic but the implementation has hidden logics inside.

app/src/main/kotlin/com/waz/zclient/feature/backup/io/BatchReader.kt
```diff
@@ -0,0 +1,55 @@
...
+ */
+suspend fun <T, R> BatchReader<T>.mapRight(action: suspend (T) -> Either<Failure, R>): Either<Failure, List<R>> {
+    val rightValues = mutableListOf<R>()
+    var failure: Failure? = null
+
+    suspend fun performAction(value: T) =
+        when (val res = action(value)) {
+            is Either.Right -> rightValues += res.b
+            is Either.Left -> failure = res.a
+        }
```
gizemb, line 45: we should also switch to `Either` methods (onSuccess, fold, etc.) instead of `when` checks here

app/src/main/kotlin/com/waz/zclient/feature/backup/io/BatchReader.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    while (hasNext() && failure == null) {
+        when (val next = readNext()) {
+            is Either.Right -> next.b?.let { performAction(it) }
+            is Either.Left -> failure = next.a
+        }
+    }
```
gizemb, line 52: this can be shortened with Either methods. it should be sth like that (not tested):

```
readNext().flatMap {
  action(it).onSuccess { list.add(it) }
}.onFailure {
    failure = it
}
```

app/src/main/kotlin/com/waz/zclient/feature/backup/io/database/BatchDatabaseIOHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+import com.waz.zclient.feature.backup.io.mapRight
+import com.waz.zclient.storage.db.BatchDao
+
+class BatchDatabaseIOHandler<E>(
+    private val batchDao: BatchDao<E>,
+    private val batchSize: Int = BATCH_SIZE
+) : BackUpIOHandler<E, Unit> {
+    override suspend fun write(iterator: BatchReader<List<E>>): Either<Failure, List<Unit>> =
+        iterator.mapRight {
+            batchDao.insertAll(it)
```
gizemb, line 17: if `insertAll` throws an exception here, where does `mapRight` catch it? I used `requestDatabase` here before, to catch that exception but i think here it's just thrown and not caught anywhere.

app/src/main/kotlin/com/waz/zclient/feature/backup/io/database/BatchDatabaseIOHandler.kt
```diff
@@ -0,0 +1,35 @@
...
+            Either.Right(Unit)
+        }
+
+    override fun readIterator(): BatchReader<List<E>> = object : BatchReader<List<E>> {
+        private var count = 0
+        override suspend fun readNext(): Either<Failure, List<E>?> = requestDatabase {
+            batchDao.nextBatch(count, Math.min(batchDao.count() - count, batchSize)).also {
```
gizemb, line 24: again, no need to call `batchDao.count()` every time.

app/src/main/kotlin/com/waz/zclient/feature/backup/usecase/CreateBackUpUseCase.kt
```diff
@@ -0,0 +1,33 @@
...
+import kotlinx.coroutines.Dispatchers
+import kotlinx.coroutines.SupervisorJob
+import kotlinx.coroutines.async
+import kotlinx.coroutines.awaitAll
+
+import java.io.File
+
+class CreateBackUpUseCase(
+    private val backUpRepositories: List<BackUpRepository<File>>,
+    private val coroutineScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
```
gizemb, line 18: how does this scope work? I think it's not tied to the viewmodel's scope like we always do? is that something we want?

app/src/test/kotlin/com/waz/zclient/feature/FileTestHelpers.kt
```diff
@@ -0,0 +1,28 @@
...
+    deleteOnExit()
+}
+
+fun uniqueTextFileName(): String {
+    Thread.sleep(1)
+    return "ZipHandlerTest_${System.currentTimeMillis()}.txt"
+}
+
+fun uniqueZipFileName(): String {
+    Thread.sleep(1)
```
gizemb, line 21: why is there a sleep here?

app/src/test/kotlin/com/waz/zclient/feature/backup/BatchReaderTestHelpers.kt
```diff
@@ -0,0 +1,38 @@
...
+
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.feature.backup.io.BatchReader
+import com.waz.zclient.feature.backup.io.forEach
+import org.junit.Assert.assertEquals
+import org.junit.Assert.fail
+import org.mockito.Mockito.`when`
+
+suspend fun <T> BatchReader<T>.mockNextItems(items: List<T>) {
+    val itemsArray = items.map { Either.Right(it) }.plus(Either.Right(null)).toTypedArray()
```
gizemb, line 11: there is also the logic i mentioned above. it puts a null item at the end because this denotes that all the input is read. All the tests that are using this method should be reviewed if we're changing the "contract" here.

app/src/test/kotlin/com/waz/zclient/feature/backup/ZipHandlerTest.kt
```diff
@@ -0,0 +1,118 @@
...
+
+    @Test
+    fun `given a non-empty input text file, when zipped, then return a zipped non-empty file`() {
+        val tempDir = createTempDir()
+        val textFile = createTextFile(tempDir)
+        textFile.length() `should be greater than` 0
+
+        val zipHandler = ZipHandler(tempDir)
+
+        when (val res = zipHandler.zip(uniqueZipFileName(), listOf(textFile))) {
```
gizemb, line 25: I again prefer `onSuccess/onFailure` here instead of `when`

app/src/main/kotlin/com/waz/zclient/feature/backup/EncryptionHandler.kt
makingthematrix: Sorry, I merged it by mistake. I made a new PR for encryption: https://github.com/wireapp/wire-android/pull/2990

app/src/test/kotlin/com/waz/zclient/feature/backup/ZipHandlerTest.kt
makingthematrix: Changed in https://github.com/wireapp/wire-android/pull/2988/commits/33b9807244877b35dabaecc77665de3b55efd1d9

app/src/main/kotlin/com/waz/zclient/feature/backup/io/BatchReader.kt
makingthematrix: Fixed in https://github.com/wireapp/wire-android/pull/2988/commits/5eb2f1302f0791c7a7fb9d801cefda5037197a25 by making `readNext` return non-nullable results only.

app/src/main/kotlin/com/waz/zclient/feature/backup/io/BatchReader.kt
makingthematrix: Fixed in https://github.com/wireapp/wire-android/pull/2988/commits/5eb2f1302f0791c7a7fb9d801cefda5037197a25

app/src/main/kotlin/com/waz/zclient/feature/backup/io/database/BatchDatabaseIOHandler.kt
makingthematrix: In https://github.com/wireapp/wire-android/pull/2988/commits/3b236a713c6c3919e932fa88395f6cd9ed2fd709#diff-0153cfd65f695bc623e5befaa9dc0f1f I added a private variable, `daoCount`, which is used as a cache for the number of elements in dao. The dao is accessed only once to get that number.

app/src/test/kotlin/com/waz/zclient/feature/FileTestHelpers.kt
makingthematrix: Just a small hack to get a unique name in an easy way. In the next line, I use the current time in milliseconds as the suffix of the name. It turns out, in one or two tests the computer managed to ask for a new unique filename twice during one millisecond.

app/src/test/kotlin/com/waz/zclient/feature/backup/BatchReaderTestHelpers.kt
makingthematrix: Fixed in https://github.com/wireapp/wire-android/pull/2988

app/src/main/kotlin/com/waz/zclient/feature/backup/usecase/CreateBackUpUseCase.kt
BradleyWire: This was implemented before I knew a ViewModel would exist (didn't know how we'd attach the Scala / Kotlin code at that point), so happy to or would even encourage the use of a `viewModelScope`

mejdoo (approved):

app/src/main/kotlin/com/waz/zclient/feature/backup/EncryptionHandler.kt
mejdoo: 👍

makingthematrix:
build again

PR #2954, opened 2020-08-04
Feat: Back up - Part 9: Testing + PR comments 

BradleyWire:
## What's new in this PR?

This PR continues from: https://github.com/wireapp/wire-android/pull/2953

PR comments from part 1 through part 8 addressed here + some missing test cases for backup dao classes 
#### APK
[Download build #2456](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2456/artifact/build/artifact/wire-dev-PR2954-2456.apk)

PR #2956, opened 2020-08-04
Feat: Back up - Part 10: Single batched tables

BradleyWire:
## What's new in this PR?

This is a follow on from: https://github.com/wireapp/wire-android/pull/2954

### Issues

In this PR i've started adding the backup dependencies for the folders table and the conversation role actions table. These are the first two of four single batched tables. 

### Testing

Added test cases for the mappers, DAOs, backup DAOs and test data providers inside `:common-test`


app/src/main/kotlin/com/waz/zclient/feature/backup/conversations/ConversationRolesBackupDao.kt
```diff
@@ -0,0 +1,15 @@
+package com.waz.zclient.feature.backup.conversations
+
+import com.waz.zclient.feature.backup.io.database.SingleReadDao
+import com.waz.zclient.storage.db.conversations.ConversationRoleActionDao
+import com.waz.zclient.storage.db.conversations.ConversationRoleActionEntity
+
+class ConversationRolesBackupDao(private val conversationRoleActionDao: ConversationRoleActionDao) :
```
android10, line 7: Is there any especial reason why you need this extra layer? What I see here is that there is no extra logic, only delegation, which means that you could use directly the parent?

app/src/main/kotlin/com/waz/zclient/feature/backup/folders/FoldersDataSource.kt
```diff
@@ -0,0 +1,28 @@
...
+import com.waz.zclient.feature.backup.BackUpDataSource
+import com.waz.zclient.feature.backup.BackUpIOHandler
+import com.waz.zclient.storage.db.folders.FoldersEntity
+import kotlinx.serialization.Serializable
+
+@Serializable
+data class FoldersBackUpModel(
+    val id: String,
+    val name: String,
+    val type: Int
```
makingthematrix, line 13: I'd suggest to give a default value to any field which is not "id" or "key". Kotlinx will use those default values when it doesn't find real values in JSON strings. This will give us flexibility: both for unit tests, where we will be able to specify only a few fields instead of all of them, and in future migrations. 
Imagine we change the data class and add a field, but then we use a backup made from before the change was made. If the only difference between the two is a new field, and the field has a default value specified, no migration will be needed. 

To go the other way - to be able to ignore fields from old backups which we removed - we should set `JsonConfiguration` to
```
JsonConfiguration.Stable.copy(isLenient = true, ignoreUnknownKeys = true)
```

app/src/main/kotlin/com/waz/zclient/feature/backup/conversations/ConversationRolesBackupDao.kt
BradleyWire: I think this was a solution to overcome:

`A failure occurred while executing org.jetbrains.kotlin.gradle.internal.KaptExecution`

when using collections via inheritance on a DAO. But, I've optimised this by `@JvmSuppressWildcards` instead of a wrapper. Please check the new implementation.

PR #2958, opened 2020-08-04
Backup 11: List batch readers

makingthematrix:
## What's new in this PR?

I try to recreate in the new architecture the idea form the old one that data from tables can be written to files as proper json lists instead of manually, one json entry after another. To avoid OOM errors, for big tables we will create more than one file - in each file there will only maximum of `batchSize` entries.


app/src/main/kotlin/com/waz/zclient/feature/backup/usecase/CreateBackUpUseCase.kt
```diff
@@ -11,16 +11,15 @@ import kotlinx.coroutines.SupervisorJob
...
+
 class CreateBackUpUseCase(
-    private val backUpRepositories: List<BackUpRepository>,
+    private val backUpRepositories: List<BackUpRepository<File>>,
     private val coroutineScope: CoroutineScope = CoroutineScope(SupervisorJob() + Dispatchers.Default)
 ) : UseCase<Unit, Unit> {
 
-    override suspend fun run(params: Unit): Either<Failure, Unit> {
-        //TODO would be nice to log the actual exception somewhere
-        //TODO rollback changes if something goes wrong
```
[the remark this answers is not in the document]
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/core/extension/Either.kt
```diff
@@ -12,3 +12,28 @@ suspend fun <L, R, T> Either<L, R>.foldSuspendable(fnL: suspend (L) -> T?, fnR:
...
     }
+
+/**
+ * Maps over an iterable of T where the mapping function returns an Either<L, R>.
+ * If the mapping function returns Left<L>, the whole mapRight is interrupted and
+ * the left value is returned. If the mapping function returns Right for all elements, the result
+ * is Right<List<R>>. You can think of this method as a functional version of a map wrapped in
+ * a try/catch because the mapping function can throw an exception.
+ */
+fun <T, L, R> Iterable<T>.mapRight(fn: (T) -> Either<L, R>): Either<L, List<R>> {
```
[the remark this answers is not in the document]
makingthematrix: This is the version that can be used for any `Iterable`. But BatchReader is not an Iterable, so it needs its own version. Anyway, I'd like to add this more generic version too as I think it can be valuable in the codebase. I'll write unit tests for both.

app/src/main/kotlin/com/waz/zclient/feature/backup/di/BackUpModule.kt
```diff
@@ -23,6 +23,10 @@ import org.koin.dsl.module
 const val KEY_VALUES_FILE_NAME = "KeyValues"
 const val MESSAGES_FILE_NAME = "Messages"
 
+const val BATCH_SIZE = 1000
```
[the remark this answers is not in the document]
makingthematrix: ok

app/src/test/kotlin/com/waz/zclient/feature/backup/BackUpDataSourceTest.kt
```diff
@@ -57,8 +59,8 @@ class BackUpDataSourceTest : UnitTest() {
         runBlocking {
             databaseBatchReader.mockNextItems(entitiesList)
             `when`(databaseLocalDataSource.readIterator()).thenReturn(databaseBatchReader)
-            entitiesList.forEachIndexed { index, i ->
-                `when`(mapper.fromEntity(i)).thenReturn(modelsList[index])
+            entitiesList[0].forEachIndexed { index, i ->
```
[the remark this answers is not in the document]
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/feature/backup/di/BackUpModule.kt
```diff
@@ -31,17 +35,17 @@ val backUpModule = module {
 
     // KeyValues
     factory { KeyValuesBackUpDao(get()) }
-    factory { SingleReadDatabaseIOHandler<KeyValuesEntity>(get()) }
+    factory { BatchDatabaseIOHandler<KeyValuesEntity>(get(), batchSize = BATCH_SIZE) }
```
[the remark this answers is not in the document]
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/feature/backup/io/BatchReader.kt
```diff
@@ -24,18 +25,30 @@ interface BatchReader<T> {
...
-            action(it!!).foldSuspendable({
-                next = Either.Left(it)
-            }) {
-                next = readNext()
-                Unit
+/**
+ * A bit more specific version of [[com.waz.zclient.core.extension.mapRight]].
+ * BatchReader does not extend [[Iterable]] and can't use the generic mapRight.
+ */
+@SuppressWarnings("NestedBlockDepth")
```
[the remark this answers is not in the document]
makingthematrix: I think detekt is too strict with this, but ok. I created a nested method from a part of the inside `when(...) { is Right -> ... }`.

app/src/main/kotlin/com/waz/zclient/core/extension/Either.kt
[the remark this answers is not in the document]
android10: I would ask for more arguments. I think that could be useful is not valid for me. What are the use cases?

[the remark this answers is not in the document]
makingthematrix: Imagine any situation where we map a collection, but the mapping operation may end up with an error. Since we want to use `Either` instead of exceptions, the standard `.map` method won't stop at the error, but will produce a `List<Either<Error, Result>>` for all original elements and only after that you be able to go through the list and check if there was an error. Sometimes this is what you want to do, but sometimes you are interested only in getting the results if there was no error. In such case, using `.map` can be a waste of time, because it will go though the whole original collection anyway, perform the mapping operation for every element, and only then results of those operations will be dismissed. `.mapRight` is an optimization of this. It stops at the first error, and if the whole process is successful, then you get the results in `Either.Right` - no need for additional mapping of `List<Either<Error, Result>>` into `List<Result>`.

app/src/main/kotlin/com/waz/zclient/feature/backup/di/BackUpModule.kt
```diff
@@ -23,6 +23,10 @@ import org.koin.dsl.module
 const val KEY_VALUES_FILE_NAME = "KeyValues"
 const val MESSAGES_FILE_NAME = "Messages"
 
+const val BATCH_SIZE = 1000
+
+val targetDirForBackupFiles = Environment.getExternalStorageDirectory()
```
mejdoo, line 28: you can declare it as a singleton inside the backUpModule
`single { Environment.getExternalStorageDirectory() }`

app/src/main/kotlin/com/waz/zclient/feature/backup/di/BackUpModule.kt
```diff
@@ -31,17 +35,17 @@ val backUpModule = module {
 
     // KeyValues
     factory { KeyValuesBackUpDao(get()) }
-    factory { SingleReadDatabaseIOHandler<KeyValuesEntity>(get()) }
+    factory { BatchDatabaseIOHandler<KeyValuesEntity>(get(), batchSize = BATCH_SIZE) }
     factory { JsonConverter(KeyValuesBackUpModel.serializer()) } //TODO check if koin can resolve generics. use named parameters otherwise.
-    factory { BackUpFileIOHandler<KeyValuesBackUpModel>(KEY_VALUES_FILE_NAME, get()) }
+    factory { BackUpFileIOHandler<KeyValuesBackUpModel>(KEY_VALUES_FILE_NAME, get(), targetDirForBackupFiles) }
```
mejdoo, line 40: and change this to 
`  factory { BackUpFileIOHandler<KeyValuesBackUpModel>(KEY_VALUES_FILE_NAME, get(), get()) }`

app/src/main/kotlin/com/waz/zclient/feature/backup/di/BackUpModule.kt
```diff
@@ -31,17 +35,17 @@ val backUpModule = module {
...
     factory { KeyValuesBackUpMapper() }
     factory { KeyValuesBackUpDataSource(get(), get(), get()) } bind BackUpRepository::class
 
     // Messages
     factory { MessagesBackUpDao(get()) }
-    factory { BatchDatabaseIOHandler<MessagesEntity>(get(), batchSize = 20) }
+    factory { BatchDatabaseIOHandler<MessagesEntity>(get(), batchSize = BATCH_SIZE) }
     factory { JsonConverter(MessagesBackUpModel.serializer()) }
-    factory { BackUpFileIOHandler<MessagesBackUpModel>(MESSAGES_FILE_NAME, get()) }
+    factory { BackUpFileIOHandler<MessagesBackUpModel>(MESSAGES_FILE_NAME, get(), targetDirForBackupFiles) }
```
mejdoo, line 48: and this to `    factory { BackUpFileIOHandler<MessagesBackUpModel>(MESSAGES_FILE_NAME, get(), get()) }`

app/src/main/kotlin/com/waz/zclient/feature/backup/di/BackUpModule.kt
makingthematrix: ok

app/src/main/kotlin/com/waz/zclient/feature/backup/di/BackUpModule.kt
makingthematrix: ok.

app/src/main/kotlin/com/waz/zclient/feature/backup/di/BackUpModule.kt
makingthematrix: ok

makingthematrix:
> You are adding a core functionality: Iterable. Please test it properly.

Unit tests added.

app/src/test/kotlin/com/waz/zclient/feature/backup/io/file/BackUpFileIOHandlerTest.kt
```diff
@@ -19,69 +19,127 @@ class BackUpFileIOHandlerTest : UnitTest() {
...
+
+                batchReader.mockNextItems(listOf(listOf(1, 2, 3)))
+                `when`(jsonConverter.toJsonList(listOf(1, 2, 3))).thenReturn(jsonStr)
+
+                val tempDir = createTempDir()
+                backUpFileIOHandler = BackUpFileIOHandler(name, jsonConverter, tempDir)
+                backUpFileIOHandler.write(batchReader)
+                val createdFiles = getFilesWithPrefix(tempDir, name)
+                assertEquals(1, createdFiles.size)
+                with(createdFiles[0]) {
```
BradleyWire, line 46: We can use `also` here too to make it a little more readable: 

```
getFilesWithPrefix(tempDir, name).also {
    assertEquals(1, it.size)
    assertEquals(jsonStr, it.first().readText())
}
```

app/src/test/kotlin/com/waz/zclient/feature/backup/io/database/BatchDatabaseIOHandlerTest.kt
```diff
@@ -10,68 +10,70 @@ import com.waz.zclient.feature.backup.io.forEach
...
 
     @Test
     fun `given a batchReadableDao, when readIterator() is called, returns an iterator which reads in batches`() {
-        val allItems = mutableListOf(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12)
+        val allItems = listOf(1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12)
+        val itemsInBatches = listOf(listOf(1, 2, 3), listOf(4, 5, 6), listOf(7, 8, 9), listOf(10, 11, 12))
         val batchSize = 3
 
-        batchReadableDao = spy(batchReadableDaoOf(allItems))
+        batchReadableDao = Mockito.spy(batchReadableDaoOf(allItems))
```
BradleyWire, line 36: We can import the `Mockito` functions directly for consistency

app/src/test/kotlin/com/waz/zclient/feature/backup/io/file/BackUpFileIOHandlerTest.kt
```diff
@@ -19,69 +19,127 @@ class BackUpFileIOHandlerTest : UnitTest() {
...
+                val tempDir = createTempDir()
+
+                `when`(jsonConverter.toJsonList(listOf(1, 2, 3))).thenReturn(jsonStr1)
+                `when`(jsonConverter.toJsonList(listOf(4, 5, 6))).thenReturn(jsonStr2)
+
+                backUpFileIOHandler = BackUpFileIOHandler(name, jsonConverter, tempDir)
+
+                batchReader.mockNextItems(listOf(listOf(1, 2, 3)))
+                backUpFileIOHandler.write(batchReader)
+                val createdFiles1 = getFilesWithPrefix(tempDir, name)
```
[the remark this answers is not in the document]
BradleyWire: A question comes to mind when reviewing this also, why do we only rely on the first item of the list? Will there never be more than one object in a list? If so, what value does it add making everything a list?

app/src/test/kotlin/com/waz/zclient/feature/backup/io/database/BatchDatabaseIOHandlerTest.kt
makingthematrix: This is my AndroidStudio having its own ideas. I corrected it.
There is at one line a call to `Mockito.any()` which can't be imported and used as `any()` because then it's ambiguous.  I let my AndroidStudio fix it automatically and it added `Mockito.` to all other calls as well.

app/src/test/kotlin/com/waz/zclient/feature/backup/io/file/BackUpFileIOHandlerTest.kt
makingthematrix: ok

app/src/test/kotlin/com/waz/zclient/feature/backup/io/file/BackUpFileIOHandlerTest.kt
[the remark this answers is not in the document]
makingthematrix: There will be more objects. It's just that this test is about something else. It's about a rare case when we write a file over an already existing file. (Usually, the old files will be deleted).

I added a new test where two batches are used and two files are created.

app/src/main/kotlin/com/waz/zclient/feature/backup/BackUpDataSource.kt
```diff
@@ -4,29 +4,34 @@ import com.waz.zclient.core.exception.Failure
...
 
-    override suspend fun saveBackup(): Either<Failure, Unit> {
+    override suspend fun saveBackup(): Either<Failure, List<File>> {
         val readIterator = databaseLocalDataSource.readIterator()
-        val writeIterator: BatchReader<T> = object : BatchReader<T> {
-            override suspend fun readNext(): Either<Failure, T?> = readIterator.readNext().map {
-                it?.let { mapper.fromEntity(it) }
+        val writeIterator: BatchReader<List<T>> = object : BatchReader<List<T>> {
+            override suspend fun readNext(): Either<Failure, List<T>?> = readIterator.readNext().map {
+                it?.map { mapper.fromEntity(it) }
```
BradleyWire, line 18: When I branched from this I noticed some shadowed `it` naming. We normally advise the outer `it` reference be named. I'll fix these in my PR though to not block you any further.

app/src/main/kotlin/com/waz/zclient/feature/backup/io/BatchReader.kt
```diff
@@ -24,18 +25,28 @@ interface BatchReader<T> {
...
+
+/**
+ * A bit more specific version of [[com.waz.zclient.core.extension.mapRight]].
+ * BatchReader does not extend [[Iterable]] and can't use the generic mapRight.
+ */
+suspend fun <T, R> BatchReader<T>.mapRight(action: suspend (T) -> Either<Failure, R>): Either<Failure, List<R>> {
+    val rightValues = mutableListOf<R>()
+    var failure: Failure? = null
+
+    suspend fun performAction(value: T) =
```
BradleyWire, line 38: For me this is much more readable, so i'd agree with detekt here

app/src/main/kotlin/com/waz/zclient/core/extension/Either.kt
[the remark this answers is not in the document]
android10: I know you want to turn much of this stuff into FP because this is the way you like to do things (it is totally fine because we all have our opinions). I also have to say I'm interested in a real use case and not potential use cases. 

This is up to you and the team. You are the ones deciding what goes in and out. I'm not a big fan of this because from my perspective all this compromises readability and makes the code more complex.  

I'm personally skeptical to introduce too much of FP (due to bad experiences), I think what Kotlin provides should be enough. We even introduced Flow which is a reactive framework with reactive operators if we want to use it. 

I would like to involve everyone in this kind of discussions: @mejdoo @BradleyWire @gizemb. 

These are my 2 cents. I will be neutral on this decision and leave it up to you, as soon as everyone is comfortable and  confident that this is the way to go.

app/src/main/kotlin/com/waz/zclient/feature/backup/BackUpDataSource.kt
makingthematrix: @BradleyWire : ok, I'll remember. So no changes are needed here for now?

BradleyWire: Nope, just a note really

app/src/main/kotlin/com/waz/zclient/core/extension/Either.kt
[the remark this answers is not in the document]
BradleyWire: For me, I don't see an immediate need for this but i'd be more than happy for this to be introduced if it was being used by something concrete (in a future PR maybe). Right now it looks like an unused duplicate, so maybe let's hold off? Happy to get this into a collective discussion when @gizemb is back.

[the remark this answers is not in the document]
makingthematrix: OK. I removed the generic `mapRight` and its tests from this PR.

makingthematrix:
build again

makingthematrix:
build again

PR #2990, opened 2020-08-18
New backups (17+2e): Encryption

makingthematrix:
EncryptionHandler restored as a new PR.

I rewrote this code as closesy as possible from Scala. Please take a look at com.waz.service.backup.EncryptionBackupHeader and com.waz.utils.crypto.LibSodiumUtils. 
We know that code works, but it's complex and hard to read. As for this rewritten version, first I'd like to
write some tests (or test it manually).
I'd suggest not to refactor it until we are sure our changes won't break anything.

### DIFF
--- a/app/src/main/kotlin/com/waz/zclient/feature/backup/encryption/EncryptionHandlerDataSource.kt
+++ b/app/src/main/kotlin/com/waz/zclient/feature/backup/encryption/EncryptionHandlerDataSource.kt
@@ -1,0 +1,300 @@
+package com.waz.zclient.feature.backup.encryption
+
+import com.waz.model.UserId
+import com.waz.zclient.core.exception.Failure
+import com.waz.zclient.core.exception.FeatureFailure
+import com.waz.zclient.core.exception.GenericUseCaseError
+import com.waz.zclient.core.exception.IOFailure
+import com.waz.zclient.core.functional.Either
+import com.waz.zclient.core.functional.Either.Left
+import com.waz.zclient.core.functional.Either.Right
+import com.waz.zclient.core.functional.flatMap
+import com.waz.zclient.core.functional.map
+import com.waz.zclient.core.logging.Logger.Companion.error
+import com.waz.zclient.core.logging.Logger.Companion.warn
+import com.waz.zclient.core.logging.Logger.Companion.verbose
+import org.libsodium.jni.NaCl
+import org.libsodium.jni.Sodium
+import java.io.File
+import java.io.IOException
+import java.nio.ByteBuffer
+import java.security.SecureRandom
+
+class EncryptionHandlerDataSource : EncryptionHandler {
+    override fun encrypt(backupFile: File, userId: UserId, password: String): Either<Failure, File> =
+        try {
+            loadLibrary
+
+            val salt = generateSalt()
+
+            getMetaDataBytes(salt, userId).flatMap { meta ->
+                val backupBytes = backupFile.readBytes()
+                encrypt(backupBytes, password, salt).map { encryptedBytes ->
+                    val encryptedFile = File(backupFile.parentFile, backupFile.name + "_encrypted").apply {
+                        writeBytes(meta)
+                        writeBytes(encryptedBytes)
+                    }
+                    encryptedFile
+                }
+            }
+        } catch (ex: IOException) {
+            Left(IOFailure(ex))
+        }
+
+    override fun decrypt(backupFile: File, userId: UserId, password: String): Either<Failure, File> {
+        loadLibrary
+
+        val metadata = EncryptedBackupHeader.readEncryptedMetadata(backupFile)
+        return if (metadata != null) {
+            val hash = hash(userId.str(), metadata.salt)
+            if (hash != null && hash.contentEquals(metadata.uuidHash)) {
+                val encryptedBackupBytes = ByteArray(EncryptedBackupHeader.totalHeaderLength)
+                backupFile.inputStream().buffered().read(encryptedBackupBytes)
+                decrypt(encryptedBackupBytes, password, metadata.salt).map { decryptedBackupBytes ->
+                    File.createTempFile("wire_backup", ".zip").apply { writeBytes(decryptedBackupBytes) }
+                }
+            } else if (hash != null) {
+                Left(EncryptionFailure("Uuid hashes don't match"))
+            } else {
+                Left(EncryptionFailure("Uuid hashing failed"))
+            }
+        } else {
+            Left(EncryptionFailure("metadata could not be read"))
+        }
+    }
+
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/com/waz/zclient/feature/backup/encryption/EncryptionHandlerDataSource.kt
```diff
@@ -0,0 +1,300 @@
...
+                }
+            }
+        } catch (ex: IOException) {
+            Left(IOFailure(ex))
+        }
+
+    override fun decrypt(backupFile: File, userId: UserId, password: String): Either<Failure, File> {
+        loadLibrary
+
+        val metadata = EncryptedBackupHeader.readEncryptedMetadata(backupFile)
```
BradleyWire, line 47: We'll need to avoid these static calls if we want to properly unit test this. The `EncryptedBackupHeader` does alot considering it's a `data` class, we can probably move this function out of it (and the others based on the comments i've made already)

app/src/main/kotlin/com/waz/zclient/feature/backup/encryption/EncryptionHandlerDataSource.kt
```diff
@@ -0,0 +1,300 @@
...
+    private fun initPush(key: ByteArray, header: ByteArray): ByteArray? = initializeState(key, header) {
+        s: ByteArray, h: ByteArray, k: ByteArray -> Sodium.crypto_secretstream_xchacha20poly1305_init_push(s, h, k)
+    }
+
+    private fun initPull(key: ByteArray, header: ByteArray): ByteArray? = initializeState(key, header) {
+        s: ByteArray, h: ByteArray, k: ByteArray -> Sodium.crypto_secretstream_xchacha20poly1305_init_pull(s, h, k)
+    }
+
+    companion object {
+        data class EncryptionFailure(val msg: String) : FeatureFailure()
```
BradleyWire, line 205: This doesn't need to belong to the `companion object `

app/src/main/kotlin/com/waz/zclient/feature/backup/encryption/EncryptionHandlerDataSource.kt
```diff
@@ -0,0 +1,300 @@
...
+            }
+        } catch (ex: IOException) {
+            Left(IOFailure(ex))
+        }
+
+    override fun decrypt(backupFile: File, userId: UserId, password: String): Either<Failure, File> {
+        loadLibrary
+
+        val metadata = EncryptedBackupHeader.readEncryptedMetadata(backupFile)
+        return if (metadata != null) {
```
BradleyWire, line 48: This can be (maybe even broken down into private functions even more):

```
 override fun decrypt(backupFile: File, userId: UserId, password: String): Either<Failure, File> {
        loadLibrary
        return EncryptedBackupHeader.readEncryptedMetadata(backupFile)?.let { metaData ->
            hash(userId.str(), metaData.salt)?.let {
                when (it.contentEquals(metaData.uuidHash)) {
                    true -> {
                        val encryptedBackupBytes = ByteArray(EncryptedBackupHeader.totalHeaderLength)
                        backupFile.inputStream().buffered().read(encryptedBackupBytes)
                        decrypt(encryptedBackupBytes, password, metaData.salt).map { decryptedBackupBytes ->
                            File.createTempFile("wire_backup", ".zip").apply { writeBytes(decryptedBackupBytes) }
                        }
                    }
                    false -> Left(EncryptionFailure("Uuid hashes don't match"))
                }
            } ?: Left(EncryptionFailure("Uuid hashing failed"))
        } ?: Left(EncryptionFailure("metadata could not be read"))
    }

```

app/src/main/kotlin/com/waz/zclient/feature/backup/encryption/EncryptionHandlerDataSource.kt
```diff
@@ -0,0 +1,300 @@
...
+                val backupBytes = backupFile.readBytes()
+                encrypt(backupBytes, password, salt).map { encryptedBytes ->
+                    val encryptedFile = File(backupFile.parentFile, backupFile.name + "_encrypted").apply {
+                        writeBytes(meta)
+                        writeBytes(encryptedBytes)
+                    }
+                    encryptedFile
+                }
+            }
+        } catch (ex: IOException) {
```
BradleyWire, line 40: Where are we catching this `IOException` from?

app/src/main/kotlin/com/waz/zclient/feature/backup/encryption/EncryptionHandlerDataSource.kt
```diff
@@ -0,0 +1,300 @@
...
+    val opsLimit: Int,
+    val memLimit: Int
+) {
+    companion object {
+        const val TAG = "EncryptedBackupHeader"
+
+        private const val androidMagicNumber: String = "WBUA"
+        const val currentVersion: Short = 2
+        private const val saltLength = 16
+        const val uuidHashLength = 32
```
BradleyWire, line 245: This can live inside the `EncryptionHandlerDataSource` class

BradleyWire: same with the other constants only used inside that class

BradleyWire (changes requested):
Some key points for the Encryption class: 

1. It should only be decryption / encryption logic 
2. `Sodium` specific logic should be placed in a collaborator you can inject 
3. `EncryptionBackupHeader` does too much, we can break it down into multiple classes 
4.  More idiomatic ways to writing this in Kotlin without refactoring the logic (I haven't addressed them all with these comments)

app/src/main/kotlin/com/waz/zclient/feature/backup/encryption/EncryptionHandlerDataSource.kt
```diff
@@ -0,0 +1,300 @@
...
+import com.waz.zclient.core.logging.Logger.Companion.warn
+import com.waz.zclient.core.logging.Logger.Companion.verbose
+import org.libsodium.jni.NaCl
+import org.libsodium.jni.Sodium
+import java.io.File
+import java.io.IOException
+import java.nio.ByteBuffer
+import java.security.SecureRandom
+
+class EncryptionHandlerDataSource : EncryptionHandler {
```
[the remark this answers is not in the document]
mejdoo: Regarding the naming , renaming this class to **"Encryptor"** makes more sense to me. We have to think about this as a separate component for encryption and decryption.

mejdoo (changes requested):
Totally agree with @BradleyWire the encryption stuff is too complex and must be simplified by moving `Sodium` logic outside.

PR #2577, opened 2020-01-29
A fix for backup restoration

makingthematrix:
Fixes https://wearezeta.atlassian.net/browse/AN-6604

Restoring backup means replacing the database files with the ones from the backup. Currently we don't take into account that the app might use the database in the same time for initialization of other components. This PR wraps the backup restore operation in a transaction, makes the account manager initialization lazy, and moves some method calls around in order to make sure that nothing stands in the way of replacing the files.

Also, I fixed the password dialog appearing on restoration. Previously it had the same title and message as the backup *creation* popup, which was very confusing. Now it uses the same title as iOS and the message is empty. 
And I made the password required. It was optional which was against the specs. I maintained the check of restoration if the password is empty, to still have backward compatibility if anyone made a non-encrypted backup in the past (plus some unit tests require it). But all new backups have to be encrypted with the password.

#### APK
[Download build #1087](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1087/artifact/build/artifact/wire-dev-PR2577-1087.apk)
[Download build #1152](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1152/artifact/build/artifact/wire-dev-PR2577-1152.apk)
[Download build #1178](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1178/artifact/build/artifact/wire-dev-PR2577-1178.apk)
[Download build #1185](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1185/artifact/build/artifact/wire-dev-PR2577-1185.apk)
[Download build #1251](http://10.10.124.11:8080/job/Pull%20Request%20Builder/1251/artifact/build/artifact/wire-dev-PR2577-1251.apk)

PR #2993, opened 2020-08-19
No unencrypted traffic to Wire

franziskuskiefer:
While Android < 9 is still supported something like this should be added (ideally for all URLs, not just Wire).
Also see https://developer.android.com/training/articles/security-config#CleartextTrafficPermitted.
#### APK
[Download build #2577](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2577/artifact/build/artifact/wire-dev-PR2993-2577.apk)

### DIFF
--- a/app/src/main/res/xml/network_security_config.xml
+++ b/app/src/main/res/xml/network_security_config.xml
@@ -19,6 +19,10 @@
 
 -->
 <network-security-config>
+    <domain-config cleartextTrafficPermitted="false">
+        <domain includeSubdomains="true">wire.com</domain>
+        <domain includeSubdomains="true">zinfra.io</domain>
+    </domain-config>
     <debug-overrides>
         <trust-anchors>
             <!-- Trust user added CAs while debuggable only -->


mejdoo (approved):

app/src/main/res/xml/network_security_config.xml
```diff
@@ -19,6 +19,10 @@
 
 -->
 <network-security-config>
+    <domain-config cleartextTrafficPermitted="false">
```
mejdoo, line 22: Good catch @franziskuskiefer. I wonder how insure this as well for custom backends.

mejdoo (approved):

app/src/main/res/xml/network_security_config.xml
franziskuskiefer: I don't see a way to do this in general for all domains. So we'd have to add each custom backend here.

gizemb:
@franziskuskiefer please open an issue in JIRA with steps to test this & other comments (such as custom backend issue).

app/src/main/res/xml/network_security_config.xml
BradleyWire: @franziskuskiefer is this config mandatory? Could we not achieve the same thing by removing `ConnectionSpec.CLEARTEXT` from the `ConnectionSpecs` field when we build a HTTP request? That may be more dynamic for custom backends too?

franziskuskiefer: Sure, you should do that anyway. But this config makes sure that no one can add any clear text requests.

marcoconti83:
Do I understand correctly that this will block any HTTP (as opposed to HTTPS) from the app? We do allow opening arbitrary hosts for link previews, if we block HTTP, all HTTP websites will stop working with link previews.
It could be a valid product decision, but then we need to figure out a way to explain to the users why some websites work and some don't.

franziskuskiefer:
> Do I understand correctly that this will block any HTTP (as opposed to HTTPS) from the app? We do allow opening arbitrary hosts for link previews, if we block HTTP, all HTTP websites will stop working with link previews.
> It could be a valid product decision, but then we need to figure out a way to explain to the users why some websites work and some don't.

No, this is only blocking HTTP traffic to the specified domains.
Generally we shouldn't make HTTP requests for link previews though. But that's a separate issue.
Note however that Android 9+ blocks all HTTP traffic anyway. So that this is already the case for modern Android devices.

makingthematrix:
@marcoconti83 , @franziskuskiefer : Was any decision about it made? Technically the PR is ok.

franziskuskiefer:
@marcoconti83 is out for two weeks.
With the small set of URLs we currently have in here I'd recommend to merge this as this shouldn't change anything and is only an additional safety net. I filed an issue to look into link previews separately.

app/src/main/res/xml/network_security_config.xml
android10: @BradleyWire that is also another idea. If you want to do it more dynamic, keep in mind that we can always override resources for custom builds. Whatever solution we pick is fine from my perspective. 

In any case thanks @franziskuskiefer for the contribution. :heart:

android10 (approved):
:+1:

BradleyWire (approved):

PR #3027, opened 2020-09-18
Refactoring WSPushService + other changes required to make wire-android work with wire-signals 0.2.2

makingthematrix:
In `wire-signals 0.2.2`, I added methods `EventStream.from(future: Future)` and `EventStream.from(future: CancellableFuture)` which I wanted to use in `WSPushService`, becase until now to make an event stream out of a future this class used `EventStream.from(Signal.from(future: CancellableFuture))` which created an unnecessary signal in the middle. But while I worked on that, I noticed in the logs that WebSocket is opened and closed multiple times when only one operation was sufficient. I rewrote that code. I also experienced a bug when the WebSocket stopped responding after switcihng between two accounts several times, but I can't be sure if it was there before. Now I can't reproduce it anymore.

For information on what's included in `wire-signals 0.2.2`, please check here: https://github.com/wireapp/wire-signals/pull/5

#### APK
[Download build #2777](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2777/artifact/build/artifact/wire-dev-PR3027-2777.apk)

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/push/WSPushService.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/push/WSPushService.scala
@@ -21,7 +21,6 @@ import java.net.URL
 import java.util.concurrent.atomic.AtomicInteger
 
 import com.waz.log.LogSE._
-import com.waz.api.impl.ErrorResponse
 import com.waz.log.BasicLogging.LogTag
 import com.waz.model.UserId
 import com.waz.model.otr.ClientId
@@ -39,6 +38,7 @@ import com.waz.znet2.WebSocketFactory.SocketEvent
 import com.waz.znet2.http.{Body, Method, Request}
 import com.waz.znet2.{WebSocket, WebSocketFactory, http}
 
+import scala.concurrent.Future
 import scala.concurrent.duration._
 import scala.util.Left
 
@@ -85,8 +85,6 @@ object WSPushServiceImpl {
       ExponentialBackoff.standardBackoff
     )(ev)
   }
-
-
 }
 
 class WSPushServiceImpl(userId:              UserId,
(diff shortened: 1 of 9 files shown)


wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/push/WSPushService.scala
```diff
@@ -102,59 +100,66 @@ class WSPushServiceImpl(userId:              UserId,
...
+      case SocketEvent.Closed(_, Some(errorResponse)) =>
         error(l"WebSocket closed with error: $errorResponse")
-        connected ! false
+        deactivate()
         activate(initialDelay = backoff.delay(retryCount.incrementAndGet()))
-      case Right(SocketEvent.Closed(_, _)) =>
-        connected ! false
+      case SocketEvent.Closed(_, _) =>
+        verbose(l"WebSocket closed")
+        deactivate()
```
makingthematrix, line 156: As I understand, it's possible that we will not receive a "closing" event, but we will get directly the "closed" one. So we can't rely on that WebSocket will be deactivated in "closing".

makingthematrix: To be honest, that makes me suspicious if we can get a delayed "closed" event after the WebSocket has been already reopened...

makingthematrix:
build again

wire-android-sync-engine/zmessaging/src/main/scala/com/waz/service/push/WSPushService.scala
```diff
@@ -102,59 +100,66 @@ class WSPushServiceImpl(userId:              UserId,
...
-      accessTokenResult <- EventStream.from(Signal.from(accessTokenProvider.currentToken()))
-      event             <- accessTokenResult match {
-                             case Right(token)        => webSocketFactory.openWebSocket(requestCreator(token)).map(Right.apply)
-                             case Left(errorResponse) => EventStream.from(Signal.const(Left(errorResponse)))
-                           }
-    } yield event
-
-    events.on(dispatcher) {
-      case Left(errorResponse) =>
-        error(l"Error while access token receiving: $errorResponse")
```
makingthematrix, line 136: Until now, this part was called if the event stream was constructed with success, and then we received an error response when opening the WebSocket. But if the construction failed for whatever reason, it wasn't handled at all. Now, this part of the code works in both cases.

android10 (approved):
:+1:

BradleyWire (approved):

PR #3036, opened 2020-09-28
BI spec changes

CalumMcCall:
### Issues

This PR introduces some fixes that are due to changes in the BI spec, and some bug fixes.

The countly service is now pointed at Wire's hosted instance, using the demo and internal app keys for build where appropriate.

The `app.open` event is now only sent once, and the init is only performed once, per launch.

I also noticed that the old method of setting the `TrackingId` was resulting in new id every launch sometimes. So I switched to a more manual method which should be more reliable.

### Testing

Testing was done manually.
#### APK
[Download build #2791](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2791/artifact/build/artifact/wire-dev-PR3036-2791.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/MainActivity.scala
+++ b/app/src/main/scala/com/waz/zclient/MainActivity.scala
@@ -32,7 +32,6 @@ import com.waz.model._
 import com.waz.service.AccountManager.ClientRegistrationState.{LimitReached, PasswordMissing, Registered, Unregistered}
 import com.waz.service.AccountsService.UserInitiated
 import com.waz.service.ZMessaging.clock
-import com.waz.service.tracking.TrackingService
 import com.waz.service.{AccountManager, AccountsService, ZMessaging}
 import com.waz.threading.Threading
 import com.waz.threading.Threading._
@@ -275,8 +274,6 @@ class MainActivity extends BaseActivity
     for {
       _      <- initTracking
       _      <- inject[GlobalTrackingController].start(this)
-      userId <- inject[Signal[ZMessaging]].head.map(_.selfUserId)
-      _      <- inject[TrackingService].appOpen(userId)
     } yield ()
 
     val intent = getIntent
--- a/app/src/main/scala/com/waz/zclient/MainActivity.scala
+++ b/app/src/main/scala/com/waz/zclient/MainActivity.scala
@@ -246,6 +246,9 @@ class MainActivity extends BaseActivity
   private def initTracking: Future[Unit] =
     for {
       prefs            <- userPreferences.head
+      id               <- prefs.preference(CountlyTrackingId).apply()
+      _                <-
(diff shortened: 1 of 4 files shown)


AndroidBob:

BradleyWire:
ok to test

mejdoo (approved):

android10 (approved):
:+1:

app/src/main/scala/com/waz/zclient/MainActivity.scala
```diff
@@ -246,6 +245,9 @@ class MainActivity extends BaseActivity
   private def initTracking: Future[Unit] =
     for {
       prefs            <- userPreferences.head
+      id               <- prefs.preference(CountlyTrackingId).apply()
+      _                <-
+        if(id.isEmpty) prefs.preference(CountlyTrackingId) := Some(TrackingId()) else Future.successful(())
```
BradleyWire, line 250: Not urgent but this isn't formatted correctly regarding the brackets, it should be `if (`

app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
```diff
@@ -98,16 +108,16 @@ class GlobalTrackingController(implicit inj: Injector, cxt: WireContext, eventCo
   private def setUserDataFields(): Future[Unit] = {
     for {
       Some(z)         <- accountsService.activeZms.head
-      teamMember      = z.teamId.isDefined
+      teamId          =  z.teamId.getOrElse(TeamId("n/a"))
```
BradleyWire, line 111: Does "n/a" trigger anything particular here? Could we not use `TeamId.Empty` for this?

app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
```diff
@@ -45,22 +46,31 @@ class GlobalTrackingController(implicit inj: Injector, cxt: WireContext, eventCo
...
+    accountsService.activeAccount.foreach(_.foreach(user => tracking.appOpen(user.id)))
+  }
+
   def init(): Future[Unit] = {
     for {
-      ap          <- tracking.isTrackingEnabled.head if(ap)
-      trackingId  <- am.head.flatMap(_.storage.userPrefs(CountlyTrackingId).apply())
-      logsEnabled <- inject[LogsService].logsEnabled
+      ap               <- tracking.isTrackingEnabled.head if(ap)
+      inited           <- initialized.head if(!inited)
```
BradleyWire, line 59: formatting here too

BradleyWire (approved):

mejdoo:
@BradleyWire  I will rebase the main PR and fix the minor requests

makingthematrix (approved):

Issue #3002, opened 2020-08-25
Android 11 (R) Breaks All Transitions, Causing UI to be Unreadable


dmz38:
**Describe the bug**
After updating to Android 11, Wire's transitions cause the application to be unreadable after certain actions.

- Scrolling on the contact list on the homepage
- Closing a conversation and going back to the main contact list.

All of these actions will result in the text to "train" with the scrolling instead of moving in place.  For security reasons, I have included only a partial screenshot below, but I've been able to reproduce it on every Android beta device I have.  

**To Reproduce**
Home Page:
1. Open the app
2. Scroll the contract list

Closing the conversation and going back to the home screen:
1. Hit the back arrow or use the back gesture.

**Expected behavior**
The text should still be readable and not act like someone dragged the text across the screen.

**Screenshots**
![Screenshot_20200825-055556~2](https://user-images.githubusercontent.com/6808885/91161585-ecc56b80-e698-11ea-8be2-b10ef5d1e422.png)

**Smartphone (please complete the following information):**
- Wire version that exhibits the issue: 3.53.934
- Last Wire version that did not exhibit the issue (if applicable): None
- Android OS version used to run Wire: 11
- Is this a custom firmware or a stock one: Stock Android Beta Program
- Mobile phone model/manufacturer: Google Pixel 3, 4a
- Mobile network type (EDGE/LTE/Wi-Fi/Offline): All

**Additional context**
The Android beta program is available on all Pixel devices generation 2 and later.  Android 11 is expected to launch in September according to reports.  To me, this issue makes the application unstable and virtually unusable.  The only way to "fix" the issue is to close and open the app every time.

Thank you for your help in correcting this issue!

makingthematrix:
Hi,

Thanks for bringing this up.
We are currently busy working on a few big things and for sure we will not be ready for Android 11 when it comes out. We will try to address these issues during the next few months.

BradleyWire:
Just as an update, i'll be working on resolving Android 11 issues as high priority.  Will post updates as and when we have them.

dmz38:
Thanks so much for doing this!  I'm really glad to see a fix and some memory leak cleanups.

I saw this was just merged into the release branch.  Does that mean it'll be on the beta channel of the Google Play listing relatively soon?

BradleyWire:
[...]

PR #3051, opened 2020-10-09
Fix: Android 11 UI Distortion + memory leaks along the way 

BradleyWire:
## What's new in this PR?

JIRA ticket (1): https://wearezeta.atlassian.net/browse/AN-7108
JIRA ticket (2): https://wearezeta.atlassian.net/browse/AN-7100
Github: https://github.com/wireapp/wire-android/issues/3002

### Issues

Some Pixel devices are creating an unreadable system UI upon various different actions 

### Causes

It looks like a new reaction to an old problem, when we're setting the `getWindow.setBackgroundDrawable(new ColorDrawable(Color.TRANSPARENT))` the reaction by the system UI overloads the `MainActivity`. This I believe is because we invalidate and re-create the `PhoneWindow` with the new drawable and opacity.

### Solutions

Remove this call to `getWindow().setBackgroundDrawable(..)` and replace with a color resource rendered when the view has been attached and created. 

### Testing

How to reproduce without this fix: 

- Login as a user without a profile picture
- open a conversation
- come back from the conversation via back arrow or swiping from left to right
-> contact list is broken
- try swiping a conversation to the right
-> contact list gets even worse

You should be able to use the app as normal with this fix. 

## Notes

In order for me to find the problem I spent hours analysing memory dumps, GPU profiles and looking for potential memory leaks. So this PR also includes plugs to some of those leaks. 
#### APK
[Download build #2833](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2833/artifact/build/artifact/wire-dev-PR3051-2833.apk)
[Download build #2835](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2835/artifact/build/artifact/wire-dev-PR3051-2835.apk)
[Download build #2836](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2836/artifact/build/artifact/wire-dev-PR3051-2836.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/security/ActivityLifecycleCallback.scala
+++ b/app/src/main/scala/com/waz/zclient/security/ActivityLifecycleCallback.scala
@@ -30,6 +30,7 @@ import com.waz.zclient.{BuildConfig, Injectable, Injector, LaunchActivity}
 import com.wire.signals._
 
 import scala.collection.convert.DecorateAsScala
+import scala.util.Try
 
 class ActivityLifecycleCallback(implicit injector: Injector)
   extends Application.ActivityLifecycleCallbacks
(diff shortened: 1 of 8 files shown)


app/src/main/scala/com/waz/zclient/security/ActivityLifecycleCallback.scala
```diff
@@ -70,12 +72,13 @@ class ActivityLifecycleCallback(implicit injector: Injector)
   override def onActivityCreated(activity: Activity, bundle: Bundle): Unit = {}
 
   override def onActivityResumed(activity: Activity): Unit = {
+    val activityRef = new WeakReference[Activity](activity)
     if (BuildConfig.FORCE_HIDE_SCREEN_CONTENT) {
-      activity.getWindow.addFlags(FLAG_SECURE)
+      activityRef.get().getWindow.addFlags(FLAG_SECURE)
```
makingthematrix, line 77: Please wrap in `Option` to prevent NPE.

makingthematrix (changes requested):

makingthematrix (approved):

mejdoo (approved):

PR #3052, opened 2020-10-13
fix bug with UTF-16 encoding for chinese characters

kongpf8848:
fix bug for UTF-16 encoding mentions info before send message

### DIFF
--- a/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/model/MessageData.scala
+++ b/wire-android-sync-engine/zmessaging/src/main/scala/com/waz/model/MessageData.scala
@@ -555,14 +555,22 @@ object MessageData extends
   private val UTF_16_CHARSET  = Charset.forName("UTF-16")
 
   private def encode(text: String) = {
-    val bytes = UTF_16_CHARSET.encode(text).array
-
-    if (bytes.length < 3 || bytes.slice(2, bytes.length).forall(_ == 0))
+    if (text == null) {
       Array.empty[Byte]
-    else if (bytes(2) == 0)
-      bytes.slice(2, bytes.lastIndexWhere(_ > 0) + 1)
-    else
-      Array[Byte](0) ++ bytes.slice(2, bytes.lastIndexWhere(_ > 0) + 1)
+    }
+    else {
+      val bytes = UTF_16_CHARSET.encode(text).array
+      /**
+       * UTF-16BE,first two bytes with 0xFEFF
+       * UTF-16LE,first two bytes with 0xFFFE
+       */
+      if (bytes.length < 3 || bytes.slice(2, bytes.length).forall(_ == 0))
+        Array.empty[Byte]
+      else {
+        val index = (text.length + 1) * 2
+        bytes.slice(2, index)
+      }
+    }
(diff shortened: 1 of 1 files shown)


AndroidBob:

BradleyWire:
Can you please update the description to share more details about the bug? The following template will help guide you: 

**Describe the bug**:

**To Reproduce**:

**Expected behavior**: 

**Screenshots**

**Smartphone (please complete the following information):**
- Wire version that exhibits the issue: 
- Last Wire version that did not exhibit the issue (if applicable): 
- Android OS version used to run Wire: 
- Is this a custom firmware or a stock one: 
- Mobile phone model/manufacturer: 
- Mobile network type (EDGE/LTE/Wi-Fi/Offline): 

**Additional context**

Thank you for your help in correcting this issue!

makingthematrix:
The change doesn't look right. `text` should never be `null` anyway. And there was a reason why we checked if the first two bytes are `0`, although it looks like a hack... What's the way to reproduce the problem? Is it enough if I try to send a mention where a part of the text is in Chinese characters?

kongpf8848:
> Can you please update the description to share more details about the bug? The following template will help guide you:
> 
> **Describe the bug**:
> 
> **To Reproduce**:
> 
> **Expected behavior**:
> 
> **Screenshots**
> 
> **Smartphone (please complete the following information):**
> 
> * Wire version that exhibits the issue:
> * Last Wire version that did not exhibit the issue (if applicable):
> * Android OS version used to run Wire:
> * Is this a custom firmware or a stock one:
> * Mobile phone model/manufacturer:
> * Mobile network type (EDGE/LTE/Wi-Fi/Offline):
> 
> **Additional context**:
> 
> Thank you for your help in correcting this issue!
Wire 2020-10-13 at 10-21-09

Wire Version：3.55.939
Android OS Version：Android 9
Mobile：Samsung Galaxy S8/Samsung
Network：LTE

kongpf8848:
> hack
1.  I think for not empty string after UTF-16 encoding, first two bytes either 0xFFFE(UTF-16LE) or  0xFEFF(UTF-16BE),after encode, every byte has value possible large than 0 or small than 0,so use lastIndexWhere(_ > 0) for last Index is not correct in some case(eg, for chinese characters).
2. you can send a mention where a part of the text is in chinese characters to reproduce this problem.

gizemb (changes requested):
Could you please add a test case about this fix? I think adding another scenario to`Adjust mentions` feature in `MessageDataSpec` would suffice. This way, we could also see more clearly what was the problem and whether it's really fixed.

gizemb:
@kongpf8848 any chance you could add a test case? note that in case of an update you'd have to rebase this branch to `develop` since we've made some changes there for the build pipeline.

gizemb:
hi @kongpf8848 , I see that your forked branch isn't available anymore. we want to get this PR merged. would you be willing to help us by adding some tests?

gizemb:
I'm closing this PR as we couldn't get a response from the author for quite some time. I duplicated it in #3356 and will make sure the author's contributions are mentioned in release notes.

Issue #2865, opened 2020-05-28
Tapping a username or profile icon dosn't do anything


Transisto:
Expected behavior:

Jump to profile or DM.

That's what the desktop version is doing and every single chat app that had groups.

Wire is almost unusable combined with https://github.com/wireapp/wire-android/issues/2858
The only way to DM a user at the moment is to scroll down may many pages to find the user.

makingthematrix:
Thanks for the report. We're working on it.

makingthematrix:
@Transisto : Sorry, but what do you mean by "The only way to DM a user at the moment is to scroll down may many pages to find the user"?

Transisto:
I was wrong, we can find the users we wish to DM by typing their entire
name in the search.
Is that how its supposed to work?

makingthematrix:
@Transisto : I mean, I don't understand what screen are we talking about. The search? The conversation list? The conversation itself?

Transisto:
> Wire is almost unusable when this bug is combined with #2858

Did you read https://github.com/wireapp/wire-android/issues/2858 ?

makingthematrix:
@Transisto :  Ok, so if I understand you correctly, clicking on an entry in the search doesn't open the conversation for you. But I can't reproduce it.

Transisto:
No, I was pointing out that "user search" was broken and tapping a username
in a group chat could have been an alternative to DM this user.

This issue is simply about tapping a user inside a group chat to get to DM.

It works on desktop version but not android.

makingthematrix:
@Transisto : Then that's right. We don't have this feature. If I remember correctly, the decision was that on smaller screens it was too close to the "Like" button and we could have users "liking" messages by mistake.

Transisto:
Feel free to add the "Feature" tag, I've checked the UI and I agree the ❤ is close the the user icon circle, but it's not close the the Username itself.  Tapping the username to get to DM is how most other app works too.

PR #3057, opened 2020-10-16
Trying to overcome file sharing ban with Intent.FLAG_GRANT_READ_URI_PERMISSION

makingthematrix:
fixes https://wearezeta.atlassian.net/browse/AN-7114

I have found this possible fix for sharing files on new Androids: https://github.com/lubritto/flutter_share/pull/20.
I tweaked the sharing file functionality a bit so that I was able to put the copied code in one place, `AssetIntentsManager`,
and access it from all places where we attempt to send files from the external storage to a conversation.

Please review and test, especially on Android 11 and Android 10 on new Pixel phones. Please also test sharing files from outside the app. It didn't work before in some cases, so if it doesn't work now, it doesn't mean it's a regression. But it shouldn't get worse.
#### APK
[Download build #2865](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2865/artifact/build/artifact/wire-dev-PR3057-2865.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/pages/main/conversation/AssetIntentsManager.java
+++ b/app/src/main/java/com/waz/zclient/pages/main/conversation/AssetIntentsManager.java
@@ -24,6 +24,8 @@ import android.content.ClipData;
 import android.content.Context;
 import android.content.Intent;
 import android.content.pm.PackageManager;
+import android.content.pm.ResolveInfo;
+import android.net.Uri;
 import android.os.Build;
 import android.provider.MediaStore;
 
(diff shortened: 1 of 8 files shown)


makingthematrix:
By the way, I see that if I share file from the outside to a conversation different than the first one on the list, the conversation does not open, but the file is being sent. You can see it if you open the conversation manually. But if the file is sent to only one conversation and it's not the current one, the app should switch to it.

BradleyWire (approved):

mejdoo (approved):

PR #3071, opened 2020-10-27
Fix: Various fixes for bugs regarding the Camera2 API

BradleyWire:
## What's new in this PR?

1: https://wearezeta.atlassian.net/browse/AN-7126
2: https://wearezeta.atlassian.net/browse/AN-7125

## Issues 
1. Camera crashes when you touch to focus the top left corner of the screen (due to the rect values being negative)
2. Camera flash mode shouldn't be enabled for front facing view 
3. Camera flash mode selected does not represent the flash when the picture is taken
4. If you take more than 3 photos in the same session, it'll crash the app 

**For issue 1:** 
When we touch to focus we set a `MeteringRectangle` with a `Rect` param to let the `Camera2` API know where we touch and where we'd like to have focus, the problem is though that that metering rectangle does not accept any negative values. So, it was throwing an exception: 

`E/AndroidRuntime: java.lang.IllegalArgumentException: rect.left must be nonnegative
`

**For issues 2: and 3:** 

The flash mode doesn't correctly reset itself or co-ordinate with the UI correctly 

The preview requests and capture requests rely on having knowledge of the flash intentions for the device. So when we preview we need to determine the flash and set a rolling request, but, when we capture we need to ensure when that is complete we start the rolling preview again (which explains why the flash stays on after the image is captured). 

**For issue 4:** 

When we capture the image we determine this via an image reader, we use the image readers surface to determine what photos we'd like to take from the buffer (i.e. a session capture). The problem we have is to help performance we set a maximum buffer for the images to be stored in cache of the `ImageReader` to 3, once we take more than this if we're calling `imageReader.aquireNextImage()`, it'll crash the app with the following issue: 

```
  throw new IllegalStateException(
                        String.format(
                                "maxImages (%d) has already been acquired, " +
                                "call #close before acquiring more.", mMaxImages));
```

## Testing

**Scenario 1:** 
1. Log in
2. Go to conversation
3. Click take a photo 
4. Open full screen  
5. Switch to front view 
6. You shouldn't see a flash button 

**Scenario 2:** 
1. Log in
2. Go to conversation
3. Click take a photo 
4. Open full screen  
5. Switch to back view 
6. Click the flash mode to change to whatever mode you need
7. When you take a photo it'll do the particular flash you need 

**Scenario 3:** 
1. Log in
2. Go to conversation
3. Click take a photo 
4. Open full screen  
5. Tap to focus in the top left corner 
6. The app shouldn't crash 

**Scenario 4:** 
1. Log in
2. Go to conversation
3. Click take a photo 
4. Open full screen  
5. Take 3 photos (whilst cancelling the final image) 
6. The app shouldn't crash 

#### APK
[Download build #2907](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2907/artifact/build/artifact/wire-dev-PR3071-2907.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/camera/controllers/AndroidCamera2.scala
+++ b/app/src/main/scala/com/waz/zclient/camera/controllers/AndroidCamera2.scala
@@ -125,7 +125,6 @@ class AndroidCamera2(cameraData: CameraData,
     } catch {
       case ex: SecurityException          => error(l"The app has no permission to access the camera", ex)
       case ex: CameraAccessException      => error(l"Camera access error when opening camera: ", ex)
-      case ex: SecurityException          => error(l"The app has no permission to access the camera", ex)
       case ex: IllegalArgumentException   => error(l"Opening the camera failed", ex)
     }
   }
--- a/app/src/main/scala/com/waz/zclient/camera/controllers/AndroidCamera2.scala
+++ b/app/src/main/scala/com/waz/zclient/camera/controllers/AndroidCamera2.scala
@@ -41,10 +41,12 @@ class AndroidCamera2(cameraData: CameraData,
                      texture: SurfaceTexture)
   extends WireCamera with DerivedLogTag {
 
+  import AndroidCamera2._
   import ExifInterface._
   import WireCamera._
   import com.waz.zclient.log.LogUI._
-  import AndroidCamera2._
+
+  private var updatedFlash: FlashMode = flashMode
 
   private lazy val cameraThread = returning(new HandlerThread("CameraThread")) {
     _.start()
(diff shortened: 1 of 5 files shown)


app/src/main/scala/com/waz/zclient/camera/controllers/AndroidCamera2.scala
```diff
@@ -125,9 +123,10 @@ class AndroidCamera2(cameraData: CameraData,
         }
       }, cameraHandler)
     } catch {
-      case ex: CameraAccessException    => error(l"Camera access error when opening camera: ", ex)
-      case ex: SecurityException        => error(l"The app has no permission to access the camera", ex)
-      case ex: IllegalArgumentException => error(l"Opening the camera failed", ex)
+      case ex: SecurityException          => error(l"The app has no permission to access the camera", ex)
+      case ex: CameraAccessException      => error(l"Camera access error when opening camera: ", ex)
+      case ex: SecurityException          => error(l"The app has no permission to access the camera", ex)
```
gizemb, line 128: duplicate w/ line 126

mejdoo, line 128: `case ex: SecurityException          => error(l"The app has no permission to access the camera", ex)` is duplicated. Check line 126

gizemb (changes requested):

mejdoo (approved):

gizemb (approved):

android10 (approved):
:+1:

PR #3081, opened 2020-11-03
fix: Get random profile pictures from Unsplash

makingthematrix:
fixes https://wearezeta.atlassian.net/browse/AN-7047

I'm not exactly sure how long this functionality was broken but it looks like it was for a long time. The previous code
tried to treat an Unsplash URL as if it was a regular asset that was supposed to be downloaded through our backend.
The backend failed because it didn't understand the request.I rewrote `AccountManager.addUnsplashPicture()` and added
methods to `AssetService` and `AssetClient` which download the picture from Unsplash. Then the picture is set as
the profile one and sent to the backend exactly as if it was a regular profile picture.

I also added a button in Dev Settings to test it. From now on it can be used if we want to test any change in code connected to profile pictures.

#### APK
[Download build #2926](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2926/artifact/build/artifact/wire-dev-PR3081-2926.apk)
[Download build #2927](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2927/artifact/build/artifact/wire-dev-PR3081-2927.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/assets/AndroidUriHelper.scala
+++ b/app/src/main/scala/com/waz/zclient/assets/AndroidUriHelper.scala
@@ -39,7 +39,7 @@ class AndroidUriHelper(context: Context) extends UriHelper with DerivedLogTag {
   private def cursor(uri: URI): Managed[Cursor] =
     Managed.create(
       context.getContentResolver.query(androidUri(uri), null, null, null, null)
-    )(_.close())
+    ) { c => Option(c).foreach(_.close()) }
 
   override def openInputStream(uri: URI): Try[InputStream] = Try {
     context.getContentResolver.openInputStream(androidUri(uri))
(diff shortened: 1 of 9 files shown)


app/src/main/scala/com/waz/zclient/assets/AndroidUriHelper.scala
```diff
@@ -39,7 +39,7 @@ class AndroidUriHelper(context: Context) extends UriHelper with DerivedLogTag {
   private def cursor(uri: URI): Managed[Cursor] =
     Managed.create(
       context.getContentResolver.query(androidUri(uri), null, null, null, null)
-    )(_.close())
+    ) { c => Option(c).foreach(_.close()) }
```
makingthematrix, line 42: A fix for a potential bug: the cursor can be `null` and when `close()` is called on it, it can fail with `NullPointerException`.

zmessaging/src/main/scala/com/waz/sync/client/AssetClient.scala
```diff
@@ -142,6 +150,7 @@ object AssetClient {
   implicit val DefaultExpiryTime: Expiration = 1.hour
 
   val AssetsV3Path = "/assets/v3"
+  val UnsplashUrl: URL = new URL("https://source.unsplash.com/800x800/?landscape")
```
mejdoo, line 153: Define then "https://source.unsplash.com/800x800/?landscape" as constant

mejdoo (approved):

android10 (approved):
:+1:

PR #3073, opened 2020-10-27
Build System Improvement - Part 1 : main build.gradle file 

mejdoo:
## What's new in this PR?

This PR comes with a refactoring solution to improve the readability the current build system which is very complicated and hard to understand.

## Main improvements
- Move the gradle tasks and configurations to different scripts
- Manage the versioning from a single place : `Dependencies.kt`

#### APK
[Download build #2912](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2912/artifact/build/artifact/wire-dev-PR3073-2912.apk)
[Download build #2916](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2916/artifact/build/artifact/wire-dev-PR3073-2916.apk)

PR #3083, opened 2020-11-04
Build System Improvement - Part2

mejdoo:
## What's new in this PR?

This is a follow up of #3073 

### Main improvements

- Merge quality related stuff in one gradle file `quality.gradle` and move all the configs below the root `/config` folder
- Improve versioning
- Manage FeatureFlags from Kotlin DSL.


mejdoo:
build again

buildSrc/src/main/kotlin/FeatureFlags.kt
```diff
@@ -0,0 +1,6 @@
+// Kotlin Feature Flags
+object FeatureFlags {
+    const val REGITRATION = false
```
gizemb, line 3: typo

scripts/quality.gradle
```diff
@@ -0,0 +1,177 @@
...
+            xml.enabled = true
+            html.enabled = true
+        }
+    }
+}
+
+//**************** Lint *****************//
+
+subprojects {
+    if (project.plugins.hasPlugin('com.android.application')) {
```
gizemb, line 79: what about library projects? storage, zmessaging, etc?

scripts/quality.gradle
```diff
@@ -0,0 +1,177 @@
...
+        android {
+            lintOptions {
+                // set to true to turn off analysis progress reporting by lint
```
gizemb, line 82: we can get rid of those comments IMO. isn't there a lint task documentation somewhere?

scripts/quality.gradle
```diff
@@ -0,0 +1,177 @@
...
+                xmlReport true
+                // if true, generate an HTML report (with issue explanations, sourcecode, etc)
+                htmlReport true
+                // set to true to have all release builds run lint on issues with severity=fatal
+                // and abort the build (controlled by abortOnError above) if fatal issues are found
+                checkReleaseBuilds false
+
+
+                lintConfig file("${project.rootDir}/config/lint/lint-config.xml")
+
```
gizemb, line 112: a little formatting would be nice

scripts/quality.gradle
```diff
@@ -0,0 +1,177 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+//**************** Checkstyle *****************//
+
```
gizemb, line 20: I think we can break this thing down even more into their own files and add
```
apply from checkstyle
apply from pmd
...
```
here

app/config/lint-config.xml
```diff
@@ -1,232 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
gizemb, line 1: what was the difference between this config and `lint-config-with-unused.xml` exactly?

gizemb (changes requested):

scripts/quality.gradle
mejdoo: I started by doing a single gradle file for each section in #3073 `detekt.gradle` and `jacoco.gradle`. Then, @BradleyWire suggested to do it in a single place `quality.gadle`. It's not a requirement but think it's better to follow the same approach of `wire-android-reloaded`.

scripts/quality.gradle
mejdoo: No idea about the documentation, but I'm fine with deleting those comments.

scripts/quality.gradle
BradleyWire: Yes, we can still do a single quality.gradle file and apply that but the quality.gradle file can have its own plugins

build.gradle
```diff
@@ -37,16 +37,8 @@ allprojects {
...
-    kotlinCore = kotlinSettings || kotlinRegistration
-}
-
 apply from: "./scripts/avs.gradle"
-apply from: "./scripts/detekt.gradle"
-apply from: "./scripts/jacoco.gradle"
+apply from: "./scripts/quality.gradle"
 apply from: "./scripts/licensing.gradle"
-apply from: "./scripts/customization.gradle"
+apply from: "./scripts/whitelabel.gradle"
```
android10, line 43: :+1:

scripts/licensing.gradle
```diff
@@ -1,23 +1,59 @@
...
-//            "**/*build*",
-//            "**/*target*",
-//            "**/*gen*",
-//            "**/*generated*",
-//            ".idea/**"
-//    ])
-//}*/
-//licenseFormat.dependsOn licenseFormatAndroid
\ No newline at end of file
+subprojects {
```
android10, line 32: We have to be careful with this, because we do not want to get ownership over code taken from other projects. For Example there are some files that were taken from my project...I do not mind to be honest, but changing the author could bring even legal problems and as far as I understand, that will apply to all files right?

scripts/quality.gradle
android10: I jump in here, because I was the one putting everything inside the ```quality``` file. 
At the time being, the file is not big and perfectly understandable. Personally I found this the best way of organizing scripts by ```aspects```. If we start creating these tiny scripts, we are loosing consistency, since there are aspects and technologies mixed up. 

We have 2 options, either we put everything in one file as soon as the file does not grow until it turns into a God Script or we change our strategy and organization related to scripts. 
If you feel like the second option, please drop your suggestions.

android10 (approved):
:+1: I'm approving but I would definitely like to start a discussion or have an answer to my comments.

scripts/licensing.gradle
mejdoo: Good point, I totally agree with you opinion regarding copyright. This will apply to all files in the all the modules with the following extensions `.java .scala, gradle,xml.` `.kt` is missing btw. What do you suggest then? Should we ignore some specific files or disable completely the licensing plugin to stay in the safe side?

android10: I would suggest a way to check whether this header exist first. If does not, then add it. 
I do not know the plugin and I do it manually in my projects, actually the IDE does it for me within new files.

In any case, we have to be responsible if we are copying code from other projects. It is a commitment. 
So just make sure we are taking any ownership from code that does not belong to us.

android10: I guess we have to investigate a bit here, Unless someone else has experience on this topic. 
cc @makingthematrix @gizemb @BradleyWire

makingthematrix: As far as I understand, every source code file should start with the header. If the code contains a snipped taken from elsewhere, that code should additionally be preceded by a comment about who is the author and where was it taken from. We should copy-paste whole files from other projects.

scripts/quality.gradle
gizemb: I personally find it big, but I don't mind if it stays as it is. the main reason why I requested changes was the [typo](https://github.com/wireapp/wire-android/pull/3083#discussion_r517330115) and [this](https://github.com/wireapp/wire-android/pull/3083#discussion_r517336875) comment. the typo has been fixed but I still don't have an answer for the other question. As soon as I get an answer to it I can revoke my CR.

scripts/quality.gradle
mejdoo: lint was enabled only for the `app` module, if that makes sense we can enabled as well for the other modules.

gizemb: hmm, ok. I saw that there is another configuration ready for library modules, but didn't know that it was disabled for them.

gizemb (approved):

mejdoo:
I'm closing this deprecated  PR. The job was failing due to a lint config problem. I don't want to invest more time on it. We have other priorities.

PR #2950, opened 2020-07-31
Feat: Back up - Part 5 : Convert Iterator to suspendable one with Either

gizemb:
### Issues

Kotlin's Iterator does not support suspend functions and throws exceptions when something goes wrong.

```
public interface Iterator<out T> {
    public operator fun next(): T     <-- not a `suspend fun`, returns the next item or throws exception
    public operator fun hasNext(): Boolean
}
```
Most `next()` operations such as reading items from the database are implemented as `suspend` functions in our project. Also, we try to avoid throwing exceptions and want to encapsulate them in `Either` structure. 

### Solutions

This PR creates a new Iterator, `BatchReader`, that has a `suspend` function which returns an `Either`. There is also a `forEach` operation provided, which iterates over all items in the iterator as long as there is no `Failure`.

```
interface BatchReader<T> {
    suspend fun readNext(): Either<Failure, T?>
}
```

### Testing

Unit tests are added for BatchReader, and affected ones are updated.

#### Notes
#2951 was merged into this PR. 

#### APK
[Download build #2396](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2396/artifact/build/artifact/wire-dev-PR2950-2396.apk)
[Download build #2406](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2406/artifact/build/artifact/wire-dev-PR2950-2406.apk)
[Download build #2413](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2413/artifact/build/artifact/wire-dev-PR2950-2413.apk)
[Download build #2415](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2415/artifact/build/artifact/wire-dev-PR2950-2415.apk)

Issue #2949, opened 2020-07-30
No timestamps for "You called" or "X Called"


Transisto:
self explanatory.

PR #3092, opened 2020-11-12
feat: Shared tracking id

makingthematrix:
implements https://wearezeta.atlassian.net/browse/SQCORE-121

The tracking ids are already present in the client code, but now every device has its own tracking id and they are not connected. This PR introduces sharing tracking ids among devices belonging to the same user. The general idea is that every time a new device is added to the user, that device generates a new tracking id and sends it to all other devices. This happens also when the device updates for the first time to the version containing this feature. We are aware there is a risk of race condition when two devices send their tracking ids in the same time, but we decided it's small enough and we can ignore it.

The change requires an update in generic-message-proto. The new version is 1.27.1. It contains a new field called DataTransfer and inside it there's a field trackingIdentifier which is the trackingId that should replace the local tracking id. In the future, DataTransfer might contain more fields, but for now it's only this one, so in the Scala code I made a shortcut that let us easier unpack the tracking id from the incoming message.

I also added a new button to the Dev Settings that can be used to generate a new tracking id without the need to remove and re-add a device. This can help in testing.

#### APK
[Download build #2949](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2949/artifact/build/artifact/wire-dev-PR3092-2949.apk)
[Download build #2950](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2950/artifact/build/artifact/wire-dev-PR3092-2950.apk)
[Download build #2952](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2952/artifact/build/artifact/wire-dev-PR3092-2952.apk)

### DIFF
--- a/app/src/main/res/values/strings_no_translate.xml
+++ b/app/src/main/res/values/strings_no_translate.xml
@@ -158,6 +158,8 @@
     <string translatable="false" name="pref_dev_websocket_ping_interval_title">WebSocket Ping Frequency (ms)</string>
     <string translatable="false" name="pref_dev_random_last_notification_id_title">Generate a random new value for LastStableNotification</string>
     <string translatable="false" name="pref_dev_random_last_notification_id_summary">Generate a new, random value for the LastStableNotification preference. Use it to test history loss.</string>
+    <string translatable="false" name="pref_dev_random_tracking_id_title">Generate a random new value for TrackingId</string>
+    <string translatable="false" name="pref_dev_random_tracking_id_summary">Generate a new, random value for the CountlyTrackingId preference. Use it to test history loss.</string>
     <string translatable="false" name="pref_dev_full_conv_key">>PREF_FULL_CONV</string>
     <string translatable="false" name="pref_dev_full_conv_summary">Instead of an empty conversation, create one populated with all contacts in the list (up to 256)</string>
     <string translatable="false" name="pref_dev_full_conv_title">Create full conversation</string>
--- a/app/src/main/res/values/strings_no_translate.xml
+++ b/app/src/main/res/values/strings_no_translate.xml
@@ -159,7 +159,7 @@
     <string translatable="false" name="pref_dev_random_last_notification_id_title">Generate a random new value for LastStableNotification</string>
     <string translatable="false" name="pref_dev_random_last_notification_id_summary">Generate a new, random value for the LastStableNotification preference. Use it to test history loss.</string>
     <string translatable="false" name="pref_dev_random_tracking_id_title">Generate a random new value for TrackingId</string>
-    <string translatable="false" name="pref_dev_random_tracking_id_summary">Generate a new, random value for the CountlyTrackingId preference. Use it to test history loss.</string>
+    <string translatable="false" name="pref_dev_random_tracking_id_summary">Generate a new, random value for the CurrentTrackingId preference. Use it to test tracking.</string>
     <string translatable="false" name="pref_dev_full_conv_key">>PREF_FULL_CONV</string>
     <string translatable="false" name="pref_dev_full_conv_summary">Instead of an empty conversation, create one populated with all contacts in the list (up to 256)</string>
     <string translatable="false" name="pref_dev_full_conv_title">Create full conversation</string>
--- a/app/src/main/scala/com/waz/zclient/MainActivity.scala
+++ b/app/src/main/scala/com/waz/zclient/MainActivity.scala
@@ -240,7 +240,7 @@ class MainActivity extends BaseActivity
   private def initTracking: Future[Unit] =
     for {
       prefs            <- userPreferences.head
-      id               <- prefs.preference(CountlyTrackingId).apply()
+      id               <- prefs.preference(CurrentTrackingId).apply()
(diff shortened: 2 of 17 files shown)


app/src/main/res/values/strings_no_translate.xml
```diff
@@ -158,6 +158,8 @@
     <string translatable="false" name="pref_dev_websocket_ping_interval_title">WebSocket Ping Frequency (ms)</string>
     <string translatable="false" name="pref_dev_random_last_notification_id_title">Generate a random new value for LastStableNotification</string>
     <string translatable="false" name="pref_dev_random_last_notification_id_summary">Generate a new, random value for the LastStableNotification preference. Use it to test history loss.</string>
+    <string translatable="false" name="pref_dev_random_tracking_id_title">Generate a random new value for TrackingId</string>
+    <string translatable="false" name="pref_dev_random_tracking_id_summary">Generate a new, random value for the CountlyTrackingId preference. Use it to test history loss.</string>
```
BradleyWire, line 162: I know this is only a developer's button, but, I don't think this should be so tightly coupled to Countly, atleast if we change the analytics service we use we don't have to worry about changing strings with it

zmessaging/src/main/scala/com/waz/sync/handler/TrackingSyncHandler.scala
```diff
@@ -0,0 +1,22 @@
...
+
+import com.waz.model.GenericContent.DataTransfer
+import com.waz.model.{ConvId, GenericMessage, TrackingId, Uid, UserId}
+import com.waz.sync.SyncResult
+import com.waz.sync.otr.OtrSyncHandler
+import com.waz.sync.otr.OtrSyncHandler.TargetRecipients
+
+import scala.concurrent.Future
+
+class TrackingSyncHandler(selfUserId: UserId, otrSync: OtrSyncHandler) {
```
BradleyWire, line 11: Are we able to test this?

app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
```diff
@@ -53,6 +53,28 @@ class GlobalTrackingController(implicit inj: Injector, cxt: WireContext)
...
 
+  tracking.onTrackingIdChange.foreach(setTrackingId)
+
+  private def setTrackingId(id: TrackingId): Future[Unit] =
+    for {
+      am            <- am.head
+      inited        <- initialized.head
+      curTrackingId <- am.storage.userPrefs(CountlyTrackingId).apply()
+      isNew         =  curTrackingId.contains(id)
+      _             =  if (isNew) am.storage.userPrefs(CountlyTrackingId) := Some(id)
```
BradleyWire, line 64: Again, I don't think we should be sharing implementation details here around `CountlyTrackingId` incase we change the service in the future. This can be something like `CurrentTrackingId`. 

Obviously, if the value of `CountlyTrackingId` has `Countly` in the key name, we can't avoid that change without migrating data but the variable name is still fine to change.

app/src/main/scala/com/waz/zclient/MainActivity.scala
```diff
@@ -241,19 +241,18 @@ class MainActivity extends BaseActivity
...
-            .flatMap(_ => prefs(TrackingEnabledOneTimeCheckPerformed) := true)
-        } else Future.successful(())
-      _                <-
-        if(analyticsEnabled && isProUser) {
-          inject[GlobalTrackingController].init()
-        } else Future.successful(())
+      _                <- if(!check)
+                            (prefs(TrackingEnabled) := isProUser).flatMap(_ => prefs(TrackingEnabledOneTimeCheckPerformed) := true)
+                          else Future.successful(())
+      _                <- if(analyticsEnabled && isProUser) trackingCtrl.init() else Future.successful(())
```
BradleyWire, line 255: Formatting is off here with `if(` rather than `if ( ` (and above)

BradleyWire (changes requested):

zmessaging/src/main/scala/com/waz/sync/handler/TrackingSyncHandler.scala
makingthematrix: @BradleyWire : Rught now, this handler has only one trivial method that only calls `OtrSyncHandler.postOtrMessage`. We have unit tests for `OtrSyncHandler`.

app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
makingthematrix: ok

app/src/main/scala/com/waz/zclient/MainActivity.scala
makingthematrix: ok

app/src/main/res/values/strings_no_translate.xml
makingthematrix: ok

makingthematrix:
build again

BradleyWire (approved):

mejdoo (approved):

PR #2982, opened 2020-08-14
New backups (15): Put together backup creation in the use case

makingthematrix:
#### APK
[Download build #2548](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2548/artifact/build/artifact/wire-dev-PR2982-2548.apk)
[Download build #2550](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2550/artifact/build/artifact/wire-dev-PR2982-2550.apk)
[Download build #2554](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2554/artifact/build/artifact/wire-dev-PR2982-2554.apk)

PR #2970, opened 2020-08-11
New backups: Encryption handler

makingthematrix:
Work in progress.

The handler is for now written in one file. First I want to make it working, test it, and then discussr how to split it into smaller chunks. 

It's not possible to write unit tests for it, because the handler uses Sodium, a dynamically loaded library which needs a device or an emulator to work. I put the test in the `androidTest` directory, but currently when I run it I get an error:
```
> Task :app:transformClassesWithDexBuilderForAvsDebugAndroidTest FAILED
java.lang.UnsupportedOperationException: java.lang.UnsupportedOperationException: This feature requires ASM6
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
...
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:157)

Caused by: java.lang.UnsupportedOperationException: This feature requires ASM6
```
#### APK
[Download build #2545](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2545/artifact/build/artifact/wire-dev-PR2970-2545.apk)

PR #2968, opened 2020-08-11
Bump AVS version to 6.2.21

mejdoo:
## What's new in this PR?

Bump AVS to the latest version 6.2.21

#### APK
[Download build #2500](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2500/artifact/build/artifact/wire-dev-PR2968-2500.apk)

PR #3097, opened 2020-11-19
[Calling]: Feature : maximize/minimize video/screen share on double click

mejdoo:
## What's new in this PR?

Once this PR is merged, it will be possible during a call to maximise a video or screen share of any user on  double click.

**No need anymore to turn off all the videos if someone is sharing the screen.**

## Note:

The feature flag `callingVvmMaximizeMinimizeVideo`  is now set to **true** for testing purposes, till we decide either this feature will go with next release train or not. 

https://wearezeta.atlassian.net/browse/SQCALL-9
#### APK
[Download build #2964](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2964/artifact/build/artifact/wire-dev-PR3097-2964.apk)
[Download build #2965](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2965/artifact/build/artifact/wire-dev-PR3097-2965.apk)
[Download build #2968](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2968/artifact/build/artifact/wire-dev-PR3097-2968.apk)
[Download build #2970](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2970/artifact/build/artifact/wire-dev-PR3097-2970.apk)
[Download build #2978](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2978/artifact/build/artifact/wire-dev-PR3097-2978.apk)
[Download build #2979](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2979/artifact/build/artifact/wire-dev-PR3097-2979.apk)
[Download build #2981](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2981/artifact/build/artifact/wire-dev-PR3097-2981.apk)
[Download build #2982](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2982/artifact/build/artifact/wire-dev-PR3097-2982.apk)
[Download build #2984](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2984/artifact/build/artifact/wire-dev-PR3097-2984.apk)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -126,7 +126,7 @@ android {
         //Feature Flags
         buildConfigField 'boolean', 'KOTLIN_SETTINGS', "$rootProject.ext.kotlinSettings"
         buildConfigField 'boolean', 'KOTLIN_CORE', "$rootProject.ext.kotlinCore"
-        buildConfigField 'boolean', 'CALLING_VVM_MAXIMIZE_MINIMIZE_VIDEO', "$rootProject.ext.callingVvmMaximizeMinimizeVideo"
+        buildConfigField 'boolean', 'MAXIMIZE_MINIMIZE_VIDEO', "$rootProject.ext.maximizeMinimizeVideo"
     }
 
     packagingOptions {
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -123,9 +123,10 @@ android {
         buildConfigField 'String',  'COUNTLY_SERVER_URL',            "\"$buildtimeConfiguration.configuration.countly_server_url\""
         buildConfigField 'String',  'COUNTLY_APP_KEY',               "\"$buildtimeConfiguration.configuration.countly_app_key\""
 
-        //Kotlin Feature Flags
+        //Feature Flags
         buildConfigField 'boolean', 'KOTLIN_SETTINGS', "$rootProject.ext.kotlinSettings"
         buildConfigField 'boolean', 'KOTLIN_CORE', "$rootProject.ext.kotlinCore"
+        buildConfigField 'boolean', 'CALLING_VVM_MAXIMIZE_MINIMIZE_VIDEO', "$rootProject.ext.callingVvmMaximizeMinimizeVideo"
     }
 
     packagingOptions {
--- a/app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
@@ -34,7 +34,7 @@ import com.waz.zclient.log.LogUI._
 import com.waz.zclient.security.SecurityPolicyChecker
 import com.waz.zclient.utils.ContextUtils._
 import com.waz.zclient.{BuildConfig, FragmentHelper, R}
-import com.wire.signals.{Signal, Subscription}
+import com.wire.signals.Signal
 import com.waz.zclient.calling.CallingFragment.MaxVideoPreviews
 import com.waz.zclient.calling.controllers.CallController.CallParticipantInfo
 
@@ -45,10 +45,25 @@ class CallingFragment extends FragmentHelper {
   private lazy val themeController    = inject[ThemeController]
   private lazy val controlsFragment   = ControlsFragment.newInstance
   private lazy val previewCardView    = view[CardView](R.id.preview_card_view)
-  private var  videoGridSubscription  = Option.empty[Subscription]
   private lazy val videoGrid = returning(view[GridLayout](R.id.video_grid)) { vh =>
     controller.theme.map(themeController.getTheme).foreach { theme =>
-      vh.foreach { _.setBackgroundColor(getStyledColor(R.attr.wireBackgroundColor, theme)) }
+      vh.foreach {
+        _.setBackgroundColor(getStyledColor(R.attr.wireBackgroundColor, theme))
+      }
+
+      initGridLayout()
+
+      controller.initVideo.foreach { _ =>
+        initGridLayout()
+      }
+    }
+
+    def initGridLayout() =  videoGridInfo.foreach {
+      case (selfParticipant, videoUsers, infos, participants, isVideoBeingSent) =>
+        vh.foreach { grid =>
+          val isFullScreenEnabled = controller.isFullScreenEnabled.currentValue.getOrElse(false)
+          if (!isFullScreenEnabled) refreshVideoGrid(grid, selfParticipant, videoUsers, infos, participants, isVideoBeingSent)
+        }
     }
   }
 
@@ -69,11 +84,7 @@ class CallingFragment extends FragmentHelper {
       if (_) inject[SecurityPolicyChecker].run(getActivity)
     }
 
-    initVideoGrid()
-
-    controller.initVideo.foreach { _ =>
-      initVideoGrid()
-    }
+    videoGrid
   }
 
   override def onBackPressed(): Boolean =
@@ -133,14 +144,6 @@ class CallingFragment extends FragmentHelper {
 
     videoUsers.map { participant => viewMap.getOrElse(participant, createView(participant)) }
   }
-
-  private def initVideoGrid(): Unit =
-    videoGridSubscription = Option(videoGridInfo.foreach {
-      case (selfParticipant, videoUsers, infos, participants, isVideoBeingSent) =>
-        videoGrid.foreach { grid =>
-          refreshVideoGrid(grid, selfParticipant, videoUsers, infos, participants, isVideoBeingSent)
-        }
-    })
 
   private def refreshVideoGrid(grid            : GridLayout,
                                selfParticipant : Participant,
@@ -227,7 +230,6 @@ class CallingFragment extends FragmentHelper {
   def clearVideoGrid(): Unit = {
     videoGrid.foreach(_.removeAllViews())
     viewMap = Map.empty
-    videoGridSubscription.foreach(_.unsubscribe())
   }
 
   def showFullScreenVideo(participant: Participant): Unit = getChildFragmentManager
--- a/app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
@@ -71,6 +71,10 @@ class CallingFragment extends FragmentHelper {
     controller.isCallIncoming.head.foreach {
       if (_) inject[SecurityPolicyChecker].run(getActivity)
     }(Threading.Ui)
+
+    controller.initVideo.foreach { _ =>
+      initVideoGrid()
+    }
   }
 
   override def onBackPressed(): Boolean =
--- a/app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
@@ -35,7 +35,7 @@ import com.waz.zclient.log.LogUI._
 import com.waz.zclient.security.SecurityPolicyChecker
 import com.waz.zclient.utils.ContextUtils._
 import com.waz.zclient.{BuildConfig, FragmentHelper, R}
-import com.wire.signals.Signal
+import com.wire.signals.{Signal, Subscription}
 
 class CallingFragment extends FragmentHelper {
 
@@ -44,11 +44,49 @@ class CallingFragment extends FragmentHelper {
   private lazy val themeController = inject[ThemeController]
   private lazy val controlsFragment = ControlsFragment.newInstance
   private lazy val previewCardView = view[CardView](R.id.preview_card_view)
-  private lazy val fullScreenVideoContainer = view[FrameLayout](R.id.full_screen_video_layout)
   private var viewMap = Map[Participant, UserVideoView]()
+  private lazy val videoGrid = view[GridLayout](R.id.video_grid)
+  private var videoGridSubscription = Option.empty[Subscription]
 
-  private lazy val videoGrid = returning(view[GridLayout](R.id.video_grid)) { vh =>
-    Signal.zip(
+  override def onCreate(savedInstanceState: Bundle): Unit = {
+    super.onCreate(savedInstanceState)
+    controller.theme.map(themeController.getTheme).onUi(theme => videoGrid.foreach(_.setBackgroundColor(getStyledColor(R.attr.wireBackgroundColor, theme))))
+  }
+
+  override def onCreateView(inflater: LayoutInflater, container: ViewGroup, savedInstanceState: Bundle): View =
+    returning(inflater.inflate(R.layout.fragment_calling, container, false)) { v =>
+      controller.theme(t => v.asInstanceOf[ThemeControllingFrameLayout].theme ! Some(t))
+    }
+
+  override def onViewCreated(view: View, savedInstanceState: Bundle): Unit = {
+    super.onViewCreated(view, savedInstanceState)
+
+    initVideoGrid()
+
+    getChildFragmentManager
+      .beginTransaction
+      .replace(R.id.controls_layout, controlsFragment, ControlsFragment.Tag)
+      .commit
+
+    controller.isCallIncoming.head.foreach {
+      if (_) inject[SecurityPolicyChecker].run(getActivity)
+    }(Threading.Ui)
+  }
+
+  override def onBackPressed(): Boolean =
+    withChildFragmentOpt(R.id.controls_layout) {
+      case Some(f: FragmentHelper) if f.onBackPressed()               => true
+      case Some(_) if getChildFragmentManager.popBackStackImmediate() => true
+      case _ => super.onBackPressed()
+    }
+
+  override def onDestroyView(): Unit = {
+    super.onDestroyView()
+    clearVideoGrid()
+  }
+
+  def initVideoGrid(): Unit = {
+    videoGridSubscription = Option(Signal.zip(
       controller.allVideoReceiveStates,
       controller.callingZms.map(zms => Participant(zms.selfUserId, zms.clientId)),
       controller.isVideoCall,
@@ -64,7 +102,10 @@ class CallingFragment extends FragmentHelper {
         viewMap = viewMap.updated(participant, v)
         if (BuildConfig.CALLING_VVM_MAXIMIZE_MINIMIZE_VIDEO) {
           v.onDoubleClick.onUi { _ =>
-            showFullScreenVideo(v, selfParticipant, participant)
+            if (participants.size > 2) {
+              showFullScreenVideo(participant)
+              clearVideoGrid()
+            }
           }
         }
       }
@@ -73,7 +114,7 @@ class CallingFragment extends FragmentHelper {
 
       val isVideoBeingSent = !vrs.get(selfParticipant).contains(VideoState.Stopped)
 
-      vh.foreach { v =>
+      videoGrid.foreach { v =>
         val videoUsers = vrs.toSeq.collect {
           case (participant, _) if participant == selfParticipant && videoCall && incoming => participant
           case (participant, VideoState.Started | VideoState.Paused | VideoState.BadConnection | VideoState.NoCameraPermission | VideoState.ScreenShare) => participant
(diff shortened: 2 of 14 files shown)


app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
```diff
@@ -250,62 +174,36 @@ class CallingFragment extends FragmentHelper {
...
-  override def onBackPressed(): Boolean =
-    withChildFragmentOpt(R.id.controls_layout) {
-      case Some(f: FragmentHelper) if f.onBackPressed()               => true
-      case Some(_) if getChildFragmentManager.popBackStackImmediate() => true
-      case _ => super.onBackPressed()
-    }
-
-  override def onDestroyView(): Unit = {
-    super.onDestroyView()
+  def clearVideoGrid(): Unit = {
```
mejdoo, line 197: The VideoGrid should be cleared and reloaded otherwise AVS native lib will be crashing. There is a limitation right now, we can't use `com.waz.avs.VideoRenderer` with same parameters in different views, at the same time.

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
```diff
@@ -0,0 +1,94 @@
...
+    getParentFragment.asInstanceOf[CallingFragment].initVideoGrid()
+  }
+}
+
+object FullScreenVideoFragment {
+  val Tag = classOf[FullScreenVideoFragment].getName
+  val PARTICIPANT_BUNDLE_KEY = "participant"
+
+  def newInstance(participant: Participant): Fragment = {
+    val fragment = new FullScreenVideoFragment
```
BradleyWire, line 88: You can use the `returning(new FullScreenVideoFragment) { fragment =>  }` here to clean it up a little

app/src/main/scala/com/waz/zclient/calling/views/UserVideoView.scala
```diff
@@ -0,0 +1,125 @@
...
+      case Some(p) if p.isSelf => view.setText(getString(R.string.calling_self, p.displayName))
+      case Some(p)             => view.setText(p.displayName)
+      case _                   =>
+    }
+  }
+
+  Signal.zip(controller.isGroupCall, controller.controlsVisible,
+    controller.otherParticipants.map(_.size)
+  ).map {
+    case (true, false, 0 | 1 | 2) => View.GONE
```
BradleyWire, line 94: formatting here the => need to line up

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
```diff
@@ -0,0 +1,94 @@
...
+
+  def minimizeVideo(container: FrameLayout, userVideoView: UserVideoView): Unit = {
+    container.removeView(userVideoView)
+    getFragmentManager.popBackStack()
+    getParentFragment.asInstanceOf[CallingFragment].initVideoGrid()
+  }
+}
+
+object FullScreenVideoFragment {
+  val Tag = classOf[FullScreenVideoFragment].getName
```
BradleyWire, line 84: Do we want this like this? In case of obfuscation this may not display the correct tag

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
```diff
@@ -0,0 +1,94 @@
...
+      else userVideoView = new OtherVideoView(getContext, participant)
+
+      userVideoView.onDoubleClick.onUi { _ =>
+        minimizeVideo(container, userVideoView)
+      }
+
+      container.addView(userVideoView)
+
+      controller.allVideoReceiveStates
+        .map(_.getOrElse(participant, VideoState.Unknown)).onUi {
```
BradleyWire, line 70: Is this formatted properly? the indentation looks off for the `case` values

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
```diff
@@ -17,160 +17,76 @@
...
-  private val participantInfoCardView = findById[CardView](R.id.participant_info_card_view)
-
-  protected val stateMessageText = controller.stateMessageText(participant)
-  stateMessageText.onUi(msg => pausedText.setText(msg.getOrElse("")))
+class CallingFragment extends FragmentHelper {
 
-  protected val pausedTextVisible = stateMessageText.map(_.exists(_.nonEmpty))
-  pausedTextVisible.onUi(pausedText.setVisible)
+  private val maxVideoPreviews = 12
+  private lazy val controller = inject[CallController]
```
BradleyWire, line 43: If we're grouping these values together, we normally align the `=`

app/src/main/scala/com/waz/zclient/calling/views/UserVideoView.scala
```diff
@@ -0,0 +1,125 @@
...
+    case (true, false, _)     => View.VISIBLE
+    case _                    => View.GONE
+  }.onUi(participantInfoCardView.setVisibility)
+
+  protected def registerHandler(view: View) = {
+    controller.allVideoReceiveStates.map(_.getOrElse(participant, VideoState.Unknown)).onUi {
+      case VideoState.Paused | VideoState.Stopped => view.fadeOut()
+      case _                 => view.fadeIn()
```
BradleyWire, line 102: formatting here too

BradleyWire (changes requested):

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  def showMaximizedVideo(participant: Participant): Unit =
+    fullScreenVideoContainer.foreach { container =>
+
+      val selfParticipant = controller.callingZms.map(zms => Participant(zms.selfUserId, zms.clientId)).currentValue.get
+
+      var userVideoView: UserVideoView = null
+      if (participant == selfParticipant) userVideoView = new SelfVideoView(getContext, participant)
+      else userVideoView = new OtherVideoView(getContext, participant)
```
gizemb, line 61: this doesn't need to be a `var` I guess?
`val userVideoView = if (...) SelfVideoView() else OtherVideoView()`

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
```diff
@@ -0,0 +1,94 @@
...
+        case VideoState.Started | VideoState.ScreenShare =>
+        case _ => minimizeVideo(container, userVideoView)
+      }
+    }
+
+  def minimizeVideo(container: FrameLayout, userVideoView: UserVideoView): Unit = {
+    container.removeView(userVideoView)
+    getFragmentManager.popBackStack()
+    getParentFragment.asInstanceOf[CallingFragment].initVideoGrid()
+  }
```
gizemb, line 80: there's a really tight coupling between this fragment and CallingFragment. don't we have any other way to communicate?

gizemb (approved):

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
```diff
@@ -17,160 +17,76 @@
...
-
-  private val participantInfoCardView = findById[CardView](R.id.participant_info_card_view)
-
-  protected val stateMessageText = controller.stateMessageText(participant)
-  stateMessageText.onUi(msg => pausedText.setText(msg.getOrElse("")))
+class CallingFragment extends FragmentHelper {
 
-  protected val pausedTextVisible = stateMessageText.map(_.exists(_.nonEmpty))
-  pausedTextVisible.onUi(pausedText.setVisible)
+  private val maxVideoPreviews = 12
```
makingthematrix, line 42: Constants should be defined in the companion object, and the name should start with the capital letter.

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
```diff
@@ -17,160 +17,76 @@
...
+  private lazy val previewCardView = view[CardView](R.id.preview_card_view)
+  private var viewMap = Map[Participant, UserVideoView]()
+  private lazy val videoGrid = view[GridLayout](R.id.video_grid)
+  private var videoGridSubscription = Option.empty[Subscription]
 
-  protected val videoCallInfo = returning(findById[View](R.id.video_call_info)) {
-    _.setBackgroundColor(getColor(R.color.black_16))
+  override def onCreate(savedInstanceState: Bundle): Unit = {
+    super.onCreate(savedInstanceState)
+    controller.theme.map(themeController.getTheme).onUi(theme => videoGrid.foreach(_.setBackgroundColor(getStyledColor(R.attr.wireBackgroundColor, theme))))
```
makingthematrix, line 53: Instead of subscribing to the signal here, put it in the `videoGrid` definition:
```
private lazy val videoGrid = returning(view[GridLayout](R.id.video_grid)) { vh =>
  controller.theme.map(themeController.getTheme).onUi { theme =>
    vh.foreach { ... }
  }
}
```

Tak a look at other places this pattern is used.

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
```diff
@@ -179,18 +95,26 @@ class CallingFragment extends FragmentHelper {
...
-          else new OtherVideoView(getContext, participant)
-        } { v =>
-          viewMap = viewMap.updated(participant, v)
+      def createView(participant: Participant): UserVideoView = returning {
+        if (participant == selfParticipant) new SelfVideoView(getContext, participant)
+        else new OtherVideoView(getContext, participant)
+      } { v =>
+        viewMap = viewMap.updated(participant, v)
+        if (BuildConfig.CALLING_VVM_MAXIMIZE_MINIMIZE_VIDEO) {
+          v.onDoubleClick.onUi { _ =>
```
makingthematrix, line 104: Subscribing to a signal or an event stream inside a method may sometimes lead to unclear behaviour. It's more readable if `v.onDoubleClick.onUi { _ => ... }` is defined on the main level and then, inside the `onUi{ _ => ... }` the app checks what should happen because of the click. But I'm not sure yet how to do it in this particular case. I can look more into it on Monday.

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
```diff
@@ -179,18 +95,26 @@ class CallingFragment extends FragmentHelper {
...
+      }
 
-        def findParticipantNameById(userId: UserId): String = infos.find(_.userId == userId).get.displayName
+      def findParticipantNameById(userId: UserId): String = infos.find(_.userId == userId).get.displayName
 
-        val isVideoBeingSent = !vrs.get(selfParticipant).contains(VideoState.Stopped)
+      val isVideoBeingSent = !vrs.get(selfParticipant).contains(VideoState.Stopped)
 
-      vh.foreach { v =>
+      videoGrid.foreach { v =>
```
makingthematrix, line 117: This can also be moved to the definition of `videoGrid`.

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
makingthematrix: I think in general `userVideoView` should be a lazy val defined on the class level and in its initialization block we can check the bundle for info what should it be and should it subscribe to any signal.

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
```diff
@@ -0,0 +1,94 @@
...
+      if (participant == selfParticipant) userVideoView = new SelfVideoView(getContext, participant)
+      else userVideoView = new OtherVideoView(getContext, participant)
+
+      userVideoView.onDoubleClick.onUi { _ =>
+        minimizeVideo(container, userVideoView)
+      }
+
+      container.addView(userVideoView)
+
+      controller.allVideoReceiveStates
```
makingthematrix, line 69: This subscription to the signal should also be on the class level. Generally, if for some reason it's not to difficult, the logic should be inverted: First subscribe, then inside the subscription decide what to do. Or, even better, you can use `.filter`, `.map`, `.flatMap`, and `.collect` to build the logic into the signal or event stream.

app/src/main/scala/com/waz/zclient/calling/views/UserVideoView.scala
makingthematrix: And also either please put all three original signals in one line, or each in a separate line.

app/src/main/scala/com/waz/zclient/calling/views/UserVideoView.scala
```diff
@@ -0,0 +1,125 @@
...
+    }
+    view match {
+      case vr: VideoRenderer =>
+        controller.allVideoReceiveStates.map(_.getOrElse(participant, VideoState.Unknown)).onUi {
```
makingthematrix, line 106: Same issue as before - it would be better to have a subscription to a signal on the class level, and logic what to do when the signal changes implmeneted inside the subscription, or through map, filter, etc.

zmessaging/src/main/scala/com/waz/service/call/CallInfo.scala
```diff
@@ -143,7 +143,7 @@ case class CallInfo(convId:             ConvId,
 
 object CallInfo {
 
-  case class Participant(userId: UserId, clientId: ClientId, muted: Boolean = false)
+  case class Participant(userId: UserId, clientId: ClientId, muted: Boolean = false) extends Serializable
```
makingthematrix, line 146: Do you really need to extend Serializable? (btw, I'm not sure, but a case class should do it already, I think...)

makingthematrix (changes requested):

app/build.gradle
```diff
@@ -123,9 +123,10 @@ android {
         buildConfigField 'String',  'COUNTLY_SERVER_URL',            "\"$buildtimeConfiguration.configuration.countly_server_url\""
         buildConfigField 'String',  'COUNTLY_APP_KEY',               "\"$buildtimeConfiguration.configuration.countly_app_key\""
 
-        //Kotlin Feature Flags
+        //Feature Flags
         buildConfigField 'boolean', 'KOTLIN_SETTINGS', "$rootProject.ext.kotlinSettings"
         buildConfigField 'boolean', 'KOTLIN_CORE', "$rootProject.ext.kotlinCore"
+        buildConfigField 'boolean', 'CALLING_VVM_MAXIMIZE_MINIMIZE_VIDEO', "$rootProject.ext.callingVvmMaximizeMinimizeVideo"
```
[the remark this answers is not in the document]
mejdoo: I wanted to specify the squad and the parent feature name. Anyway, `MAXIMIZE_MINIMIZE_VIDEO` would be enough.

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
mejdoo: ok done

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
mejdoo: I just followed what it was done in other fragments.

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
mejdoo: Is there anyway to define auto formatting rule for Scala in Android studio? Those classed are mostly auto formatted... @makingthematrix any idea?

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
mejdoo: so not all the name should be capitalised like Java or Kotlin style?

makingthematrix: @mejdoo : You can see `CONSTANT_LIKE_THIS` too, but at least for me this style means this is aconstant that should be resolved at compile time, and `ConstantLikeThis` is just a magic number, a default string, etc. I agree it's not consistent.

mejdoo: ok anyway, I changed this to `MaxVideoPreviews` in the companion object.

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
makingthematrix: An event stream defined in a controller connecting the two fragments could help:
```
val initVideo = EventStream[Unit]() // in CallingController, for example
initVideo.foreach(callingFragment.initVideoGrid)
...
callingController.initVideo ! (()) // here in FullScreenVideoFragment
```

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
makingthematrix: There was never a problem with this, to be honest. I'm fine with this.

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
makingthematrix: This alignment is not automatic. It's just for readability and it's up to the person writing the code. If I see that something can be easier to write with some small changes in alignment then I do it, but it's not a requirement.

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
gizemb: we don't obfuscate the code

BradleyWire (approved):

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
mejdoo: @makingthematrix  I didn't get how you wanna refactor this. 

I think @gizemb proposal looks fine.

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
mejdoo: ok done!

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
mejdoo: I moved this block outside of the videoGrid lazy loading declaration because I need to clear and initialize the videoGrid multiple times when I maximize/minimize any idea. Otherwise, AVS lib will crash.

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
mejdoo: Not sure how to do it in this case, as I need to call `minimizeVideo(container, userVideoView)` with the parameter  `userVideoView` created inside `showMaximizedVideo(..)`.

app/src/main/scala/com/waz/zclient/calling/views/UserVideoView.scala
mejdoo: This is an old implementation, I just moved the classes to separate files. Could you provide a code snippet for the refactoring.  In this case,`registerHandler()` is called from the classes `SelfVideoView` and `OtherVideoView` which extends from `UserVideoView`.

zmessaging/src/main/scala/com/waz/service/call/CallInfo.scala
mejdoo: right! not needed :)

makingthematrix:
I pushed a commit with changes. My main issue was with this very big and complicated method, `initVideoGrid` which created a signal and subscribed to it - that could potentially lead to memory leaks. I tried to split the signal into a few smaller ones and connect them only later, and also move the subscription to the videoGrid `ViewHolder`. This way, the subscription can be managed automatically and for sure there will be only one at the time.

I still don't like that we subscribe to double clicks inside the method, but I think there's no risk in that.

mejdoo:
> I pushed a commit with changes. My main issue was with this very big and complicated method, `initVideoGrid` which created a signal and subscribed to it - that could potentially lead to memory leaks. I tried to split the signal into a few smaller ones and connect them only later, and also move the subscription to the videoGrid `ViewHolder`. This way, the subscription can be managed automatically and for sure there will be only one at the time.
> 
> I still don't like that we subscribe to double clicks inside the method, but now I think about it, there shouldn't be any risk in that...
> 
> I also changed a few things in `CallController`. Many signals there are initialized at once even if they are never used because there's no video call. I added `lazy` to their declarations.

Ok thank you. Let me test again the feature before merging...

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
```diff
@@ -0,0 +1,93 @@
...
+    val bundle = this.getArguments
+    if (bundle != null) {
+      val participant = bundle.getSerializable(PARTICIPANT_BUNDLE_KEY).asInstanceOf[Participant]
+      showMaximizedVideo(participant)
+      controller.isFullScreenEnabled ! true
+    }
+  }
+
+  def showMaximizedVideo(participant: Participant): Unit =
+    fullScreenVideoContainer.foreach { container =>
```
makingthematrix, line 57: Similar to the issue in `CallingFragment`:
If you make subscriptions like this inside methods, you risk memory leaks and unpredicted behaviour. Every call to the method will create a new subscription and they all will be called next time the signal refreshes, so the same logic will be executed many times when it should be executed only once.

Instead, make subscriptions to signals on the class level (or, in this case, inside `returning(view[FrameLayout](R.id.full_screen_video_container)) { ... }`), and let the method call update some data which in turn will make the signal refresh. 

For example, as I understand, we can have only one participant with maximized video. You can then create a source signal:
```
val maximizedParticipant = Signal[Option[Participant]]()
```
And `selfParticipant` is already a signal in the controller: `controller.selfParticipant`.
So, inside that `returning` clause you can zip together all three:
```
Signal.zip(
  maximizedParticipant,
  controller.selfParticipant,
  controller.allVideoReceiveStates
).foreach { ... }
```
And there you can choose to minimize and maximize views.

app/src/main/scala/com/waz/zclient/calling/FullScreenVideoFragment.scala
```diff
@@ -0,0 +1,93 @@
...
+
+      val selfParticipant = controller.callingZms.map(zms => Participant(zms.selfUserId, zms.clientId)).currentValue.get
+
+      val userVideoView = if (participant == selfParticipant) new SelfVideoView(getContext, participant)
+      else new OtherVideoView(getContext, participant)
```
makingthematrix, line 62: I'd suggest you create a var `Map[Participant, UserVideoView]` and create new views only if they're not already there.

Also, I'm not sure about comparisons of the whole `Participant` case class. Aren't we only interested in user ids of participants?

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
```diff
@@ -17,298 +17,229 @@
...
-    } yield infos.find(_.userId == participant.userId)
+      initGridLayout()
 
-  protected val nameTextView = returning(findById[TextView](R.id.name_text_view)) { view =>
-    participantInfo.onUi {
-      case Some(p) if p.isSelf => view.setText(getString(R.string.calling_self, p.displayName))
-      case Some(p)             => view.setText(p.displayName)
-      case _                   =>
+      controller.initVideo.foreach { _ =>
+        initGridLayout()
```
makingthematrix, line 57: `initGridLayout()` creates a new subscription every time it's called. I don't think this is what you want here. Instead, only one subscription to the signal should be made, in line 54, and here you only refresh the grid once. That means, you take `userGridInfo.head` and call `refreshGridLayout` with its data. 
So, something like this:
```
videoGridInfo.head.foreach { case (selfParticipant, videoUsers, infos, participants, isVideoBeingSent) =>
  controller.isFullScreenEnabled.head.collect { false =>
    vh.foreach { grid =>
      refreshVideoGrid(grid, selfParticipant, videoUsers, infos, participants, isVideoBeingSent)
     }
  }
}
```
It looks similar to `initGridLayout`, but the calls to `.head` make all the difference.

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
```diff
@@ -17,298 +17,229 @@
...
-          case VideoState.ScreenShare =>
-            vr.setShouldFill(false)
-            vr.setFillRatio(1.5f)
-          case _ =>
-            vr.setShouldFill(true)
-            vr.setFillRatio(1.0f)
+    def initGridLayout() =  videoGridInfo.foreach {
+      case (selfParticipant, videoUsers, infos, participants, isVideoBeingSent) =>
+        vh.foreach { grid =>
+          val isFullScreenEnabled = controller.isFullScreenEnabled.currentValue.getOrElse(false)
```
makingthematrix, line 64: Don't access the value of the signal like this. Here you can zip those two signals instead: 
```
Signal.zip(videoGrodInfo, controller.isFullScreenEnabled).foreach {
  case ((sp, vu, info, ps, is), false) => ...
  case _ => 
}
```

makingthematrix (changes requested):

makingthematrix (approved):

PR #3111, opened 2020-12-07
fix: Video recording on Android 11 doesn't work

makingthematrix:
fixes https://wearezeta.atlassian.net/browse/SQCORE-219

Starting Android 11, Android does not allow a third-party app to use `intent.resolveActivity` to easily learn
about other installed apps. One way to solve it is to simply stop using `intent.resolveActivity` and just attempt
to start a new activity and deal with the `ActivityNotFoundException` if it so happens that the app is unable
to find an app for the given intent. In this case, we want to open a video recording app and we can safely
assume that each device we support has it. In a few other places in the coded I decided to use the new code
to guard us from the `ActivityNotFoundException` as well.
#### APK
[Download build #3009](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3009/artifact/build/artifact/wire-dev-PR3111-3009.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/pages/main/conversation/AssetIntentsManager.java
+++ b/app/src/main/java/com/waz/zclient/pages/main/conversation/AssetIntentsManager.java
@@ -85,12 +85,10 @@ public class AssetIntentsManager {
                 intent.setType(INTENT_ALL_TYPES);
                 intent.putExtra(Intent.EXTRA_MIME_TYPES, mimeTypes.toArray());
             }
-            if (!context.getPackageManager().queryIntentActivities(intent, PackageManager.MATCH_ALL).isEmpty()) {
-                callback.openIntent(intent, tpe);
-                return;
-            }
-            Logger.info(TAG, "Did not resolve testing gallery for intent:" + intent.toString());
+
+            callback.openIntent(intent, tpe);
         }
+
         final Intent intent = new Intent(openDocumentAction()).addCategory(Intent.CATEGORY_OPENABLE);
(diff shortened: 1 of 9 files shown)


gizemb (commented):
seems like the suggested way to handle this is to add `<queries/>` tag to `AndroidManifest`. any reason why we didn't go that way?

https://developer.android.com/about/versions/11/privacy/package-visibility

makingthematrix:
> seems like the suggested way to handle this is to add `<queries/>` tag to `AndroidManifest`. any reason why we didn't go that way?
> 
> https://developer.android.com/about/versions/11/privacy/package-visibility

I followed the advise from this article: https://cketti.de/2020/09/03/avoid-intent-resolveactivity/
In short, since we talk here about something so common as video recording, we can safely assume it will work and handling the exception is a rare corner case. It's more straightforward than asking every time if video recording is possible.

mejdoo:
> seems like the suggested way to handle this is to add `<queries/>` tag to `AndroidManifest`. any reason why we didn't go that way?
> 
> https://developer.android.com/about/versions/11/privacy/package-visibility

I think it's not possible to use `<queries>`without upgrading to the latest available Android Gradle plugin. Check the last paragraph in this official blog post.

https://medium.com/androiddevelopers/package-visibility-in-android-11-cc857f221cd9

mejdoo (approved):
It's an ok workaround if we assume that we are not able to do it using `<queries/>` due to Scala Android plugin limitations.

gizemb (approved):
I approve considering the limitation @mejdoo mentioned.

PR #3115, opened 2020-12-09
[Calling] : fix for maximization feature : self user video is not maximized on double click

mejdoo:
### Issues

On a video group with 3 enabled videos,  the self user video is not maximized on double click when the call started.

https://wearezeta.atlassian.net/browse/SQCALL-81

### Causes

The number of participants is checked at the moment of the creation of the view, not at the moment of the double click.

### Solutions

Convert `isMultiParticipant` to a `Signal` and observe its value.

#### APK
[Download build #3013](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3013/artifact/build/artifact/wire-dev-PR3115-3013.apk)
[Download build #3014](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3014/artifact/build/artifact/wire-dev-PR3115-3014.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
@@ -127,10 +127,8 @@ class CallingFragment extends FragmentHelper {
       viewMap = viewMap.updated(participant, userView)
       if (BuildConfig.MAXIMIZE_MINIMIZE_VIDEO) {
 
-        val isMultiParticipant: Signal[Boolean] = controller.otherParticipants.map(_.size > 2)
-
         userView.onDoubleClick.onUi { _ =>
-          isMultiParticipant.head.foreach {
+          controller.otherParticipants.map(_.size > 2).head.foreach {
             case true =>
               showFullScreenVideo(participant)
               clearVideoGrid()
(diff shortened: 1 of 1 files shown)


BradleyWire (approved):

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
```diff
@@ -119,20 +119,22 @@ class CallingFragment extends FragmentHelper {
...
+  private def refreshViews(videoUsers: Seq[Participant], selfParticipant: Participant): Seq[UserVideoView] = {
     def createView(participant: Participant): UserVideoView = returning {
       if (participant == selfParticipant) new SelfVideoView(getContext, participant)
       else new OtherVideoView(getContext, participant)
     } { userView =>
       viewMap = viewMap.updated(participant, userView)
       if (BuildConfig.MAXIMIZE_MINIMIZE_VIDEO) {
+
+        val isMultiParticipant: Signal[Boolean] = controller.otherParticipants.map(_.size > 2)
+
```
makingthematrix, line 131: If this is the only place where this signal is used then you could inline it inside the double click function (`controller.otherParticipants.map(_.size > 2).head.foreach { ... }`). But that's ok.

makingthematrix (approved):

app/src/main/scala/com/waz/zclient/calling/CallingFragment.scala
mejdoo: I tried to put this directly below the class, the fragment was crashing.

gizemb (approved):

PR #3064, opened 2020-10-23
Reoarganize gradle scripts

mejdoo:
## What's new in this PR?

Gradle scripts are placed in deferent folders. I moved them to a single place`/scripts`

#### APK
[Download build #2891](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2891/artifact/build/artifact/wire-dev-PR3064-2891.apk)

PR #3053, opened 2020-10-13
Switch preference automation toggles

BradleyWire:
## What's new in this PR?

JIRA: https://wearezeta.atlassian.net/browse/AN-7103

### Issues

Toggle automation toggle not appearing correctly

### Causes

Lack of attribute define for Android 11 devices 

### Solutions

Add a content description to the Switch for all automation to read the automation tags for all devices 
#### APK
[Download build #2853](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2853/artifact/build/artifact/wire-dev-PR3053-2853.apk)

PR #3118, opened 2020-12-14
feat: Allow session reset from the decryption error message

makingthematrix:
see https://wearezeta.atlassian.net/browse/SQCORE-30

When the user sees an error message that the decryption failed, they should be able to fix that simply by clicking on it, instead of having to go the other user's profile and reset the session manually from there to every device of that other user.

This is the first part of the feature. It consists of a changed error message, the functionality that triggers session reset, and
a new "fixed error" system message which replaces the original error message. The other part of the feature consists of the error dialog, the spinner which appears inside the error message instead of the centre of the screen, and a change to MessageData and its Dao which will makes us able to reset the session only to the device from which the error message came from. (Now we reset sessions to all devices of the given user).
To change the error into the fixed error I made a change in how MessageViewPart accesses MessageData. It would be good to check if the new way doesn't slow down the app too much. If yes, I have an idea how to fix it.

I also used this opportunity to refactor the code a bit. Message view parts are old code - I wante to improve its readibility.
And I found that there is some unused code left after the recent changes in audio/video functionality.
#### APK
[Download build #3053](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3053/artifact/build/artifact/wire-dev-PR3118-3053.apk)
[Download build #3057](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3057/artifact/build/artifact/wire-dev-PR3118-3057.apk)
[Download build #3064](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3064/artifact/build/artifact/wire-dev-PR3118-3064.apk)
[Download build #3077](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3077/artifact/build/artifact/wire-dev-PR3118-3077.apk)
[Download build #3078](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3078/artifact/build/artifact/wire-dev-PR3118-3078.apk)
[Download build #3079](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3079/artifact/build/artifact/wire-dev-PR3118-3079.apk)
[Download build #3081](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3081/artifact/build/artifact/wire-dev-PR3118-3081.apk)


app/src/main/scala/com/waz/zclient/calling/views/UserVideoView.scala
```diff
@@ -83,14 +70,6 @@ abstract class UserVideoView(context: Context, val participant: Participant) ext
       infos   <- if (isGroup) callController.participantInfos else Signal.const(Vector.empty)
     } yield infos.find(_.id == participant.userId)
 
-  protected val nameTextView = returning(findById[TextView](R.id.name_text_view)) { view =>
```
mejdoo, line 86: I wonder why did you remove this? It' not related to sessions reset and also will broke the calling experience. Could you please address Scala styling issue outside of this PR?

app/src/main/scala/com/waz/zclient/calling/views/UserVideoView.scala
```diff
@@ -49,31 +48,19 @@ abstract class UserVideoView(context: Context, val participant: Participant) ext
...
-  private val pictureId: Signal[Picture] = for {
-    z             <- callController.callingZms
-    Some(picture) <- z.usersStorage.signal(participant.userId).map(_.picture)
-  } yield picture
-
-  protected val audioStatusImageView = findById[ImageView](R.id.audio_status_image_view)
-
-
-  protected val pausedText = findById[TextView](R.id.paused_text_view)
+  this.onClick(() => callController.controlsClick(true), () => onDoubleClick ! {})
```
mejdoo, line 51: is this exactly equivalent to ?

```
this.onClick({
     callController.controlsClick(true)
   }, {
     onDoubleClick ! {}
   })
```

mejdoo (changes requested):

app/src/main/scala/com/waz/zclient/calling/views/UserVideoView.scala
mejdoo: It could be refactored by we shouldn't remove this, otherwise the name of the users won't be shown correctly in the gridView.

app/src/main/scala/com/waz/zclient/calling/views/UserVideoView.scala
```diff
@@ -49,31 +48,19 @@ abstract class UserVideoView(context: Context, val participant: Participant) ext
 
   val onDoubleClick = EventStream[Unit]()
 
-  this.onClick({
-    callController.controlsClick(true)
-  }, {
-    onDoubleClick ! {}
-  })
-
-  private val pictureId: Signal[Picture] = for {
```
mejdoo, line 58: right, this one is no more used 👍

build.gradle
```diff
@@ -38,8 +38,9 @@ allprojects {
 }
 
 ext {
-    activeSpeakers = true
+    activeSpeakers = false
```
mejdoo, line 41: please keep it true for testing purposes...

mejdoo (changes requested):

storage/src/main/kotlin/com/waz/zclient/storage/db/UserDatabase.kt
```diff
@@ -104,13 +105,14 @@ abstract class UserDatabase : RoomDatabase() {
...
     companion object {
-        const val VERSION = 129
+        const val VERSION = 130
 
         @JvmStatic
         val migrations = arrayOf(
-                USER_DATABASE_MIGRATION_126_TO_127,
-                USER_DATABASE_MIGRATION_127_TO_128,
-                USER_DATABASE_MIGRATION_128_TO_129
+            USER_DATABASE_MIGRATION_126_TO_127,
```
mejdoo, line 112: I think you can rid of 👍 
```
USER_DATABASE_MIGRATION_126_TO_127,
USER_DATABASE_MIGRATION_127_TO_128,
USER_DATABASE_MIGRATION_128_TO_129
```

just keep  `USER_DATABASE_MIGRATION_129_TO_130`

mejdoo (changes requested):

app/src/main/scala/com/waz/zclient/calling/views/UserVideoView.scala
makingthematrix: Right. This is the case.

I think I remember when, in another place in the code, I wondered what to do with a field that wasn't used anywhere in the code, but was necessary to initialize the layout. Android Studio claims such fields are unused and it's easy to make a mistake and delete them. I'll bring it back.

app/src/main/scala/com/waz/zclient/calling/views/UserVideoView.scala
makingthematrix: Yes. It's just a cosmetic change. I thought the new version is easier to read. It shows right away that we pass a function as an argument. The older version could be understood as if we were passing the results of those two operations.

storage/src/main/kotlin/com/waz/zclient/storage/db/UserDatabase.kt
makingthematrix: I will do this later, in https://wearezeta.atlassian.net/browse/SQCORE-371 .

mejdoo (approved):

makingthematrix:
build again

Issue #3022, opened 2020-09-14
Sharing files from filesystem (external app) does not work any more


kroerig:
**Describe the bug**
Sharing files from filesystem does not work any more. No error message.

**To Reproduce**

1. Go to file browser app (Sony "Dateien" in my case)
2. Select a file you want to share
3. Share it to wire
4. Select receipient(s) in wire
5. Send the file
6. File does not get sent.

**Expected behavior**
The file should be sent to all selected receipients.

**Smartphone (please complete the following information):**
- Wire version that exhibits the issue: 3.54.938
- Last Wire version that did not exhibit the issue (if applicable): na
- Android OS version used to run Wire: 9
- Is this a custom firmware or a stock one: Sony stock
- Mobile phone model/manufacturer: Sony Xperia ZX1c
- Mobile network type (EDGE/LTE/Wi-Fi/Offline): any, tested on WiFi

** Work-a-round**
Attachting the file within wire works.

widardd:
Can confirm and your workaround works!

makingthematrix:
@widardd : And I can confirm I have exactly the same problems as you... :( Right now we are bumping Android SDK to 29, which among other things consists of a new API for sharing files. In the next months we will probably work on this issue and we will try to rewrite the code to the new API at the same time. No point in fixing it with the old one now.

kroerig:
any news on this? I still does not work. This is not very satisfing as sharing images is one of the main features of an IM.

kroerig:
*push*

nidico:
This is fixed in 3.62 - see https://medium.com/wire-news/android-3-62-6722b29d45c0 - I can confirm it's working now.

Issue #3021, opened 2020-09-14
Screen goes black and phone is unresponsive when in a Wire call


maltfield:
**Describe the bug**
Every time I'm in a Wire call, as soon as I press the "call" button (or accept someone else's call), my screen goes totally black and none of the soft buttons work.

In the worst case, the phone-call connection becomes degraded/broken, the person on the other end doesn't cancel the call, and the only thing I can do to gain control of my phone again is to hold down the power button to force a hard reboot.

**To Reproduce**

1. Put piece of opaque tape or sticker over your front-facing camera (so that it covers both the camera lens and--by accident--the proximity sensor also)
2. Make a phone call

**Expected behavior**

I should be able to use my phone in a wire call. At the very least, I should be able to exit the call without having to hard reboot.

If the proximity sensor is disabling the screen (perhaps unknown to the privacy-conscious user who covered their front-facing camera with tape), then Wire should have some sort of UI notification to let the user know that the screen has gone black *because* the proximity sensor, and it should hint that they may have accidentally covered it when covering the camera.

That, and there should also be a way to disable the "automatically disable the screen via proximity sensor" functionality in the Wire settings.

**See Also**

 * https://github.com/wardellbagby/Sensor-Disabler

maltfield:
I'm creating this bug now that I know the cause of the issue, but it took me a few months to figure it out. I was ripping my hair out thinking it was a bug with Wire and perhaps some edge-case with my device.

It wasn't until today that I realized it affected other apps. That's when I peeled-off the tape that always covers my front-facing camera and discovered the root cause of this issue.

Considering the userbase of the Wire app, I expect that this is a bug that will probably affect a lot of other folks too. Especially for people whose handset have the proximity sensor right next to their front-facing camera lens.

My suggestion:

1. There must be an option added to the Wire Settings screen to disable the "feature" for folks who always leave their front-facing camera covered and don't want their phones to lock-up when they're in a call.

2. Improve the UI so that the UX isn't so shitty for people who aren't aware that they covered-up their proximity sensor when they actually just meant to cover the front-facing camera. Otherwise, these folks will probably uninstall Wire thinking it's unusable & buggy.

maltfield:
Also, as is common in other apps, the screen should never be disabled (no matter what the proximity sensor reads) when the headphones are plugged-in.

makingthematrix:
@maltfield :
[...]

Issue #3020, opened 2020-09-13
Do not send notifications when I'm actively chatting on webapp


maximbaz:
While I'm actively chatting on webapp, I also receive every single reply message on my Android phone, which is lying nearby with screen off, and which in turn sends notifications to my smart watch, constantly and unnecessarily draining power from both phone and smart watch.

I would like notifications to arrive to the phone only when I am not actively using webapp. Maybe the simplest rule is that I don't read any new message on the desktop within... say 30 or 60 seconds. Or if you can detect that I closed browser / wire webapp tab / put laptop to sleep. But the simple 30 seconds rule would be good enough.

Important note about *any* message, I can be chatting with two people at the same time, and if one person replied to me while I'm replying to the second person on webapp, I still do not need notifications on my phone.

I also tried using "Maintain persistent connection to Wire server" in the app because I thought maybe having two different mechanisms at delivering notifications, but the issue is equally reproducible.

**Smartphone (please complete the following information):**
- Wire version that exhibits the issue: 3.54.938
- Last Wire version that did not exhibit the issue (if applicable): N/A
- Android OS version used to run Wire: 10
- Is this a custom firmware or a stock one: Stock
- Mobile phone model/manufacturer: Samsung A50
- Mobile network type (EDGE/LTE/Wi-Fi/Offline): LTE/Wi-Fi

widardd:
I have to set my smartwatch to "do not disturb" when at home due to this, I like your suggestion!

makingthematrix:
@maximbaz :
For encryption/decryption to work it's necessary that events come to the device in order. So it's not possible **not** to send an event to your phone. We discussed the possibility to delete notifications the users already received, basing on the knowledge that they were read on another device, but that proved to be too costly - Android is very strict about how much can we do while the app is in the background.

However, since the 3.78 release the notifications were reimplemented and now it should be easier to remove notifications, just opening the app make it sync with the server and notifications are removed.

PR #3121, opened 2020-12-16
fix: NetworkModeService crashes on Android 11

makingthematrix:
This fix uses a network callback to listen to changes in the network type. It avoids using NetworkInfo, TelephonyManager, and ConnectivityManager.getActiveNetwork.The only permission we need is ACCESS_NETWORK_STATE. We ask for it when the user opens the app for the first time after the update or installation, just before we register the network callback.

Instead of many types of cellular network we now have only one mode: CELLULAR. Because of this I removed code which warned the user that the network is slow. It checked if we are on 2G and EDGE networks. Anyway, it was only a partial solution - it was still possible for the user to be on 3G or 4G and have poor bandwidth.

There is an old method in NetworkModeService: `override def isDeviceIdleMode: Boolean = false`. I removed comments around it, but left it as it is for this fix. I want to change it to something more meaningful in another PR.

#### APK
[Download build #3020](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3020/artifact/build/artifact/wire-dev-PR3121-3020.apk)
[Download build #3021](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3021/artifact/build/artifact/wire-dev-PR3121-3021.apk)
[Download build #3022](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3022/artifact/build/artifact/wire-dev-PR3121-3022.apk)
[Download build #3023](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3023/artifact/build/artifact/wire-dev-PR3121-3023.apk)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -59,8 +59,6 @@
     <uses-permission android:name="android.permission.USE_BIOMETRIC" />
     <uses-permission android:name="android.permission.USE_FINGERPRINT" />
 
-    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
-
     <uses-feature
         android:name="android.hardware.camera"
         android:required="false"
--- a/zmessaging/src/main/scala/com/waz/service/NetworkModeService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/NetworkModeService.scala
@@ -23,64 +23,55 @@ import android.provider.Settings
 import com.waz.api.NetworkMode
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
 import com.waz.log.LogSE._
-import com.waz.permissions.PermissionsService
-import com.waz.threading.Threading
 import com.waz.utils.returning
 import com.wire.signals.Signal
-
-import scala.collection.immutable.ListSet
-import scala.concurrent.Future
 
 trait NetworkModeService {
   def networkMode: Signal[NetworkMode]
 
-  def registerNetworkCallback(): Future[Unit]
+  def registerNetworkCallback(): Unit
   def isDeviceIdleMode: Boolean
 
   lazy val isOnline: Signal[Boolean] = networkMode.map(NetworkModeService.isOnlineMode)
 }
 
-class DefaultNetworkModeService(context: Context, lifeCycle: UiLifeCycle, permissionService: PermissionsService)
+class DefaultNetworkModeService(context: Context, lifeCycle: UiLifeCycle)
   extends NetworkModeService with DerivedLogTag {
-  import Threading.Implicits.Background
+  private lazy val connectivityManager =
+    returning(context.getSystemService(Context.CONNECTIVITY_SERVICE).asInstanceOf[ConnectivityManager]) { connectivityManager =>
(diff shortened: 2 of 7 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -59,8 +59,6 @@
     <uses-permission android:name="android.permission.USE_BIOMETRIC" />
     <uses-permission android:name="android.permission.USE_FINGERPRINT" />
 
-    <uses-permission android:name="android.permission.READ_PHONE_STATE" />
```
mejdoo, line 62: 👍

zmessaging/src/main/scala/com/waz/service/NetworkModeService.scala
```diff
@@ -17,118 +17,83 @@
...
-        val mode = Option(connectivityManager.getActiveNetworkInfo) match {
-          case Some(info) if info.isConnected => computeMode(info, telephonyManager)
-          case _                              => NetworkMode.OFFLINE
-        }
-        verbose(l"updateNetworkMode: $mode")
-
-        networkMode ! mode
-      }
+  override def registerNetworkCallback(): Future[Unit] =
+    permissionService.requestAllPermissions(ListSet(android.Manifest.permission.ACCESS_NETWORK_STATE)).map {
```
mejdoo, line 52: you don't need to request this permission `ACCESS_NETWORK_STATE` : it's granted automatically because it's a normal permission not dangerous. 
https://developer.android.com/reference/android/Manifest.permission#ACCESS_NETWORK_STATE

mejdoo (changes requested):

mejdoo (approved):

zmessaging/src/main/scala/com/waz/service/NetworkModeService.scala
```diff
@@ -17,118 +17,73 @@
...
-      case true => {
-        val mode = Option(connectivityManager.getActiveNetworkInfo) match {
-          case Some(info) if info.isConnected => computeMode(info, telephonyManager)
-          case _                              => NetworkMode.OFFLINE
+        override def onLost(network: Network): Unit =
+          if (Settings.Global.getInt(context.getContentResolver, Settings.Global.AIRPLANE_MODE_ON, 0) != 0) {
+            info(l"new network mode: OFFLINE (the airplane mode is on)")
+            networkMode ! NetworkMode.OFFLINE
+          } else {
+            val mode = computeMode(network)
```
gizemb, line 56: what does `onLost` mean?

also, in the documentation it says 
```
 * <p>Do NOT call {@link #getNetworkCapabilities(Network)} or
         * {@link #getLinkProperties(Network)} or other synchronous ConnectivityManager methods in
         * this callback as this is prone to race conditions ; calling these methods while in a
         * callback may return an outdated or even a null object.
```
but we call it in `computeMode`🤔

makingthematrix: We need a way to switch to OFFLINE. I tested it manually and it looks like onUnavailable() is never called (I implemented it anyway just to be on the safe side), but onLost(network) is called every time a wifi or a cellular network is turned off.
I got a few times into a situation when I got null as networkCapabilities so I wrapped that call in an option.
But I didn't find a better way to ensure that we react to going offline.

gizemb (approved):

PR #3122, opened 2020-12-17
[Calling] Feature : Active speaker - part 2

mejdoo:
This is a follow up for the calling feature **Active Speaker Detection** introduced in #3106 

## What's new ?

- Highlight the top 4 active speakers having their videos enabled during a conference call.
- Switching between All participants and active speakers using a toggle in call controls.
- Manage the case when there is no active speakers

Design spec : https://zpl.io/bAoEQeP
#### APK
[Download build #3025](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3025/artifact/build/artifact/wire-dev-PR3122-3025.apk)
[Download build #3031](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3031/artifact/build/artifact/wire-dev-PR3122-3031.apk)
[Download build #3033](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3033/artifact/build/artifact/wire-dev-PR3122-3033.apk)

PR #3126, opened 2020-12-21
[Calling] : Show only the loudest active speaker in participants list

mejdoo:
## What's new in this PR?

As a follow up of #3122, there was a request to show only the loudest speaker in participants list instead of many active speakers. This does not apply for the video grid.

https://wearezeta.atlassian.net/browse/SQCALL-43
#### APK
[Download build #3041](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3041/artifact/build/artifact/wire-dev-PR3126-3041.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
@@ -178,10 +178,11 @@ class CallController(implicit inj: Injector, cxt: WireContext)
     })
 
   def isTopSpeaker(userId: UserId, clientId: ClientId): Signal[Boolean] = activeSpeakers.map { speakers =>
-    val topSpeaker = speakers.maxBy(_.audioLevel)
-    topSpeaker.clientId == clientId && topSpeaker.userId == userId
+    if (speakers.nonEmpty) {
+      val topSpeaker = speakers.maxBy(_.audioLevel)
+      topSpeaker.clientId == clientId && topSpeaker.userId == userId
+    } else false
   }
-
 
   val flowManager = callingZms.map(_.flowmanager)
 
--- a/app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
@@ -111,7 +111,7 @@ class CallController(implicit inj: Injector, cxt: WireContext)
   lazy val cbrEnabled                 = currentCall.map(_.isCbrEnabled)
   private lazy val duration           = currentCall.flatMap(_.durationFormatted)
   lazy val selfParticipant            = currentCall.map(_.selfParticipant)
-  lazy val otherParticipants          = currentCall.map(_.otherParticipants)
+  lazy val allParticipants          = currentCall.map(_.allParticipants)
   lazy val activeSpeakers             = currentCall.map(_.activeSpeakers)
 
 
@@ -152,7 +152,7 @@ class CallController(implicit inj: Injector, cxt: WireContext)
   lazy val participantInfos: Signal[Vector[CallParticipantInfo]] =
     for {
       cZms         <- callingZms
-      participants <- otherParticipants.map(_.toSeq)
+      participants <- allParticipants.map(_.toSeq)
       ids           = participants.map(_.userId)
       users        <- cZms.usersStorage.listSignal(ids)
       videoStates  <- mergedVideoStates
@@ -176,6 +176,12 @@ class CallController(implicit inj: Injector, cxt: WireContext)
     activeSpeakers.map(_.exists { activeSpeaker =>
       activeSpeaker.clientId == clientId && activeSpeaker.userId == userId && activeSpeaker.audioLevel > 0
     })
+
+  def isTopSpeaker(userId: UserId, clientId: ClientId): Signal[Boolean] = activeSpeakers.map { speakers =>
+    val topSpeaker = speakers.maxBy(_.audioLevel)
+    topSpeaker.clientId == clientId && topSpeaker.userId == userId
+  }
+
 
   val flowManager = callingZms.map(_.flowmanager)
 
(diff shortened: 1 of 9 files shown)


app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
```diff
@@ -101,15 +101,15 @@ class SingleUserRowView(context: Context, attrs: AttributeSet, style: Int)
...
   private val isMuted = Signal(false)
 
-  private val activeSpeakerData = Signal(Option.empty[(UserId, ClientId)])
-  private val isActiveSpeaker = for {
-    Some((userId, clientId)) <- activeSpeakerData
-    isActive                 <- callController.isActiveSpeaker(userId, clientId)
+  private val topSpeakerData = Signal(Option.empty[(UserId, ClientId)])
+  private val isTopSpeaker = for {
+    Some((userId, clientId)) <- topSpeakerData
+    isActive                 <- callController.isTopSpeaker(userId, clientId)
```
makingthematrix, line 107: Please change the name `isActive` to `topSpeaker`.

makingthematrix (changes requested):

zmessaging/src/main/scala/com/waz/service/call/CallInfo.scala
```diff
@@ -93,7 +93,7 @@ case class CallInfo(convId:             ConvId,
     case (s,             _)                  => s
   }
 
-  def canOthersDialogue: Boolean = otherParticipants.size > 1
+  def canOthersDialogue: Boolean = allParticipants.size > 1
```
gizemb, line 96: doesn't the logic change here? (though I'm not sure what `canOthersDialogue` means)

app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
```diff
@@ -219,7 +226,7 @@ class CallController(implicit inj: Injector, cxt: WireContext)
   val conversationName: Signal[Name] = zmsConvId.flatMap { case (z, cId) => z.conversations.conversationName(cId) }
   val conversationMembers: Signal[Map[UserId, ConversationRole]] = zmsConvId.flatMap { case (z, cId) => z.conversations.convMembers(cId) }
 
-  private lazy val otherUser = Signal.zip(isGroupCall, userStorage, otherParticipants.map(_.toSeq.headOption)).flatMap {
+  private lazy val otherUser = Signal.zip(isGroupCall, userStorage, allParticipants.map(_.toSeq.headOption)).flatMap {
```
gizemb, line 229: this doesn't break anything, right? the variable name is `otherUser`, so I got suspicious

makingthematrix: Right. I always thought that `otherParticipants` here don't contain the self user. If they do, there is a risk that `otherUser` will be the same as the self user. @mejdoo , could you check it?

app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
mejdoo: ok done

zmessaging/src/main/scala/com/waz/service/call/CallInfo.scala
mejdoo: No change for the logic here. I changed the name because it's reflecting a wrong thing. `other` means `allParticipants` except `selfParticipant`.

app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
mejdoo: Let me double check again and give you some results of AVS api

mejdoo: I can confirm it contains the self user after checking the logs of AVS (tested with group calls of 3 participants)

makingthematrix: Okay, so that means we have a possible bug here. Just to be on the safe side, please filter the self user out:
```
private lazy val otherUser = Signal.zip(isGroupCall, userStorage, allParticipants, selfParticipant).flatMap {
  case (false, usersStorage, participants, self) => 
      participants.find(_.userId != self.userId) match {
        case Some(participant) => usersStorage.optSignal(participant.userId)
        case None              => Signal.const[Option[UserData]](None)
      }
  case _ => Signal.const[Option[UserData]](None)
}
```

makingthematrix (changes requested):

gizemb:
@mejdoo does `otherUser` work properly with this change in **everywhere** it is used?
if yes, then we should simply change the name from `otherUser` to something else, like "user".

Because when talking about participants, "other" refers to any participant who is not the self user. This naming could create confusion for people who don't check how it's computed.

makingthematrix:
> @makingthematrix filtering the self user may make the calling features working improperly. The self user will not be displayed in participants list + `videoGrid` and `previewCardView` won't behave properly.

I mean only filtering the self user out for the `otherUser` signal. In other cases, the self user should stay.

mejdoo:
> any participant who

@gizemb yes, I changed this everywhere to `AllParticipants` instead of `otherParticipants`

mejdoo:
@makingthematrix @gizemb Sorry I misunderstood you in my previous comments. I got you meant, I will change the code as @makingthematrix requested.

app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
mejdoo: Good catch @makingthematrix 👍

mejdoo:
> @mejdoo does `otherUser` work properly with this change in **everywhere** it is used?
> if yes, then we should simply change the name from `otherUser` to something else, like "user".
> 
> Because when talking about participants, "other" refers to any participant who is not the self user. This naming could create confusion for people who don't check how it's computed.

@gizemb sorry for the misunderstood. For this specific signal, `otherUser` refers to any participant who is not the self user.

makingthematrix (approved):

app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
```diff
@@ -101,15 +101,15 @@ class SingleUserRowView(context: Context, attrs: AttributeSet, style: Int)
 
   private val isMuted = Signal(false)
 
-  private val activeSpeakerData = Signal(Option.empty[(UserId, ClientId)])
-  private val isActiveSpeaker = for {
-    Some((userId, clientId)) <- activeSpeakerData
-    isActive                 <- callController.isActiveSpeaker(userId, clientId)
-  } yield isActive
+  private val topSpeakerData = Signal(Option.empty[(UserId, ClientId)])
```
gizemb, line 104: shouldn't this stay as `activeSpeakerData`? I agree with `isTopSpeaker` change because it performs a check for top speaker, but seems like this signal emits values regardless of whether they're top speaker or not

gizemb (approved):
I approved but had a question

app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
mejdoo: I'm fine with reverting variable names if @makingthematrix agrees as well.

Issue #3060, opened 2020-10-19
Can't save backup as a file


missytake:
**Describe the bug**
After saving the message history in a backup, I'm asked with which app I want to share it; but saving it to my device as a file isn't possible.

**To Reproduce**
1. Go to the Account Settings
2. Click on 'Back Up Conversations"
3. Click on 'Back Up Now'
4. Set a Passphrase
5. Wait while it's backing up
6. See a prompt: 'Share With' (see screenshots above)

**Expected behavior**
One of the options should be "save to device", where I can then choose a location to store the file.

**Screenshots**

![Screenshot_20201019-171428_Android_System](https://user-images.githubusercontent.com/57442838/96474468-2054ea00-1233-11eb-9e6e-56ee6aa57947.png)

![Screenshot_20201019-182609_Android_System](https://user-images.githubusercontent.com/57442838/96479470-48dfe280-1239-11eb-8809-174a2f6853bd.png)

**Smartphone (please complete the following information):**
- Wire version that exhibits the issue: 3.48.915
- Android OS version used to run Wire: 9
- Is this a custom firmware or a stock one: LineageOS
- Mobile phone model/manufacturer: SM-G900F
- Mobile network type (EDGE/LTE/Wi-Fi/Offline): Wi-Fi

r-a-y:
I experienced this problem myself and worked around this by installing the Share by File System app: https://play.google.com/store/apps/details?id=com.boxhead.android.sharetofilesystem&hl=en&gl=US. It's an older app, but still works even on Android 13.

PR #3137, opened 2021-01-08
feat: Feature Flag App Lock

makingthematrix:
https://wearezeta.atlassian.net/browse/SQSERVICES-58

Show App-lock feature setting in Wire account setting based on App-lock feature status.
The feature requires communication with a new endpoint for checking the feature flag status  every time when the app goes to the foreground.
If the flag is enabled, we use the returned config to check if the app lock is enforced and what is the timeout.
Contrary to the "forced" setting in the build config, if the app lock is forced by the feature flag, we don't wipe the data clean if the user fails to validate.
If the flag is disabled, local settings apply, just as until now.
The feature flags work only for team accounts. If the account is personal, we act as if the flag was disabled.
The PR also contains some refactoring - I moved the AppLockEnabled preference from GlobalPreferences to UserPreferences, I created user preferences for the "forced" setting and for the timeout, and I created a temporary flag in GlobalPreferences to migrate AppLockEnabled: If the user had AppLockEnabled set to true until now, that setting should be copied to UserPreferences.

#### APK
[Download build #3060](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3060/artifact/build/artifact/wire-dev-PR3137-3060.apk)
[Download build #3061](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3061/artifact/build/artifact/wire-dev-PR3137-3061.apk)
[Download build #3062](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3062/artifact/build/artifact/wire-dev-PR3137-3062.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/common/controllers/global/PasswordController.scala
+++ b/app/src/main/scala/com/waz/zclient/common/controllers/global/PasswordController.scala
@@ -49,19 +49,9 @@ class PasswordController(implicit inj: Injector) extends Injectable with Derived
 
   private lazy val sodiumHandler = inject[SodiumHandler]
 
+  // TODO: Remove after everyone migrates to UserPreferences.AppLockEnabled
   if (BuildConfig.APP_LOCK_FEATURE_FLAG) {
-    // TODO: Remove after everyone migrates to UserPreferences.AppLockEnabled (let's say, June 2021?)
-    inject[GlobalPreferences].preference(GlobalPreferences.GlobalAppLockDeprecated).apply().foreach {
-      case true =>
-      case false =>
-        inject[GlobalPreferences].preference(GlobalPreferences.AppLockEnabled).apply().foreach { globalAppLock =>
-          prefs.map(_.preference(AppLockEnabled)).head.foreach { userAppLockPref =>
-            verbose(l"Migrating the AppLockEnabled preference (set to $globalAppLock) from global to user preferences")
-            userAppLockPref := globalAppLock
-            inject[GlobalPreferences].preference(GlobalPreferences.GlobalAppLockDeprecated) := true
-          }
-        }
-    }
+    appLockPrefMigration()
   }
 
   appInBackground.foreach {
@@ -183,4 +173,18 @@ class PasswordController(implicit inj: Injector) extends Injectable with Derived
       case Some(id) =>
         sodiumHandler.hash(password, id.str.replace("-", ""))
     }(Threading.Background)
+
+  // TODO: Remove after everyone migrates to UserPreferences.AppLockEnabled
+  private def appLockPrefMigration() =
+    inject[GlobalPreferences].preference(GlobalPreferences.GlobalAppLockDeprecated).apply().foreach {
+      case true =>
+      case false =>
+        inject[GlobalPreferences].preference(GlobalPreferences.AppLockEnabled).apply().foreach { globalAppLock =>
+          prefs.map(_.preference(AppLockEnabled)).head.foreach { userAppLockPref =>
+            verbose(l"Migrating the AppLockEnabled preference (set to $globalAppLock) from global to user preferences")
+            userAppLockPref := globalAppLock
+            inject[GlobalPreferences].preference(GlobalPreferences.GlobalAppLockDeprecated) := true
+          }
+        }
+    }
 }
--- a/app/src/main/scala/com/waz/zclient/common/controllers/global/PasswordController.scala
+++ b/app/src/main/scala/com/waz/zclient/common/controllers/global/PasswordController.scala
@@ -49,26 +49,30 @@ class PasswordController(implicit inj: Injector) extends Injectable with Derived
 
   private lazy val sodiumHandler = inject[SodiumHandler]
 
-  // TODO: Remove after everyone migrates to UserPreferences.AppLockEnabled (let's say, June 2021?)
-  inject[GlobalPreferences].preference(GlobalPreferences.GlobalAppLockDeprecated).apply().foreach {
-    case true =>
-    case false =>
-      inject[GlobalPreferences].preference(GlobalPreferences.AppLockEnabled).apply().foreach { globalAppLock =>
-        prefs.map(_.preference(AppLockEnabled)).head.foreach { userAppLockPref =>
-          verbose(l"Migrating the AppLockEnabled preference (set to $globalAppLock) from global to user preferences")
-          userAppLockPref := globalAppLock
-          inject[GlobalPreferences].preference(GlobalPreferences.GlobalAppLockDeprecated) := true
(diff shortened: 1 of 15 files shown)


app/src/main/scala/com/waz/zclient/common/controllers/global/PasswordController.scala
```diff
@@ -41,15 +41,46 @@ class PasswordController(implicit inj: Injector) extends Injectable with Derived
...
   private val ssoPassword = prefs.map(_.preference(UserPreferences.SSOPassword))
   private val ssoPasswordIv = prefs.map(_.preference(UserPreferences.SSOPasswordIv))
 
   private lazy val sodiumHandler = inject[SodiumHandler]
 
-  val ssoEnabled:       Signal[Boolean] = accounts.isActiveAccountSSO
-  val ssoPasswordEmpty: Signal[Boolean] = ssoPassword.flatMap(_.signal.map(_.isEmpty))
-  val appLockEnabled:   Signal[Boolean] = inject[GlobalPreferences].preference(AppLockEnabled).signal
-  lazy val password:    Signal[Option[Password]] = accounts.activeAccount.map(_.flatMap(_.password)).disableAutowiring()
+  // TODO: Remove after everyone migrates to UserPreferences.AppLockEnabled (let's say, June 2021?)
```
mejdoo, line 52: We can maybe decide later about this after checking active users analytics in Google Play.

zmessaging/src/main/scala/com/waz/content/Preferences.scala
```diff
@@ -414,15 +414,16 @@ object GlobalPreferences {
 
   lazy val RootDetected: PrefKey[Boolean] = PrefKey[Boolean]("root_detected", customDefault = false)
 
-  lazy val AppLockEnabled: PrefKey[Boolean] = PrefKey[Boolean]("app_lock_enabled", customDefault = false)
-
   lazy val IncognitoKeyboardEnabled: PrefKey[Boolean] = PrefKey[Boolean]("incognito_keyboard_enabled", customDefault = false)
 
   //DEPRECATED!!! Use the UserPreferences instead!!
```
mejdoo, line 419: Can we get rid of the deprecated prefs ?

zmessaging/src/main/scala/com/waz/content/Preferences.scala
```diff
@@ -414,15 +414,16 @@ object GlobalPreferences {
...
   lazy val _DarkTheme = PrefKey[Boolean]("DarkTheme")
   lazy val _SoundsPrefKey = PrefKey[String]("PREF_KEY_SOUND")
   lazy val _DownloadImages = PrefKey[String]("zms_pref_image_download")
 
+  //deprecated
```
mejdoo, line 424: Why do we need to create a new `PrefKey` for the deprecated app lock?

makingthematrix: Because the app needs to know if the migration was done already. The whole process looks like this:
1. The current version of the app has only the global `AppLockEnabled` pref. The user could have already switched it on.
2. The app is updated and started for the first time after the update.
3. The code in PasswordController checks that the migration was not performed yet, so it copies the value of the global `AppLockEnabled` to the user's `AppLockEnabled` and sets `GlobalAppLockDeprecated` to true.
4. The next time the app is started, `PasswordManager` sees that `GlobalAppLockDeprecated` is true so it does not have to copy the pref again.
5. After some time we can remove both the global `AppLockEnabled` and `GlobalAppLockDeprecated`. Everyone should have the pref copied to the user's `AppLockEnabled` by then.
6. If we instead decide to fall back and remove the "app lock feature flag" feature, the global `AppLockEnabled` will still be there with the value originally set by the user.

makingthematrix:
@android10 : Here's the PR for "the app lock feature flag" feature.

app/src/main/scala/com/waz/zclient/common/controllers/global/PasswordController.scala
```diff
@@ -41,15 +41,46 @@ class PasswordController(implicit inj: Injector) extends Injectable with Derived
...
+  inject[GlobalPreferences].preference(GlobalPreferences.GlobalAppLockDeprecated).apply().foreach {
```
gizemb, line 53: can we move this piece of code inside a method for readability purposes?

app/src/main/scala/com/waz/zclient/common/controllers/global/PasswordController.scala
```diff
@@ -41,15 +41,46 @@ class PasswordController(implicit inj: Injector) extends Injectable with Derived
...
+
+  val ssoEnabled:       Signal[Boolean]                = accounts.isActiveAccountSSO
+  val ssoPasswordEmpty: Signal[Boolean]                = ssoPassword.flatMap(_.signal.map(_.isEmpty))
+  val appLockEnabled:   Signal[Boolean]                = prefs.flatMap(_.preference(AppLockEnabled).signal)
+  val appLockForced:    Signal[Boolean]                = prefs.flatMap(_.preference(AppLockForced).signal)
+  lazy val password:    Signal[Option[Password]]       = accounts.activeAccount.map(_.flatMap(_.password)).disableAutowiring()
+  val appLockTimeout: Signal[Int] = prefs.flatMap(_.preference(AppLockTimeout).signal).map {
+    case None           => BuildConfig.APP_LOCK_TIMEOUT
+    case Some(duration) => duration.toSeconds.toInt
+  }
```
gizemb, line 83: does it make sense to put all password related things inside a single `PasswordController` class?
it seems like sso password logic and app lock password logic have very little in common. I personally would move app lock related stuff to a separate class. it would (potentially) make the app lock logic here more testable.

gizemb (changes requested):

zmessaging/src/main/scala/com/waz/content/Preferences.scala
makingthematrix: done

app/src/main/scala/com/waz/zclient/common/controllers/global/PasswordController.scala
makingthematrix: They are closely related. We use a lot of common code in both cases. I'm rather against separating it and definitely not in this PR.

app/src/main/scala/com/waz/zclient/common/controllers/global/PasswordController.scala
makingthematrix: ok

app/src/main/scala/com/waz/zclient/common/controllers/global/PasswordController.scala
gizemb: ok, if you say so 🙂

gizemb (approved):

PR #3093, opened 2020-11-13
Cleanup : Remove deprecated registration code

mejdoo:
## What's new in this PR?

Let's get rid of the dead code :)
#### APK
[Download build #2953](http://10.10.124.11:8080/job/Pull%20Request%20Builder/2953/artifact/build/artifact/wire-dev-PR3093-2953.apk)

PR #3154, opened 2021-01-25
[Calling] Hide active speakers/all toggle

mejdoo:
## What's new in this PR?

The calling squad decided to disable all/speaker toggle to perform some optimizations in terms of user experience before releasing it to public. The code is now commented till further notice.
#### APK
[Download build #3093](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3093/artifact/build/artifact/wire-dev-PR3154-3093.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/calling/ControlsFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/ControlsFragment.scala
@@ -94,21 +94,22 @@ class ControlsFragment extends FragmentHelper {
       }
     }
 
-    if (BuildConfig.ACTIVE_SPEAKERS) {
-      Signal.zip(
-        controller.isCallEstablished,
-        controller.isGroupCall,
-        controller.isVideoCall,
-        controller.isFullScreenEnabled
-      ).onUi {
-        case (true, true, true, false) => speakersLayoutContainer.foreach(_.setVisibility(View.VISIBLE))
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/calling/ControlsFragment.scala
```diff
@@ -94,21 +94,22 @@ class ControlsFragment extends FragmentHelper {
...
-      }
-    }
-    else {
-      speakersLayoutContainer.foreach(_.setVisibility(View.INVISIBLE))
-    }
-
+    //TODO : The calling squad decided to disable all/speaker toggle to perform some optimizations
+    // in terms of user experience before releasing it to public
+
+    /* if (BuildConfig.ACTIVE_SPEAKERS) {
```
gizemb, line 100: why didn't you just change the flag?

mejdoo: Because we want to release the active speaker feature except this toggle.

makingthematrix (approved):

app/src/main/scala/com/waz/zclient/calling/ControlsFragment.scala
makingthematrix: @mejdoo : Please comment out also the unused imports.

mejdoo: ok good catch @makingthematrix

Issue #3132, opened 2020-12-31
Can't running wire android: error in jacoco.gradle


andriansandi:
I'm experiencing an issue when running the wire-android using Android Studio in Windows 10

- Gradle Version: 6.5
- Android Gradle Version: 3.6.0

The error says:
````A problem occurred evaluating script.
> Cannot set the value of read-only property 'sourceDirectories' for task ':app:coverage' of type org.gradle.testing.jacoco.tasks.JacocoReport.```

Any missing steps ?

PR #3173, opened 2021-02-09
Remove old code

makingthematrix:
I was looking through the `macrosupport` code to see if we still need it. It turns out that yes, but only because of one small macro - `returning`. The rest of the code is not used anymore. I removed it, and I removed and simplified code in other classes which called code from `macrosupport` but did nothing with it.
#### APK
[Download build #3132](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3132/artifact/build/artifact/wire-dev-PR3173-3132.apk)


macrosupport/src/main/scala/com/waz/utils/returning.scala
```diff
@@ -18,7 +18,7 @@
 package com.waz.utils
 
 import scala.language.experimental.macros
-import scala.reflect.macros.blackbox.Context
+import scala.reflect.macros.blackbox
 
 object returning {
```
mejdoo, line 23: Can we move this class somewhere else and get rid completely of `macrosupport` module?

makingthematrix: @mejdoo : I tried. But this is a macro and apparently it needs to be in a separate module to be compiled before the other ones.

mejdoo (approved):

PR #3180, opened 2021-02-16
fix: Optimize push notifications

makingthematrix:
In Google Play Console we get constant ANRs about that we didn't manage to process an event in time in WebSocketService ("Context.startForegroundService() did not then call Service.startForeground()"). Also, we get reports a lot of reports about that new messages come only after the user opens the app - which means that FCM messages are also not received or not processed in time.

I have read through the code for both web socket and FCM events/notifications and made some optimizations. (And also
it's possible I fixed something with the default notification channel for Android 8.0+ that wasn't working before).
I'm not sure about the results, but it shouldn't be worse than before.
#### APK
[Download build #3144](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3144/artifact/build/artifact/wire-dev-PR3180-3144.apk)

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -289,7 +289,13 @@
             android:value="AIzaSyBA1m9faI3-P4sP6S9B_JH3XTNIxHBz2Qg"/>
 
         <service
-            android:name="com.waz.services.fcm.FCMHandlerService">
+            android:name="com.waz.services.fcm.FCMHandlerService"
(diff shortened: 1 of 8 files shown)


mejdoo (approved):
Let's merge this and test it internally. In worst case, we revert the PR.

Kyoshiro-san:
#3197 
This code might be the culprit of systematic crashes on de-googled phones.

Could you please check and if so, revert and make a quick release ?
Thank you so much !

PR #3147, opened 2021-01-15
Fix: Asset selection intent for developer mode

gizemb:
### Issues

If app is in development mode, and Testing Gallery is installed on the phone, Testing Gallery should be opened instead of phone's own file storage.

For Android 10 and later, this functionality is broken and Testing Gallery is never opened.

### Causes

After Android 10, `context.getPackageManager().queryIntentActivities(intent, PackageManager.MATCH_ALL)` check always returns an empty list. A similar problem was experienced before for Video sharing and fixed similarly on #3111 . Please refer to that PR for detailed explanation.

### Solutions

Return a boolean to denote whether the app is opened via callback.
### Testing

Steps:
1. Open a conversation
2. Select attachment icon in bottom conversation menu

**Scenario 1:**
- Wire Dev or Internal app is installed.
- Testing Gallery is installed

Expected Result:
Testing Gallery should be opened.

**Scenario 2:**
- Wire Dev or Internal app is installed.
- Testing Gallery is not installed

Expected Result:
Phone's Gallery should be opened.

**Scenario 3:**
- Wire Prod app is installed.
- Testing Gallery is installed

Expected Result:
Phone's Gallery should be opened.

**Scenario 4:**
- Wire Prod app is installed.
- Testing Gallery is not installed

Expected Result:
Phone's Gallery should be opened.

## Notes

_Specify here any other facts that you think are important for this issue. (Optional)_

#### APK
[Download build #3074](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3074/artifact/build/artifact/wire-dev-PR3147-3074.apk)

PR #3181, opened 2021-02-18
Changes needed for work with wire-signals-0.4.0

makingthematrix:
Wire Signals 0.4.0 brings a fix for the throttled signal which before sometimes ignored values sent to it from the original signal (https://github.com/wireapp/wire-signals/pull/48) and it splits the signals project into two: wire-signals and wire-signals-extensions. wire-signals will contain the "core" functionality which depends only on the standard Scala library, and "extensions" will be a small (but growing, hopefully) set of event streams and signals built on top of the core. They can use additional dependencies and can be designed for specific tasks while the ones in the core should be rather generic. 

The reasoning behind this is that the core library is enough to perform the main task of event streams and signals, which is to transport events across the app. For some users this might be all they need and so they will be interested more in lack of additional dependencies, small codebase, and performance.

The "extensions" library contains for now only `ClockSignal` and `ButtonSignal` and depends on ThreeTenBP.

#### APK
[Download build #3147](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3147/artifact/build/artifact/wire-dev-PR3181-3147.apk)

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -407,6 +407,7 @@ dependencies {
     implementation BuildDependencies.countly
 
     implementation BuildDependencies.wireSignals
+    implementation BuildDependencies.wireSignalsExtensions
 
     implementation BuildDependencies.libSodium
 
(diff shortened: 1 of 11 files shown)


makingthematrix:
build again

makingthematrix:
build again

mejdoo (approved):

PR #3193, opened 2021-03-01
refactor: Use new protobuf 1.28.2

makingthematrix:
I know this is huge.

The old protobuf implementation used the now obsolete "nano" mode which created very small Java classes which had to be accessed and modified directly, i.e. every field was public and it was up to us to ensure we use it in the right way. "nano" is no longer used in protobuf, so when I updated protobuf, I hade to change to the more standard "lite" implementation. "lite" uses getters and for modification it has builder classes. That changed a lot how
the protobuf classes need to be handled in our `com.waz.model._` classes, esp. `GenericMessage` and `GenericContent`. On top of that, we recently merged two protobuf libraries: generic-message-proto and backend-api-proto. I had to make so many changes that instead of trying to contain them and change as little as possible (not possible) I decided to refactor also the code which touched those classes, so that it all works better together.

Before, we used a lot of implicit objects with `apply` and `unapply` methods for convertions. In many places it meant that we converted the proto data into logic case classes, used them for checking something, then dismissed them, and then we made convertions again when we needed them again. I removed those implicit objects and introduced "newtypes" - wrapper case classes around proto data. Each of them has a `lazy val unpack`, equivalent to the previous `unapply`, but this time after the first convertion, the converted ("unpacked") version is remembered so we don't get the performance
penalty when we need to mae another convertion. Instead, I guess the app will now need a bit more memory. We need to test it.

Also, I made some performance optimizations in `GenericMessageService`. Until now with every new event we created and then left for garbage collection a lot of intermediate collections. I decided it happens so often that it might make sense to use one mutable cache instead. This too needs to be tested. Fortunately, both `unpack` and this can be easily converted to versions which will be slower but maybe safer, so if we find a bug in this implementation, we may try it.
#### APK
[Download build #3209](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3209/artifact/build/artifact/wire-dev-PR3193-3209.apk)
[Download build #3212](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3212/artifact/build/artifact/wire-dev-PR3193-3212.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
+++ b/app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
@@ -85,7 +85,7 @@ class GlobalTrackingController(implicit inj: Injector, cxt: WireContext)
     } yield {
         verbose(l"Using countly Id: ${trackingId.str}")
         val config = new CountlyConfig(cxt, GlobalTrackingController.countlyAppKey, BuildConfig.COUNTLY_SERVER_URL)
-          .setLoggingEnabled(logsEnabled)
+          .setLoggingEnabled(false)
           .setIdMode(DeviceId.Type.DEVELOPER_SUPPLIED)
           .setDeviceId(trackingId.str)
           .setRecordAppStartTime(true)
--- a/app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
+++ b/app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
@@ -85,7 +85,7 @@ class GlobalTrackingController(implicit inj: Injector, cxt: WireContext)
     } yield {
         verbose(l"Using countly Id: ${trackingId.str}")
         val config = new CountlyConfig(cxt, GlobalTrackingController.countlyAppKey, BuildConfig.COUNTLY_SERVER_URL)
-          .setLoggingEnabled(false)
+          .setLoggingEnabled(logsEnabled)
           .setIdMode(DeviceId.Type.DEVELOPER_SUPPLIED)
           .setDeviceId(trackingId.str)
           .setRecordAppStartTime(true)
--- a/zmessaging/src/main/scala/com/waz/service/GenericMessageService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/GenericMessageService.scala
@@ -18,13 +18,15 @@
 package com.waz.service
 
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
-import com.waz.model
+import com.waz.model.GenericContent.{ReadReceipt => GReadReceipt}
+import com.waz.model.{ReadReceipt => MReadReceipt}
 import com.waz.model.GenericContent._
 import com.waz.model._
 import com.waz.service.conversation.{ConversationOrderEventsService, ConversationsContentUpdaterImpl}
 import com.waz.service.messages.{MessagesContentUpdater, ReactionsService, ReceiptService}
 import com.waz.service.tracking.TrackingService
 
+import scala.concurrent.Future
 import scala.concurrent.Future.traverse
 
 class GenericMessageService(selfUserId: UserId,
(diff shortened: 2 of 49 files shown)


makingthematrix:
build again

zmessaging/src/main/scala/com/waz/service/GenericMessageService.scala
```diff
@@ -39,63 +41,86 @@ class GenericMessageService(selfUserId: UserId,
...
+    clearCaches()
+    events.foreach { case GenericMessageEvent(_, time, from, content) =>
+      content.unpackContent match {
+        case r: Reaction =>
+          val (msg, action) = r.unpack
+          incomingReactions += Liking(msg, from, time, action)
+        case lr: LastRead =>
+          lastRead += lr.unpack
+        case c: Cleared =>
+          cleared += c.unpack
```
gizemb, line 81: is `if userId == selfUserId` check missing here?

zmessaging/src/main/scala/com/waz/sync/client/MessagesClient.scala
```diff
@@ -63,14 +65,16 @@ object MessagesClient extends DerivedLogTag {
...
+    import scala.collection.JavaConverters._
+    val builder = Otr.NewOtrMessage.newBuilder()
+    builder.setSender(OtrClient.clientId(m.sender))
+    builder.setNativePush(m.nativePush)
+    builder.addAllRecipients(m.recipients.userEntries.toIterable.asJava)
+    m.external.foreach { ext => builder.setBlob(ByteString.copyFrom(ext)) }
+    m.report_missing.foreach { missing => builder.addAllReportMissing(missing.map(OtrClient.userId).asJava) }
 
-    val bytes = MessageNano.toByteArray(msg)
+    val msg = builder.build()
```
gizemb, line 76: doesn't builder allow method chaining?

zmessaging/src/main/scala/com/waz/sync/handler/MessagesSyncHandler.scala
```diff
@@ -283,7 +306,7 @@ class MessagesSyncHandler(selfUserId: UserId,
               proto = GenericMessage(
                 msg.id.uid,
                 msg.ephemeral,
-                Proto.Asset(uploadAssetOriginal, Some(previewAsset), expectsReadConfirmation = false)
+                GAsset(uploadAssetOriginal, Some(previewAsset), expectsReadConfirmation = false)
```
gizemb, line 309: what's `GAsset`?

zmessaging/src/test/scala/com/waz/sync/handler/UsersSyncHandlerSpec.scala
```diff
@@ -54,7 +53,11 @@ class UsersSyncHandlerSpec extends AndroidFreeSpec {
       // then
       (otrSync.broadcastMessage _).expects(*, *, *, *).once().onCall {
         (message: GenericMessage, _: Int, _: EncryptedContent, recipients: Option[Set[UserId]]) =>
-          message.getAvailability.`type` shouldEqual Messages.Availability.AVAILABLE
+          message.unpackContent match {
+            case content: GenericContent.AvailabilityStatus =>
+              content.proto.getType shouldEqual Messages.Availability.Type.AVAILABLE
+            case _ => fail("Availability should be set to AVAILABLE")
+          }
```
gizemb, line 60: we can write a utility method for that

gizemb (changes requested):
I couldn't review all of it but here are some of my findings so far

makingthematrix:
build again

app/src/main/scala/com/waz/zclient/tracking/GlobalTrackingController.scala
```diff
@@ -85,7 +85,7 @@ class GlobalTrackingController(implicit inj: Injector, cxt: WireContext)
     } yield {
         verbose(l"Using countly Id: ${trackingId.str}")
         val config = new CountlyConfig(cxt, GlobalTrackingController.countlyAppKey, BuildConfig.COUNTLY_SERVER_URL)
-          .setLoggingEnabled(logsEnabled)
+          .setLoggingEnabled(false)
```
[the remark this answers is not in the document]
makingthematrix: Right. I think I changed it at some point because I thought `logsEnabled` is always false, but actually it's not.

makingthematrix:
build again

gizemb (approved):

PR #3194, opened 2021-03-02
Remove bintray

Gardosen:
## What's new in this PR?
bintray is no longer used or supported, there for we have moved all necessary libs for now to our nexus server. (public sonartype follows)

this PR removes the bintray relation and moved the nexus load to the build.gradle script in order to make it possible to fetch the libraries also for the buildscript section.
#### APK
[Download build #3179](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3179/artifact/build/artifact/wire-dev-PR3194-3179.apk)
[Download build #3193](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3193/artifact/build/artifact/wire-dev-PR3194-3193.apk)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,12 +1,36 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    repositories {
-        maven {
-            url "https://dl.bintray.com/wire-android/third-party"
+
+    final String NEXUS_URL_ENV_VAR = "NEXUS_URL"
+    final String NEXUS_URL_LOCAL_VAR = "nexus.url"
+    final String LOCAL_PROPERTIES_FILE_NAME = "local.properties"
+
+    String nexusUrl = ""
+    Properties properties = new Properties()
+    def propertiesFile = project.rootProject.file(LOCAL_PROPERTIES_FILE_NAME)
+    if (propertiesFile.exists()) {
+        properties.load(propertiesFile.newDataInputStream())
+        nexusUrl = System.getenv(NEXUS_URL_ENV_VAR) ?: properties.getProperty(NEXUS_URL_LOCAL_VAR) ?: ""
+    } else {
+        nexusUrl = System.getenv(NEXUS_URL_ENV_VAR) ?: ""
+    }
+
+    if (!nexusUrl.isEmpty()) {
+        buildscript {
+            repositories {
+                maven { url nexusUrl }
+            }
         }
+        allprojects {
+            repositories {
+                maven { url nexusUrl }
+            }
+        }
+    }
+
+    repositories {
         google()
-        jcenter()
         mavenCentral()
     }
     dependencies {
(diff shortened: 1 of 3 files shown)


makingthematrix (approved):

build.gradle
```diff
@@ -1,12 +1,36 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    repositories {
-        maven {
-            url "https://dl.bintray.com/wire-android/third-party"
+
+    final String NEXUS_URL_ENV_VAR = "NEXUS_URL"
```
mejdoo, line 5: @gizemb any thoughts about exporting this logic to another gradle file? Afaik we cannot apply any plugin at this stage.

Gardosen:
build again

Gardosen:
build again

Gardosen:
build again

build.gradle
```diff
@@ -1,12 +1,36 @@
...
+    final String NEXUS_URL_LOCAL_VAR = "nexus.url"
+    final String LOCAL_PROPERTIES_FILE_NAME = "local.properties"
+
+    String nexusUrl = ""
+    Properties properties = new Properties()
+    def propertiesFile = project.rootProject.file(LOCAL_PROPERTIES_FILE_NAME)
+    if (propertiesFile.exists()) {
+        properties.load(propertiesFile.newDataInputStream())
+        nexusUrl = System.getenv(NEXUS_URL_ENV_VAR) ?: properties.getProperty(NEXUS_URL_LOCAL_VAR) ?: ""
```
gizemb, line 14: if this url is empty, where does a person outside the company get the libraries from? (especially the ones that we own, like gradle scala plugin)

gizemb: another question: why do we need a jenkins environment variable and a properties file? can't we do the same thing with only environment variable? the simpler the logic the better imo.

build.gradle
AndroidBob: the problem is in general that we can not add the nexus url to the buildscript section from another file, as the build.gradle file is the first in line. 
and the buildscript section also needs the nexus for loading certain libs.

build.gradle
AndroidBob: the idea is on a second level now to load all the libs we have on our privat nexus, also lto load them to sonartype, so everyone else get's them from sonartype

scripts/avs.gradle
```diff
@@ -23,18 +23,12 @@ if (propertiesFile.exists()) {
 }
```
gizemb, line 24: looks like there's some duplication here. can't we use `nexusUrl` from root build.gradle here?

build.gradle
gizemb: so, until we add sonatype, people outside the company won't be able to compile the code. but I guess that's ok because that's broken now as well. is that right?

build.gradle
gizemb: hmm, I don't have a solution for this right now. I need to try it out myself.

build.gradle
Gardosen: yes that is correct, but we are talking here about some minutes as i have all the required libraries locally prepared and only need to push them to sonartype

scripts/avs.gradle
Gardosen: i will try that out

mejdoo:
@Gardosen when are you planning to merge this PR?

mejdoo (approved):

gizemb (approved):

PR #3211, opened 2021-03-23
Feature: Legal Hold - Indicate legal hold in conversation list

gizemb:
### Issues

Jira: [SQSERVICES-337](https://wearezeta.atlassian.net/browse/SQSERVICES-337)

When the user is under legal hold, we want to show an indicator in the main page,  on top of conversation list.

### Testing

Tested manually w/ mock data.

#### APK
[Download build #3217](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3217/artifact/build/artifact/wire-dev-PR3211-3217.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListTopToolbar.scala
+++ b/app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListTopToolbar.scala
@@ -21,7 +21,7 @@ import android.content.Context
 import android.util.AttributeSet
 import android.view.View
 import android.view.View.OnClickListener
-import android.widget.{FrameLayout, ImageView}
+import android.widget.{FrameLayout, ImageButton, ImageView}
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
 import com.waz.model.{Availability, UserData}
 import com.waz.service.teams.TeamsService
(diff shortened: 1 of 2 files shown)


app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListTopToolbar.scala
```diff
@@ -118,6 +123,14 @@ class NormalTopToolbar(override val context: Context, override val attrs: Attrib
       drawable.setInfo(NameParts.maybeInitial(user.name).getOrElse(""), TeamIconDrawable.UserShape, selected = false)
   }
 
+  (for {
+    selfUser        <- usersController.selfUser
+    legalHoldActive <- inject[LegalHoldController].isLegalHoldActive(selfUser.id)
+  } yield (legalHoldActive)).onUi({ isActive =>
+    if (isActive) legalHoldIndicatorButton.setVisibility(View.VISIBLE)
+    else legalHoldIndicatorButton.setVisibility(View.INVISIBLE)
+  })
```
makingthematrix, line 132: You can import `ContextUtils`:
```
import com.waz.zclient.utils.ContextUtils._
```
and then simplify the last lines here to:
```
} yield legalHoldActive).onUi(legalHoldIndicatorButton.setVisible)
```

By the way, maybe it would be useful to create a method in `LegalHoldController` to ask for the legal hold of the self user. (Do we ever ask for the elegal hold of another user?)

makingthematrix (changes requested):

app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListTopToolbar.scala
gizemb: `setVisible` makes the view `GONE`, bu here I need it to be `INVISIBLE`. When it's invisible it keeps occupying space, which helps username stay centered on the screen.

For other users, I think we'll need to check legal hold for User Search page, but I'm not sure 🤔 I'd prefer not to introduce this logic into `LegalHoldController` for now, maybe if self user's state is frequently needed I can add it later.

makingthematrix (approved):

PR #3217, opened 2021-03-30
Define legal hold request

johnxnguyen:
### Issues

JIRA: [SQSERVICES-348](https://wearezeta.atlassian.net/browse/SQSERVICES-348)

The legal hold request (that will be eventually received from the backend) should be stored locally so it can be processed by the self user at an appropriate time. 

### Solutions

Define request model. Instances of the legal hold request are intended to be stored in the properties table, so we also add encoding and decoding functionality.

### Testing

Added tests for encoding to and decoding from JSON objects.

#### APK
[Download build #3242](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3242/artifact/build/artifact/wire-dev-PR3217-3242.apk)

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/LegalHoldRequest.scala
+++ b/zmessaging/src/main/scala/com/waz/model/LegalHoldRequest.scala
@@ -17,41 +17,37 @@
  */
 package com.waz.model
 
-import com.waz.model.LegalHoldRequest.Client
-import com.waz.model.otr.PreKeyEncoder
+import com.waz.model.otr.{ClientId, PreKeyEncoder}
 import com.waz.utils.{JsonDecoder, JsonEncoder}
 import com.wire.cryptobox.PreKey
 import org.json.JSONObject
 
-final case class LegalHoldRequest(client: Client, lastPreKey: PreKey)
+final case class LegalHoldRequest(clientId: ClientId, lastPreKey: PreKey)
 
 object LegalHoldRequest {
-
-  final case class Client(id: String)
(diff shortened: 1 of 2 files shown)


zmessaging/src/main/scala/com/waz/model/LegalHoldRequest.scala
```diff
@@ -0,0 +1,62 @@
...
+ */
+package com.waz.model
+
+import com.waz.model.LegalHoldRequest.{Client, Prekey}
+import com.waz.utils.{JsonDecoder, JsonEncoder}
+import org.json.JSONObject
+
+final case class LegalHoldRequest(client: Client, lastPrekey: Prekey)
+
+object LegalHoldRequest {
```
gizemb, line 26: I guess you can use already existing data types (`ClientId` and `PreKey`) and make something very similar to `ClientDecoder` in `OtrClient.scala`. I don't know if `PreKey` type makes it hard to store it in db, but seems like you can at least use `ClientId` and its encoder/decoder instead of creating a new `Client` type here. we also have `PreKeyDecoder`/`PreKeyEncoder`

gizemb (changes requested):

zmessaging/src/main/scala/com/waz/model/LegalHoldRequest.scala
johnxnguyen: Done

gizemb (approved):

PR #3221, opened 2021-03-31
Removing icu4j-shrunk and fixing tests

makingthematrix:
We use a custom library, icu4j-shrunk, for transliteration. Since Android 7 Android has its own transliteration (https://developer.android.com/guide/topics/resources/internationalization) so I decided to remove it instead of trying to migrate to Sonatype. The removal required me to rewrite some unit tests.

Please test if this works well when you write in non-Latin scripts. It seems fine with Polish characters ąęćźżółś, but it's not exactly the same thing :) 

By the way, in a few places in UserSearchService I changed the rules of filtering to make it disregard the lowercase/uppercase distinction. That's because after I updated the unit tests, a few of them started to fail on that. The search should disregard the letter case, so the changed shouldn't break anything. In fact, I suspect this was a bug in the search functionality.


build.gradle
```diff
@@ -58,6 +58,7 @@ allprojects {
     repositories {
         google()
         mavenCentral()
+        maven { url "https://oss.sonatype.org/content/repositories/snapshots" }
```
mejdoo, line 61: do you still need this?

mejdoo: @makingthematrix

makingthematrix: Removed.

mejdoo (approved):

makingthematrix:
build again

PR #3224, opened 2021-03-31
Feature: Legal Hold - Create a general purpose confirmation dialog with password input

gizemb:
## What's new in this PR?

Jira: [SQSERVICES-351](https://wearezeta.atlassian.net/browse/SQSERVICES-351)

### Issues

Given that I am a team user
When a team admin enables legal hold for me
Then I should be informed with a pop up

### Solutions

We already have a similar layout with the same UX, called `RemoveDeviceDialog`, only with different texts to be displayed. I extracted main parts to a base class called `ConfirmationWithPasswordDialog` and had `RemoveDeviceDialog` extend it. The base class will later be extended again for legal hold acceptance pop up. UI or logic-wise, nothing has changed.

### Testing

Manually tested by deleting a device from devices list.

#### APK
[Download build #3252](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3252/artifact/build/artifact/wire-dev-PR3224-3252.apk)
[Download build #3261](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3261/artifact/build/artifact/wire-dev-PR3224-3261.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/preferences/dialogs/ConfirmationWithPasswordDialog.scala
+++ b/app/src/main/scala/com/waz/zclient/preferences/dialogs/ConfirmationWithPasswordDialog.scala
@@ -22,14 +22,14 @@ abstract class ConfirmationWithPasswordDialog extends DialogFragment with Fragme
 
   val onAccept = EventStream[Option[Password]]()
 
-  private lazy val root = LayoutInflater.from(getActivity).inflate(R.layout.remove_otr_device_dialog, null)
+  private lazy val root = LayoutInflater.from(getActivity).inflate(R.layout.confirmation_with_password_dialog, null)
 
   private def providePassword(password: Option[Password]): Unit = {
     onAccept ! password
     dismiss() // if the password is wrong a new dialog will appear
   }
 
-  private lazy val passwordEditText = returning(findById[EditText](root, R.id.acet__remove_otr__password)) { v =>
+  private lazy val passwordEditText = returning(findById[EditText](root, R.id.confirmation_with_password_edit_text)) { v =>
     v.setOnEditorActionListener(new TextView.OnEditorActionListener() {
       def onEditorAction(v: TextView, actionId: Int, event: KeyEvent) =
         actionId match {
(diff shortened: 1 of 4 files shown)


app/src/main/scala/com/waz/zclient/preferences/dialogs/ConfirmationWithPasswordDialog.scala
```diff
@@ -0,0 +1,88 @@
...
+  val onAccept = EventStream[Option[Password]]()
+
+  private lazy val root = LayoutInflater.from(getActivity).inflate(R.layout.confirmation_with_password_dialog, null)
+
+  private def providePassword(password: Option[Password]): Unit = {
+    onAccept ! password
+    dismiss() // if the password is wrong a new dialog will appear
+  }
+
+  private lazy val passwordEditText = returning(findById[EditText](root, R.id.confirmation_with_password_edit_text)) { v =>
```
mejdoo, line 32: Just asking...did we agree to avoid lazy loading for views? or we wanna keep doing this?

app/src/main/scala/com/waz/zclient/preferences/dialogs/RemoveDeviceDialog.scala
```diff
@@ -17,87 +17,30 @@
...
+  override lazy val isSSO: Boolean = getArguments.getBoolean(IsSSOARG)
 
-  private lazy val root = LayoutInflater.from(getActivity).inflate(R.layout.remove_otr_device_dialog, null)
+  override def errorMessage: Option[String] = Option(getArguments.getString(ErrorArg))
 
-  private def providePassword(password: Option[Password]): Unit = {
-    onDelete ! password
-    dismiss() // if the password is wrong a new dialog will appear
-  }
+  override def title: String = getString(
```
makingthematrix, line 31: Having all those fields as `def` in the abstract class is a good practice, because `def` is the most generic term, but in the implementing class, if the value of the field is known from the beginning, please switch it to `override val` or `override lazy val`. It helps with reading the code and can be even better for performance.

makingthematrix (changes requested):

app/src/main/scala/com/waz/zclient/preferences/dialogs/ConfirmationWithPasswordDialog.scala
makingthematrix: Please use `lazy val` as default. Change to `def` only if necessary.

gizemb: generally we should avoid lazy vals in android views, because when the framework tries to recreate the view, some operations won't be called again due to "lazy" nature. however, for this specific screen I couldn't come up with a scenario as I described above because when this pop-up needs to be created, often the fragment that opens it needs to be created as well, which triggers a brand new pop-up rather than re-triggering the lifecycle methods.

this class is a direct copy paste of RemoveDeviceDialog, only stripped from feature specific texts, and we didn't have any problems with Remove Device pop up so far. I guess for now it's safe to keep the code as it is.

app/src/main/scala/com/waz/zclient/preferences/dialogs/RemoveDeviceDialog.scala
gizemb: since all the texts are static here, I converted them to `lazy val`

makingthematrix (approved):

PR #3227, opened 2021-03-31
Fix: Empty layout on ParticipantFragment after opening "Create group" and coming back

gizemb:
### Issues

Jira: [SQCORE-582](https://wearezeta.atlassian.net/browse/SQCORE-582)

1. Go to a 1-to-1 conversation and open details.
2. You should see participant's info screen with two tabs: details and devices.
3. Click on the "Create group" button.
4. Click the back arrow icon.

Result: Participant info screen (`SingleParticipantFragment`) is visible but it has no content.

### Causes

When we come back to `SingleParticipantFragment`, its `onViewCreated` is called again to re-initialize the views. However, since the views are declared as `lazy val`, they aren't initialized again.

### Solutions

Don't store the views as `lazy`, create methods for initialization instead. I also did the same in all fragments extending from `SingleParticipantFragment`.

### Testing

Manually tested with the steps mentioned above.

#### APK
[Download build #3257](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3257/artifact/build/artifact/wire-dev-PR3227-3257.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/participants/fragments/ConnectRequestFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/participants/fragments/ConnectRequestFragment.scala
@@ -24,7 +24,7 @@ class ConnectRequestFragment extends UntabbedRequestFragment {
 
   override protected val Tag: String = ConnectRequestFragment.Tag
 
-  private lazy val ignoreButton = returning(view[ZetaButton](R.id.zb__connect_request__ignore_button)) { vh =>
+  private def initIgnoreButton(): Unit = returning(view[ZetaButton](R.id.zb__connect_request__ignore_button)) { vh =>
     vh.foreach { button =>
       button.setIsFilled(false)
       button.onClick(usersController.ignoreConnectionRequest(userToConnectId).map(_ => getActivity.onBackPressed()))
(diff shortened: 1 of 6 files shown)


app/src/main/scala/com/waz/zclient/participants/fragments/ConnectRequestFragment.scala
```diff
@@ -69,10 +69,10 @@ class ConnectRequestFragment extends UntabbedRequestFragment {
   }
 
   override protected def initViews(savedInstanceState: Bundle): Unit = {
-    detailsView
-    footerMenu
-    ignoreButton
-    acceptButton
+    initDetailsView()
```
mejdoo, line 72: I wonder why we are using too much lazy loading in the Scala codebase as it's making the buggy in such scenarios. I remember I had the same issue in one of my PRs. @makingthematrix @gizemb thoughts??

mejdoo (approved):

app/src/main/scala/com/waz/zclient/participants/fragments/ConnectRequestFragment.scala
makingthematrix: @mejdoo : `lazy val` is usually a good solution. It makes logic simpler and sometimes helps with performance. The situations like this one are corner cases.

makingthematrix (approved):

app/src/main/scala/com/waz/zclient/participants/fragments/ConnectRequestFragment.scala
gizemb: @makingthematrix @mejdoo I wouldn't say this is a corner case. when it comes to Android components, we should be aware of what lifecycle means, and use the language accordingly. this bug clearly indicates that we shouldn't be using `lazy` for view references in `Fragments`. I think it's safe to say we should avoid `lazy` for views everywhere in the app, and make use of Android framework methods. In other places where we don't open a new fragment on top of another, maybe we don't notice it yet but once we add another fragment, this bug will happen there too. 

lazy can be a good solution in some other scenarios, but not for views. also, I know that in Kotlin, `lazy` creates another object under the hood, to track whether the field is initialized. If that's also the case for Scala, excessive usage of `lazy` might be bad for memory.

PR #3233, opened 2021-04-07
fix: App freeze when the backup is being restored on a new device

makingthematrix:
fixes https://wearezeta.atlassian.net/browse/SQCORE-498

It turns out that asking for a zms instance right after backup restoration but before the end of the initialization process does not work if the app was just installed or if the storage was wiped out. The zms does not exist yet at this point.
(Which is worrying because I remember testing this case a few months ago and it worked back then).

I changed the way of accessing zms to one that shouldn't cause trouble - from the account manager we have just created and we know it exists. I also decided to postpone the sync with backend and to sync only conversations instead of all data. Syncing all data during the backup restoration might actually look like a freeze from the user's point of view, esp. if the internet connection is poor. It's better to do it in the background when the conversation list is loading up. (And then instead of a turning circle the user will see the bar at the top).

#### APK
[Download build #3279](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3279/artifact/build/artifact/wire-dev-PR3233-3279.apk)
[Download build #3293](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3293/artifact/build/artifact/wire-dev-PR3233-3293.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/Intents.scala
+++ b/app/src/main/scala/com/waz/zclient/Intents.scala
@@ -34,6 +34,7 @@ object Intents {
   private val ConvIdExtra           = "conv_id"
 
   private val OpenPageExtra         = "open_page"
+  private val InitSyncExtra         = "init_sync"
 
   type Page = String
   object Page {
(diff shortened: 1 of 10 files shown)


mejdoo (approved):

makingthematrix:
This is not working. Please don't merge it.

mejdoo:
@makingthematrix maybe close it?

makingthematrix:
> @makingthematrix maybe close it?

I think this change is good too. It's a more lightweight way of starting the sync with the backend. I'd like to keep it.
The real reason for the bug is not in the backup restoration, though, but in that after the restoration we need to update the client. There's something wrong with that and it looks like it can happen also in situations when we don't restore a backup.

PR #3235, opened 2021-04-08
fix: Another try at fixing the notification channel crash from Google Play Console

makingthematrix:
I looked through the code of the commit which most probably causes the crashes and through Stack Overflow entries on similar issues. I found this: https://stackoverflow.com/a/51378281/2975925 which says that we shouldn't use localized strings as the notification channel ids, and I followed this advise in WebSocketService, and also I decided to bring back a line that ensures the notification channel is created when the app goes to foreground (just as it was before).

We are still not able to reproduce the crash 100% but I think this PR should help.

#### APK
[Download build #3284](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3284/artifact/build/artifact/wire-dev-PR3235-3284.apk)
[Download build #3286](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3286/artifact/build/artifact/wire-dev-PR3235-3286.apk)

### DIFF
--- a/app/src/main/scala/com/waz/services/websocket/WebSocketService.scala
+++ b/app/src/main/scala/com/waz/services/websocket/WebSocketService.scala
@@ -199,8 +199,8 @@ class WebSocketService extends ServiceHelper with DerivedLogTag {
     notificationChannel
   }
 
-  // this is the same code as in WireApplication.notificationChannel but I don't want to access it
-  // from here and the other way around as well (Maciek)
+  // this is the same code as in WireApplication.notificationChannel
+  // but we don't want to access it from here and the other way around as well
   lazy val notificationChannel: Option[NotificationChannel] =
     if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O)
       Some(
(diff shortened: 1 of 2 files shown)


app/src/main/scala/com/waz/services/websocket/WebSocketService.scala
```diff
@@ -195,20 +196,28 @@ class WebSocketService extends ServiceHelper with DerivedLogTag {
...
+    notificationChannel
   }
 
-  private lazy val createNotificationChannel: Unit =
-    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O) {
-      val notificationManager = inject[NotificationManager]
-      notificationManager.createNotificationChannel(
-        returning(new NotificationChannel(ForegroundNotificationChannelId, getString(R.string.foreground_service_notification_name), NotificationManager.IMPORTANCE_LOW)) { ch =>
+  // this is the same code as in WireApplication.notificationChannel but I don't want to access it
+  // from here and the other way around as well (Maciek)
```
mejdoo, line 203: (Maciek)?

makingthematrix: yes, I wrote a comment there and left my name so if anyone wanted to do something about it, they can contact me 🙂

makingthematrix:
build again

app/src/main/scala/com/waz/services/websocket/WebSocketService.scala
mejdoo: we usually don't leave our names in the comments :)

mejdoo (approved):

PR #3241, opened 2021-04-12
Bugfix : Hide call controls overlay after zooming

mejdoo:
### Issues

The call controls overlay is always appearing after zooming. It's very annoying from UX prospective (It's the main reason why the zooming feature is not released yet).  I couldn't find a solution to fix this without forking the zooming library and doing some refactoring to add a missing callback `ZoomLayoutGestureListener.onSingleTap()`. 

https://wearezeta.atlassian.net/browse/SQCALL-203

#### APK
[Download build #3303](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3303/artifact/build/artifact/wire-dev-PR3241-3303.apk)


makingthematrix (changes requested):
Please either include this library as a dependency or rewrite the code in Scala and add a header that it is based on the original Java code. Let's not include new Java code into the codebase.

mejdoo:
> Please either include this library as a dependency or rewrite the code in Scala and add a header that it is based on the original Java code. Let's not include new Java code into the codebase.

@makingthematrix done, please review again.

buildSrc/src/main/kotlin/Dependencies.kt
```diff
@@ -152,7 +152,7 @@ object BuildDependencies {
     val pinEditText = "com.poovam:pin-edittext-field:${Versions.PIN_EDITTEXT}"
     val libSodium = "com.github.joshjdevl.libsodiumjni:libsodium-jni-aar:${Versions.LIB_SODIUM}"
     val countly = "ly.count.android:sdk:${Versions.COUNTLY}"
-    val zooming = "com.xlw.zoom:zoomlayout:${Versions.ZOOMING}"
+    val zooming = "com.wire.zoom:zoomlayout:${Versions.ZOOMING}"
```
gizemb, line 155: it should be **com.wire**:zoomlayout: (at least that's what I intended to do :D)

mejdoo: I'm not sure exactly why it's **com.wire.zoom**, maybe because @Gardosen published it with the script that he did not with the gradle task.

makingthematrix: Yes, it needs to be comwire:zoomlayout for Sonatype. All our libraries there are under com.wire

makingthematrix (approved):

PR #3264, opened 2021-04-22
Qualified ids

makingthematrix:
Changes to UserInfo and UserData which ensure that Wire Android will use qualified ids where they are available, but also will work without them. A qualified id is a pair of user id and domain (a string). The backend should return it as a part of each UserInfo, but it will also return id as a separate field in the json.  In case when qualified_id exists, we give it priority, but anyway both should be the same.
    
Additionally, I changed ClientId to a newtype, in accordance to changes of other ids from a recent PR.
#### APK
[Download build #3367](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3367/artifact/build/artifact/wire-dev-PR3264-3367.apk)
[Download build #3381](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3381/artifact/build/artifact/wire-dev-PR3264-3381.apk)
[Download build #3384](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3384/artifact/build/artifact/wire-dev-PR3264-3384.apk)

### DIFF
--- a/storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase130To131Migration.kt
+++ b/storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase130To131Migration.kt
@@ -5,12 +5,15 @@ import androidx.room.migration.Migration
 import androidx.sqlite.db.SupportSQLiteDatabase
 import com.waz.zclient.storage.db.users.migration.MigrationUtils.addColumn
 
+private const val USER_TABLE_NAME = "Users"
+private const val DOMAIN_COLUMN_NAME = "domain"
+
 val USER_DATABASE_MIGRATION_130_TO_131 = object : Migration(130, 131) {
     override fun migrate(database: SupportSQLiteDatabase) {
         addColumn(
             database = database,
-            tableName = "Users",
-            columnName = "domain",
+            tableName = USER_TABLE_NAME,
+            columnName = DOMAIN_COLUMN_NAME,
             columnType = MigrationUtils.ColumnType.TEXT
         )
     }
--- a/storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase130To131Migration.kt
+++ b/storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase130To131Migration.kt
@@ -1,0 +1,18 @@
+@file:Suppress("MagicNumber")
+package com.waz.zclient.storage.db.users.migration
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+import com.waz.zclient.storage.db.users.migration.MigrationUtils.addColumn
+
+val USER_DATABASE_MIGRATION_130_TO_131 = object : Migration(130, 131) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        addColumn(
+            database = database,
+            tableName = "Users",
+            columnName = "domain",
+            columnType = MigrationUtils.ColumnType.TEXT
+        )
+    }
+}
+
--- a/storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase130To131Migration.kt
+++ b/storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase130To131Migration.kt
@@ -15,4 +15,3 @@ val USER_DATABASE_MIGRATION_130_TO_131 = object : Migration(130, 131) {
         )
     }
 }
-
--- a/zmessaging/src/main/scala/com/waz/content/Preferences.scala
+++ b/zmessaging/src/main/scala/com/waz/content/Preferences.scala
@@ -385,7 +385,7 @@ object UserPreferences {
 
   //increment number to perform slow sync on particular type
   lazy val ShouldSyncConversations = PrefKey[Boolean]("should_sync_conversations_2", customDefault = true)
-  lazy val ShouldSyncInitial = PrefKey[Boolean]("should_sync_initial_2", customDefault = true)
+  lazy val ShouldSyncInitial = PrefKey[Boolean]("should_sync_initial_3", customDefault = true)
   lazy val ShouldSyncUsers = PrefKey[Boolean]("should_sync_users_1", customDefault = true)
   lazy val ShouldSyncTeam = PrefKey[Boolean]("should_sync_team_1", customDefault = true)
   lazy val ShouldSyncFolders = PrefKey[Boolean]("should_sync_folders", customDefault = true)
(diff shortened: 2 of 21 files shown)


storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase130To131Migration.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import androidx.room.migration.Migration
+import androidx.sqlite.db.SupportSQLiteDatabase
+import com.waz.zclient.storage.db.users.migration.MigrationUtils.addColumn
+
+val USER_DATABASE_MIGRATION_130_TO_131 = object : Migration(130, 131) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        addColumn(
+            database = database,
+            tableName = "Users",
```
mejdoo, line 12: You can declare `"Users`" and `"domain"`as constant.

zmessaging/src/main/scala/com/waz/content/Preferences.scala
```diff
@@ -385,7 +385,7 @@ object UserPreferences {
 
   //increment number to perform slow sync on particular type
   lazy val ShouldSyncConversations = PrefKey[Boolean]("should_sync_conversations_2", customDefault = true)
-  lazy val ShouldSyncInitial = PrefKey[Boolean]("should_sync_initial_2", customDefault = true)
+  lazy val ShouldSyncInitial = PrefKey[Boolean]("should_sync_initial_3", customDefault = true)
```
mejdoo, line 388: what is `should_sync_initial_2` and `should_sync_initial_3`?

zmessaging/src/main/scala/com/waz/model/QualifiedId.scala
```diff
@@ -0,0 +1,21 @@
...
+
+import com.waz.utils.JsonDecoder.opt
+import com.waz.utils.{JsonDecoder, JsonEncoder}
+import org.json.JSONObject
+
+final case class QualifiedId(id: UserId, domain: String)
+
+object QualifiedId {
+  implicit val Encoder: JsonEncoder[QualifiedId] = JsonEncoder.build(qId => js => {
+    js.put("id", qId.id.str)
```
mejdoo, line 11: `"id"` and `"domain"`can be constants

zmessaging/src/main/scala/com/waz/model/QualifiedId.scala
```diff
@@ -0,0 +1,21 @@
...
+    js.put("domain", qId.domain)
+  })
+
+  private def decode(js: JSONObject): QualifiedId =
+    QualifiedId(UserId(js.getString("id")), js.getString("domain"))
```
mejdoo, line 16: same here

zmessaging/src/main/scala/com/waz/model/UserData.scala
```diff
@@ -225,13 +230,13 @@ object UserData {
 
     override val idCol = Id
     override val table = Table(
-      "Users", Id, TeamId, Name, Email, Phone, TrackingId, Picture, Accent, SKey, Conn, ConnTime, ConnMessage,
+      "Users", Id, Domain, TeamId, Name, Email, Phone, TrackingId, Picture, Accent, SKey, Conn, ConnTime, ConnMessage,
```
mejdoo, line 233: same here for `"Users"`

zmessaging/src/main/scala/com/waz/model/UserData.scala
```diff
@@ -34,32 +34,33 @@ import com.waz.utils.wrappers.{DB, DBCursor}
...
-                    availability:          Availability           = Availability.None,
-                    handle:                Option[Handle]         = None,
-                    providerId:            Option[ProviderId]     = None,
-                    integrationId:         Option[IntegrationId]  = None,
-                    expiresAt:             Option[RemoteInstant]  = None,
-                    managedBy:             Option[ManagedBy]      = None,
-                    fields:                Seq[UserField]         = Seq.empty,
-                    permissions:           PermissionsMasks       = (0,0),
-                    createdBy:             Option[UserId]         = None) extends Identifiable[UserId] {
+final case class UserData(override val id:       UserId,
```
mejdoo, line 37: why you need to declare it as `final`?

zmessaging/src/main/scala/com/waz/model/UserInfo.scala
```diff
@@ -29,29 +29,30 @@ import org.json.{JSONArray, JSONObject}
...
-                    deleted:      Boolean                 = false,
-                    handle:       Option[Handle]          = None,
-                    privateMode:  Option[Boolean]         = None,
-                    service:      Option[Service]         = None,
-                    teamId:       Option[TeamId]          = None,
-                    expiresAt:    Option[RemoteInstant]   = None,
-                    ssoId:        Option[SSOId]           = None,
-                    managedBy:    Option[ManagedBy]       = None,
-                    fields:       Option[Seq[UserField]]  = None)
+final case class UserInfo(id:           UserId,
```
mejdoo, line 32: same question regarding `final`

mejdoo (changes requested):

zmessaging/src/main/scala/com/waz/content/Preferences.scala
makingthematrix: @mejdoo : This is to trigger the full sync on update to the new Wire version. The user will already have `should_sync_initial_2` set to `false` in their user preferences. We want to perform the full sync again, but we can't just switch that user preference to `false` or app initialization or startup, because then it will trigger the full sync every time the app is opened, and we want to do it only once, the first time the user opens the app after the update. So we switch to a new user preference `should_sync_initial_2` which is initially set to `true`. The app will perform the full sync on init, and then set it to false.

zmessaging/src/main/scala/com/waz/model/QualifiedId.scala
makingthematrix: ok

zmessaging/src/main/scala/com/waz/model/QualifiedId.scala
makingthematrix: ok

storage/src/main/kotlin/com/waz/zclient/storage/db/users/migration/UserDatabase130To131Migration.kt
makingthematrix: ok

zmessaging/src/main/scala/com/waz/model/UserData.scala
makingthematrix: ok

zmessaging/src/main/scala/com/waz/content/Preferences.scala
mejdoo: ok maybe rename it to something which makes more sense `should_sync_on_update`

mejdoo (approved):
I left a suggestion about the name of the SharedPreference.

PR #3272, opened 2021-04-27
[Feature] Process enable and disable legal hold events

johnxnguyen:
## What's new in this PR?

Process two remaining legal hold events, in both cases delete an existing legal hold request.

### Testing

Added tests to assert:
- event decoding, 
- the legal hold request is deleted when processing both enable and disable events,
- the legal hold request is not deleted when processing the events not associated with the self user

#### APK
[Download build #3391](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3391/artifact/build/artifact/wire-dev-PR3272-3391.apk)
[Download build #3412](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3412/artifact/build/artifact/wire-dev-PR3272-3412.apk)
[Download build #3442](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3442/artifact/build/artifact/wire-dev-PR3272-3442.apk)


zmessaging/src/main/scala/com/waz/service/LegalHoldService.scala
```diff
@@ -39,12 +39,22 @@ class LegalHoldServiceImpl(selfUserId: UserId,
...
+
+  private def processEvent(event: LegalHoldEvent): Future[Unit] = event match {
+    case LegalHoldRequestEvent(userId, request) if userId == selfUserId =>
+      storeLegalHoldRequest(request)
+
+    case LegalHoldEnableEvent(userId) if userId == selfUserId =>
+      deleteLegalHoldRequest()
+
+    case LegalHoldDisableEvent(userId) if userId == selfUserId =>
+      deleteLegalHoldRequest()
```
gizemb, line 54: for a request to be disabled, first it should be enabled, hence, there will be no request to delete actually, right?

gizemb (approved):

zmessaging/src/main/scala/com/waz/service/LegalHoldService.scala
```diff
@@ -39,12 +39,22 @@ class LegalHoldServiceImpl(selfUserId: UserId,
...
+  def legalHoldEventStage: Stage.Atomic = EventScheduler.Stage[LegalHoldEvent] { (_, events) =>
+    Future.traverse(events)(processEvent)
+  }
+
+  private def processEvent(event: LegalHoldEvent): Future[Unit] = event match {
+    case LegalHoldRequestEvent(userId, request) if userId == selfUserId =>
+      storeLegalHoldRequest(request)
+
+    case LegalHoldEnableEvent(userId) if userId == selfUserId =>
+      deleteLegalHoldRequest()
```
gizemb, line 51: should we remove deleteLegalHoldRequest() call on line 81 now?

zmessaging/src/main/scala/com/waz/service/LegalHoldService.scala
johnxnguyen: Not necessarily, consider the case:

1. An admin gives you a legal hold request,
2. The admin cancels the request before you approved it.

The user goes from a disabled state, to pending, and back to disabled.

zmessaging/src/main/scala/com/waz/service/LegalHoldService.scala
johnxnguyen: Indeed this line should be enough and we could skip deleting on line 81, but I don't see the harm in keeping both. What do you think?

gizemb: I don't see any harm either, let's keep it

zmessaging/src/main/scala/com/waz/service/LegalHoldService.scala
gizemb: ok, good point :)

PR #3279, opened 2021-04-29
[Feature] Add generic message legal hold status flag

johnxnguyen:
## What's new in this PR?

This PR adds support for getting and setting legal hold status flag on generic messages. The purpose of the flag is to give hints to recipients of the message whether they should verify the local legal hold status of the conversation, i.e fetch the list of clients from the backend to discover a legal hold device.

The intended use of the flag is to read it when processing messages, and to set it when sending messages.

I chose to implement it as a method on `GenericMessage` container rather than the constructors of the `GenericContent` because the value of the flag isn't necessarily known at the time the content is created, but rather at a later point such as when posting the message to the backend.

### Testing

Added unit tests for setting and getting the flag on both ephemeral and non-ephemeral messages. The distinction is important because when setting the flag on an existing message, we must take care to preserve the ephemerality of content.

#### APK
[Download build #3413](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3413/artifact/build/artifact/wire-dev-PR3279-3413.apk)
[Download build #3414](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3414/artifact/build/artifact/wire-dev-PR3279-3414.apk)
[Download build #3415](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3415/artifact/build/artifact/wire-dev-PR3279-3415.apk)
[Download build #3416](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3416/artifact/build/artifact/wire-dev-PR3279-3416.apk)
[Download build #3440](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3440/artifact/build/artifact/wire-dev-PR3279-3440.apk)
[Download build #3443](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3443/artifact/build/artifact/wire-dev-PR3279-3443.apk)


makingthematrix (approved):
This is a bit unusual way of using GenericMessage but I understand the reason.
Maybe let's come back to it when the whole feature is ready and then discuss if this needs refactoring.

johnxnguyen:
> This is a bit unusual way of using GenericMessage but I understand the reason.

I'm basing the implementation from what we have in the iOS client, which is possibly the reason. Anyway, could you elaborate why it is unusual and what maybe also outline how you would imagine to refactor it?

PR #3287, opened 2021-04-30
refactor: Changes required to work with wire-signals-0.4.2

makingthematrix:
The new version of Wire Signals brings some changes in API and Wire Android code needed some modifications.
The most important changes are:
1. I removed subscriptions to event streams and signals with the `apply` method. Until now you could have seen in the code things like `signal { _ => ... }` which was a shorthand to `signal.apply(...)` and that in turn was an alternative to `signal.foreach { _ => ... }` or `signal.onUi { _ => ... }`. It was a bit confusing because it looked as if `signal` was a method, not an object, and on top of that, sometimes it was possible to make the signal execute on the wrong execution context (in those cases, usually `Ui` was used instead of `Background`, which wasn't a bug, but we shouldn't do too much on the Ui thread if it's not necessary). So in all those places I added `.foreach` or `.onUi`.
2. `CancellableFuture` is not always cancellable. For example, if we wrap a regular future in `CancellableFuture`, the original future still can't be cancelled. Or if the cancellable future is already completed, of course. Until now there was no way to check it (it's still not as clear I would like, but it's better) and in some places in Wire Android I have found code which tried very hard to cancel an uncancellable future. Besides, those uncancellable-cancellable futures should work a bit faster now.
3. The `mapAsync` method in signals and event streams became `mapSync`, because that's a better name for a method which synchronizes mapping of events.

#### APK
[Download build #3433](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3433/artifact/build/artifact/wire-dev-PR3287-3433.apk)
[Download build #3448](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3448/artifact/build/artifact/wire-dev-PR3287-3448.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/common/controllers/ThemeController.scala
+++ b/app/src/main/scala/com/waz/zclient/common/controllers/ThemeController.scala
@@ -37,7 +37,7 @@ import scala.concurrent.duration._
 
 class ThemeController(implicit injector: Injector, context: Context)
   extends Injectable with DerivedLogTag {
-  
+
   import Threading.Implicits.Background
 
   val optionsDarkTheme:  OptionsTheme = new OptionsDarkTheme(context)
@@ -96,17 +96,17 @@ trait ThemedView extends View with ViewHelper {
 
   override def onAttachedToWindow(): Unit = {
     super.onAttachedToWindow()
-    getThemeFromParent(this)(currentTheme ! _)
+    getThemeFromParent(this).foreach(currentTheme ! _)
   }
 
-  private def getThemeFromParent(view: View): Signal[Option[Theme]] = {
+  @scala.annotation.tailrec
+  private def getThemeFromParent(view: View): Signal[Option[Theme]] =
     view.getParent match {
       case v: ThemeControllingView => v.theme
       case v: ThemedView => v.currentTheme
       case v: View => getThemeFromParent(v)
       case _ => Signal.const(None)
     }
-  }
 }
 
 object ThemeController {
--- a/zmessaging/src/main/scala/com/waz/model/errors.scala
+++ b/zmessaging/src/main/scala/com/waz/model/errors.scala
@@ -26,7 +26,7 @@ import scala.concurrent.{ ExecutionContext, Future }
 object errors {
 
   implicit class FutureOps[T](val value: Future[T]) extends AnyVal {
-    def toCancellable: CancellableFuture[T] = CancellableFuture.lift(value)
+    def toUncancellable: CancellableFuture[T] = CancellableFuture.lift(value)
     def modelToEither(implicit ec: ExecutionContext): Future[Either[ZError, T]] =
       value.map(Right(_): Either[ZError, T]).recover { case err => Left(UnexpectedError(err)) }
     def eitherToModel[A](implicit ev: T =:= Either[ZError, A], ec: ExecutionContext): Future[A] =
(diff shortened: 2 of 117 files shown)


app/src/main/scala/com/waz/zclient/common/controllers/ThemeController.scala
```diff
@@ -96,17 +96,17 @@ trait ThemedView extends View with ViewHelper {
 
   override def onAttachedToWindow(): Unit = {
     super.onAttachedToWindow()
-    getThemeFromParent(this)(currentTheme ! _)
+    getThemeFromParent(this).foreach(currentTheme ! _)
   }
 
-  private def getThemeFromParent(view: View): Signal[Option[Theme]] = {
+  @scala.annotation.tailrec
```
gizemb, line 102: what does this annotation do?

zmessaging/src/main/scala/com/waz/model/errors.scala
```diff
@@ -26,7 +26,7 @@ import scala.concurrent.{ ExecutionContext, Future }
 object errors {
 
   implicit class FutureOps[T](val value: Future[T]) extends AnyVal {
-    def toCancellable: CancellableFuture[T] = CancellableFuture.lift(value)
+    def toUncancellable: CancellableFuture[T] = CancellableFuture.lift(value)
```
gizemb, line 29: isn't it a bit confusing to return `CancellableFuture` from a function called `toUncancellable`? maybe we could return Future[T] instead of CancellableFuture[T] to avoid confusion

app/src/main/scala/com/waz/zclient/common/controllers/ThemeController.scala
makingthematrix: This method is recursive, but the recursive call is in the tailrec position. The annotation is here only to secure it - if someone tries to modify this method and the recursive call stops to be tail-recursive, the compiler will report an error. 

In practice, I don't think choosing the theme is so important that we should watch for tail-recursion, but Android Studio suggested I add this annotation anyway :)

zmessaging/src/main/scala/com/waz/model/errors.scala
makingthematrix: @gizemb : I know the naming is confusing here. Maybe I could change it from `.toUncancellable` simply to `.lift`? The reason for this method is that sometimes we have a `Future[T]` but we need to return a `CancellableFuture[T]`. So we need to wrap this `Future[T]` - or in other words "lift" it to the `CancellableFuture[T]` status. But it's still an already running `Future[T]` which can't be cancelled, no matter if we wrap it.

It was always like this, only now it's more visible.

gizemb: is `lift` a common name in Scala community (like apply etc?) then it could be `lift`. or it could be something like `toPseudoCancellable`

app/src/main/scala/com/waz/zclient/common/controllers/ThemeController.scala
gizemb: ok thanks :)

zmessaging/src/main/scala/com/waz/model/errors.scala
makingthematrix: @gizemb : I think `lift` is reasonably popular.

gizemb: ok then, let's go with `lift`

gizemb (approved):

makingthematrix:
build again

PR #3299, opened 2021-05-06
Feature: Legal Hold - Hide Legal Hold device from device list when device limit is reached

gizemb:
### Issues

[SQSERVICES-344](https://wearezeta.atlassian.net/browse/SQSERVICES-344)

When the user logs in for the first time, if device limit is reached, we show a list of user's devices so that user can pick one and remove by simply tapping on device in the list. However, the user should not be able to remove Legal Hold device himself/herself. So, we need to hide Legal Hold device if the screen is opened specially for device removal.

### Solutions

Passed a Boolean parameter to controller indicating whether the screen is opened for device removal. Controller filters the Legal Hold device if the screen is opened for removal.

### Testing

Manually tested. Observed that LH device is not visible on "device management" screen. Observed that LH device is visible when same screen is opened after login via Settings menu.


gizemb:
build again

PR #3302, opened 2021-05-07
fix: Update clients in batches

makingthematrix:
The yesterday fix for sending a first message in a conversation made it harder (or maybe even impossible) to handle large conversations. This is an attempt to merge both these problems. The clients are updated in batches (so, a fix to large conversations) and they are properly synced in new conversations.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/sync/otr/OtrClientsSyncHandler.scala
+++ b/zmessaging/src/main/scala/com/waz/sync/otr/OtrClientsSyncHandler.scala
@@ -56,16 +56,14 @@ class OtrClientsSyncHandlerImpl(context:    Context,
                                 storage:    OtrClientsStorage,
                                 cryptoBox:  CryptoBoxService)
   extends OtrClientsSyncHandler
-    with DerivedLogTag {
+    with DerivedLogTag { self =>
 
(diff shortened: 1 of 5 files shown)


zmessaging/src/main/scala/com/waz/sync/otr/OtrClientsSyncHandler.scala
```diff
@@ -56,74 +56,70 @@ class OtrClientsSyncHandlerImpl(context:    Context,
...
+        case (id, clients) => id -> clients.map(c => if (selfClient == c.id) c.copy(verified = Verification.VERIFIED) else c)
+      }
 
     for {
-      ucs <- otrClients.updateUserClients(userId, userClients, replace = true)
-      _   <- syncSessionsIfNeeded(userId, ucs.clients.keys)
-      res <- updatePreKeys(selfClient)
+      _   <- otrClients.updateUserClients(userClients, replace = true)
+      res <- syncSessionsIfNeeded()
+      res <- if (SyncResult.isSuccess(res)) updatePreKeys(selfClient) else Future.successful(res)
```
gizemb, line 109: maybe we could give different names to `res` variables on line 108 and 109. or is this intentional?

gizemb (approved):

PR #3317, opened 2021-05-20
[Feature] Legal hold system messages

johnxnguyen:
## What's new in this PR?

When a legal hold subject is discovered we append system messages into conversations to inform the self user when legal hold is enabled and when it is disabled. The system message appended to **every** conversation in which the legal hold subject shares with the self user.

A special case is when we receive a legal hold hint along with the message. In this case, we insert the system message **before** the associated message in that particular conversation to correctly indicate that the received message is part of legal hold (or vice versa). 

### Testing

Updated legal hold service tests to assert that the system messages are added when: 
- the conversation legal hold status updates,
- a legal hold hint is received

#### APK
[Download build #3503](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3503/artifact/build/artifact/wire-dev-PR3317-3503.apk)


zmessaging/src/main/scala/com/waz/service/LegalHoldService.scala
```diff
@@ -157,11 +160,39 @@ class LegalHoldServiceImpl(selfUserId: UserId,
     convIds    <- convsStorage.contents.map(_.keys)
     legalHolds <- Signal.sequence(convIds.map(convId => legalHold(convId).map(convId -> _)).toSeq: _*)
   } yield legalHolds.toMap).foreach { legalHoldMap =>
-    convsStorage.updateAll2(legalHoldMap.keys, { conv =>
-      val detectedLegalHoldDevice = legalHoldMap(conv.id)
-      if (conv.isUnderLegalHold != detectedLegalHoldDevice) conv.withNewLegalHoldStatus(detectedLegalHoldDevice)
-      else conv
-    })
+    updateConvsAndPostSystemMsgsIfNeeded(legalHoldMap)
```
gizemb, line 163: this is triggered when a we discover a participant's LH device, right?

gizemb (approved):

PR #3321, opened 2021-05-21
Fix: Legal Hold request popup on small screens

gizemb:
### Issues

In small screens, the buttons for the LH request pop up weren't visible because of the keyboard

### Causes

The `AlertDialog` tries to make whole `message` visible.

### Solutions

- Instead of the `AlertDialog`'s built-in `message` property, I added a new `TextView` for the message to our custom view, and wrapped everything inside a `ScrollView` so that the content can shrink and becomes scrollable. I used the `AlertDialog`'s message style, `TextAppearance.AppCompat.Subhead`, for the new TextView in order to keep native look.
- As an extra effort, I deleted excessive blank lines from text message.
- When we close the dialog with either one of the buttons, the keyboard used to stay visible since `MainActivity` couldn't handle that properly. Added a `hideKeyboard` method to prevent this.

### Testing

This pop up shares code with `RemoveDeviceDialog`. However, since it was already a compact dialog, nothing is changed UI-wise. I tested both pop ups in small in big screens.

#### APK
[Download build #3517](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3517/artifact/build/artifact/wire-dev-PR3321-3517.apk)

### DIFF
--- a/app/src/main/java/com/waz/zclient/ui/utils/KeyboardUtils.java
+++ b/app/src/main/java/com/waz/zclient/ui/utils/KeyboardUtils.java
@@ -45,6 +45,10 @@ public class KeyboardUtils {
         }
     }
 
+    public static void hideKeyboard(View view) {
+        getInputMethodManager(view.getContext()).hideSoftInputFromWindow(view.getWindowToken(), 0);
+    }
+
     public static boolean showKeyboard(Activity activity) {
         if (activity != null && activity.getCurrentFocus() != null) {
             getInputMethodManager(activity).showSoftInput(activity.getCurrentFocus(), InputMethodManager.SHOW_IMPLICIT);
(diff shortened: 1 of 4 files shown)


app/src/main/java/com/waz/zclient/ui/utils/KeyboardUtils.java
```diff
@@ -45,6 +45,10 @@ public static void hideKeyboard(Activity activity) {
         }
     }
 
+    public static void hideKeyboard(View view) {
```
gizemb, line 48: somehow `hideKeyboard(activity)` doesn't work for DialogFragments so I had to write this version... classic Android 🙃

mejdoo (approved):

PR #3325, opened 2021-05-24
[Feature] Show system message when policy conflict occurs

johnxnguyen:
## What's new in this PR?

If a legal hold policy conflict occurs, then the backend will remove offending a set of users from the conversation and notify the client app via the `MemberLeaveEvent` with the additional reason `legalhold-policy-conflict`. The client should append a new system message informing the user who was removed and why. Additionally, a link to a FAQ entry is provided to further explain the topic.

This PR adds a new message type `MEMBER_LEAVE_DUE_TO_LEGAL_HOLD` which will be appended only in the case that we parse a member leave event containing the legal hold reason.

### Testing

Manual testing, and added unit tests to parse the update event with and without a reason.

#### APK
[Download build #3527](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3527/artifact/build/artifact/wire-dev-PR3325-3527.apk)


app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListRow.scala
```diff
@@ -418,9 +418,11 @@ object ConversationListRow {
           getString(R.string.conversation_list__added, memberName)
         case Message.Type.MEMBER_JOIN =>
           getString(R.string.conversation_list__added, memberName)
-        case Message.Type. MEMBER_LEAVE if members.exists(_.id == selfId) && user.exists(_.id == selfId) =>
+        case Message.Type.MEMBER_LEAVE |
+             Message.Type.MEMBER_LEAVE_DUE_TO_LEGAL_HOLD if members.exists(_.id == selfId) && user.exists(_.id == selfId) =>
```
gizemb, line 422: this `if` applies to both `MEMBER_LEAVE` and `MEMBER_LEAVE_DUE_TO_LEGAL_HOLD`, right?

app/src/main/scala/com/waz/zclient/messages/parts/MemberChangePartView.scala
```diff
@@ -158,9 +167,15 @@ class MemberChangePartView(context: Context, attrs: AttributeSet, style: Int)
     case Right(d) => messageView.setIcon(d)
   }
 
-  linkText.onUi { text =>
+  linkText.onUi { case (text, msgType) =>
     messageView.setTextWithLink(text, getColor(R.color.accent_blue)) {
-      participantsController.onShowParticipants ! None
+      msgType match {
+        case MEMBER_JOIN =>
```
gizemb, line 173: I think this should be `MEMBER_JOIN | MEMBER_LEAVE` ?

app/src/main/scala/com/waz/zclient/conversationlist/views/ConversationListRow.scala
johnxnguyen: Yes, it is equivalent to 

```scala
case (oneThing | orAnother) if ... => ...
```

app/src/main/scala/com/waz/zclient/messages/parts/MemberChangePartView.scala
johnxnguyen: I checked and it appears we don't ever use the link to go to the participant screen for member leave events.... which makes sense because you wouldn't see the deleted members anyway.

gizemb: hmm, yeap, makes sense  👍

gizemb (approved):

PR #3329, opened 2021-05-25
[Feature] Show alert when failing to add conversation participants

johnxnguyen:
## What's new in this PR?

JIRA: https://wearezeta.atlassian.net/browse/SQSERVICES-467

If the user attempts to add participants to an existing conversation which would lead to a legal hold policy conflict, then the backend will return an error labeled "missing-legalhold-consent". When this happens, we display an alert to the user, as seen below.

### Testing

Manually tested the alert, and added a unit test that the error is reported to the service.

### Attachment

Wire 2021-05-25 at 5 03 PM

#### APK
[Download build #3534](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3534/artifact/build/artifact/wire-dev-PR3329-3534.apk)


app/src/main/res/values/strings.xml
```diff
@@ -1616,5 +1616,10 @@
     <string name="legal_hold_deactivated_dialog_title">Legal Hold Deactivated</string>
     <string name="legal_hold_deactivated_dialog_message">Future messages will not be recorded.</string>
     <string name="legal_hold_activated_dialog_title">Legal Hold is Active</string>
+
+    <string name="legal_hold_participant_missing_consent_alert_title">Guests cannot be added</string>
+    <string name="legal_hold_participant_missing_consent_alert_message">Due to legal hold, only team members can be added to this conversation.</string>
+    <string name="legal_hold_participant_missing_consent_alert_positive_button">OK</string>
```
gizemb, line 1622: we usually use `android.R.string.ok` for OK buttons

app/src/main/scala/com/waz/zclient/participants/fragments/GroupParticipantsFragment.scala
```diff
@@ -183,6 +189,25 @@ class GroupParticipantsFragment extends FragmentHelper {
     participantsAdapter.onBackPressed()
   }
 
+  private def handleSyncError(err: ErrorData): Unit = err.errType match {
+    case ErrorType.CANNOT_ADD_PARTICIPANT_WITH_MISSING_LEGAL_HOLD_CONSENT =>
+      ViewUtils.showAlertDialog(
```
gizemb, line 194: this dialog is cancellable when touched outside. if you want an uncancellable dialog, you can use `ContextUtils.showWarningDialog()`

johnxnguyen: Ok good to know. I think it's ok in this case.

app/src/main/res/values/strings.xml
johnxnguyen: Fixed

PR #3331, opened 2021-05-25
Add the possibility to export conversations as ZIP (XML+mediadata+viewer)

kanashius:
## What's new in this PR?

Add the possibility to export conversations as ZIP-file containing the
conversation as an XML-file.
It is possible to define the desired period of time and whether
media-files, profile pictures and the HTML-Viewer should be included in
the export.
The HTML-Viewer enables the user to view the exported data directly.

This function was developed in connection with Patrick Nolte's
bachelor thesis at the Institute for Networks and Security at
the Johannes Kepler University Linz.

### Issues

The export function was missing.

### Solutions

The option to select and export conversations was added to the settings->account->history as well as
the option to export a single conversation from the conversation_view was added

### Testing

Creation of test groups and chats with every possible message-type that could be sent were exported and verified.
This includes an older chat group with multiple participans and a lot of messages.

## Notes

There is a ZIP-file included with this pull request that can be added to the export and contains an HTML-Viewer and allows the user to view the exported data.
There is also an xml-schema included, that describes the exported XML.


AndroidBob:

makingthematrix:
Hi @kanashius ,

I see that you did a lot of work but unfortunately I dno't think we can accept it and merge it into the main codebase.
We already have the backup feature. All your conversations and other data are serialized to JSON, zipped, and encrypted. You can go through the code and see what is needed do decrypt the file, open it, and then you can display the contents in whatever way you like. So, we don't plan to include a secondary system for exporting data.

Having said that, I'm impressed by the amount of work you did and the quality of code. Maybe you'd be interested into transforming it into a standalone webapp for reviewing Wire backups?

makingthematrix:
@kanashius, @rmayr,

We talked about it in the team and we would like to ask you to turn this feature either into a standalone application (a webapp maybe?) that could be used to view Wire conversation history or you can make a fork of the wire-android repo and build your own client. In both cases, I can help you with that - answer questions, write about how to compile and how to use our code, etc.

What do you think?

makingthematrix:
@rmayr : 
I can ask about the possibility to export backups un-encrypted, but I don't think we can add any other hook - it would be too much work and there would be a risk to threaten security of the application. But in the backup JSON file you have metadata of the assets displayed in conversations and if the reader is able to connect to the Wire backend as if it was another client, it will be able to download those assets. The same process takes places when the backup is restored in the Android client - after restoring the metadata, the app contacts the backend and downloads assets. The problem is, we don't want to keep assets on the backend for too long. The max length of time is four weeks, if I remember correctly. The way I see it, Wire should not be used to store data, only to send and receive it, and then if you need it, you should save it elsewhere. The less data we handle for you, the better it is for privacy and security. 

There is also another reason to go with this solution. We are currently working on a new Wire Android version in Kotlin. The old Scala version is still being developed but we plan to replace it eventually.

makingthematrix:
@rmayr : In that case, I can only suggest you could make a fork of wire-android, cut out almost everything except the logging in, restoring backups, and downloading assets functionalities, add this code, and make a separate app that will open a backup, download the missing assets, and store them in a local database. I know it's a lot, but I don't see a better solution. We are not able to work on adding assets to backups in near future.

PR #3336, opened 2021-05-26
[Fix] Delete conversation only after alert is dismissed

johnxnguyen:
## What's new in this PR?

**JIRA:** https://wearezeta.atlassian.net/browse/SQSERVICES-476

### Issues

When creating a group conversation that contains a legal hold conflict, and after backend rejects the conversation, we will present an alert informing the user of the error. But the conversation is deleted as soon as the error is displayed, so when the error is dismissed the user is left on a deleted conversation.

### Solutions

The conversation deletion happens when dismissing the error via the error controller, which is invoked when we present the alert. Instead, we invoke this when the ok button is tapped. Additionally, the alert is now not cancelable, so the only way to dismiss the alert is with the ok button. Finally, we also pop the conversation view off the stack so the user lands on the conversation list.

### Testing

Manually tested

#### APK
[Download build #3540](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3540/artifact/build/artifact/wire-dev-PR3336-3540.apk)
[Download build #3541](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3541/artifact/build/artifact/wire-dev-PR3336-3541.apk)


app/src/main/scala/com/waz/zclient/views/ConversationFragment.scala
```diff
@@ -805,13 +805,16 @@ class ConversationFragment extends FragmentHelper {
         R.string.legal_hold_participant_missing_consent_alert_message,
         android.R.string.ok,
         R.string.legal_hold_participant_missing_consent_alert_negative_button,
-        null,
+        new DialogInterface.OnClickListener {
+          override def onClick(dialog: DialogInterface, which: Int): Unit =
+            errorsController.dismissSyncError(err.id).map(_ => getActivity.onBackPressed())
```
gizemb, line 810: why do we need onBackPressed?

johnxnguyen: After creating the conversation, we push the conversation fragment, then show the alert. When tapping OK, we delete the conversation. I don't think it makes sense to stay on the conversation fragment, since the conversation is now deleted, so going back to the conversation list is the only other option.

gizemb: ok, makes sense

gizemb (approved):

PR #3340, opened 2021-05-27
[Fix] Send anyway alert shown when receiving a message

johnxnguyen:
### Issues

The "send anyway" confirmation alert is shown when receiving a message after legal hold has been discovered in a conversation. It should only be presented the first time the self user tries to send a message.

### Causes

If read receipts are enabled, then when we receive a message the app will send confirmation message back to the sender of the received message. Since we send confirmation messages just like any message, we would ask the user for confirmation.

### Solutions

We only want to show the alert for non-hidden messages (indeed this was in the spec but overlooked on my part). 

Hidden messages include confirmation messages (read and delivery receipts), calling messages, button actions (alarming integration responses), data transfer message, reactions, delete, cleared and last read messges.  Visible messages include text (including edits and replies), asset, knock (ping), location.

To achieve this, I added a `ignoreLegalHoldStatus` flag to `postOtrMessage`. This should be set to `true` for all hidden messages and `false` for all visible messages. When `true`, we allow sending of the message regardless of the conversation's legal hold status.

### Testing

Added a test that we can post a message successfully in a conversation with a pending legal hold status when the `ignoreLegalHoldStatus` is `true`.

Also did manual tests to confirm that the alert is not shown when receiving a message in a conversation that has read receipts enabled.

#### APK
[Download build #3545](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3545/artifact/build/artifact/wire-dev-PR3340-3545.apk)
[Download build #3550](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3550/artifact/build/artifact/wire-dev-PR3340-3550.apk)


zmessaging/src/main/scala/com/waz/sync/otr/OtrSyncHandler.scala
```diff
@@ -79,11 +80,24 @@ class OtrSyncHandlerImpl(teamId:             Option[TeamId],
...
+   *
+   * @param convId The id conversation to post the message to.
+   * @param message The message being posted.
+   * @param targetRecipients Who should receive the message.
+   * @param isHidden Whether the message is a hidden message. As a rule of thumb, if it doesn't
+   *                 occupy a row in the message list, then it is hidden.
+   * @param nativePush
+   * @param enforceIgnoreMissing When missing clients should be ignored.
+   * @return
+   */
```
gizemb, line 94: 👍

gizemb (approved):

zmessaging/src/main/scala/com/waz/sync/otr/OtrSyncHandler.scala
johnxnguyen: There were some things I didn't know how to document.

PR #3344, opened 2021-05-28
[Calling] fix : calling layout is not properly handled by the feature flag

mejdoo:
## What's new in this PR?

I forgot to handle properly `fragment_calling.xml` with the feature flag for large conference calls. To solve the failing tests for the RC run, I created a temporary `fargment_new_calling.xml` and I restored  `fragment_calling.xml` to the previous state.

#### APK
[Download build #3559](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3559/artifact/build/artifact/wire-dev-PR3344-3559.apk)
[Download build #3560](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3560/artifact/build/artifact/wire-dev-PR3344-3560.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/calling/NewCallingFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/NewCallingFragment.scala
@@ -19,7 +19,7 @@ package com.waz.zclient.calling
 
 import android.os.Bundle
 import android.view.{LayoutInflater, View, ViewGroup}
-import android.widget.{FrameLayout, LinearLayout, TextView, Toast}
+import android.widget.{FrameLayout, LinearLayout, Toast}
 import androidx.cardview.widget.CardView
 import androidx.gridlayout.widget.GridLayout
 import com.waz.avs.VideoPreview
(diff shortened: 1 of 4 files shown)


app/src/main/scala/com/waz/zclient/calling/NewCallingFragment.scala
```diff
@@ -148,8 +147,6 @@ class NewCallingFragment extends FragmentHelper {
 
   private def initNoActiveSpeakersLayout(): Unit = {
 
-    noActiveSpeakersText.foreach(_.setText(getString(R.string.calling_no_active_speakers)))
```
gizemb, line 151: should we delete this as well?

gizemb (approved):

app/src/main/scala/com/waz/zclient/calling/NewCallingFragment.scala
mejdoo: yes it's handled by the layouts now

PR #3356, opened 2021-06-03
Fix by @kongpf8848: UTF-16 encoding for chinese characters

gizemb:
### Issues

A duplicate of PR #3052 since original PR is no longer maintained by its author.

### Solutions

Same as original PR

### Testing

Unit tests are added for received mentions, with new cases for Chinese text. Also manually tested. You may see the bug in the first line of the image. Other lines are tests after applying the fix:

#### APK
[Download build #3582](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3582/artifact/build/artifact/wire-dev-PR3356-3582.apk)
[Download build #3584](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3584/artifact/build/artifact/wire-dev-PR3356-3584.apk)

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/MessageData.scala
+++ b/zmessaging/src/main/scala/com/waz/model/MessageData.scala
@@ -602,21 +602,16 @@ object MessageData extends DerivedLogTag {
   private val UTF_16_CHARSET  = Charset.forName("UTF-16")
 
   private def encode(text: String) = {
-    if (text == null) {
+    val bytes = UTF_16_CHARSET.encode(text).array
+    /**
+     * UTF-16BE,first two bytes with 0xFEFF
+     * UTF-16LE,first two bytes with 0xFFFE
+     */
+    if (bytes.length < 3 || bytes.slice(2, bytes.length).forall(_ == 0))
       Array.empty[Byte]
-    }
     else {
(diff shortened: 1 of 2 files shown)


zmessaging/src/main/scala/com/waz/model/MessageData.scala
```diff
@@ -602,14 +602,22 @@ object MessageData extends DerivedLogTag {
   private val UTF_16_CHARSET  = Charset.forName("UTF-16")
 
   private def encode(text: String) = {
-    val bytes = UTF_16_CHARSET.encode(text).array
-
-    if (bytes.length < 3 || bytes.slice(2, bytes.length).forall(_ == 0))
+    if (text == null) {
```
makingthematrix, line 605: The text should never be null so this part should not be necessary. Or should it be `if (text.isEmpty)`?

makingthematrix (changes requested):

zmessaging/src/main/scala/com/waz/model/MessageData.scala
gizemb: I couldn't find a scenario where this could receive null or empty text in practice. also looks like `slice` function does check the length and is a safe operation, so I decided to erase this check.

makingthematrix (approved):

PR #3358, opened 2021-06-03
Feature: Build Config - Add a build config to keep web socket on by default

gizemb:
### Issues

Jira: [SQSERVICES-497](https://wearezeta.atlassian.net/browse/SQSERVICES-497)

Some customers want web socket to keep running by default even when app is killed.

### Solutions

Added a new flag `keep_websocket_on` to default.json. When this flag is set to true during custom builds, the websocket will be on by default. User may still turn the switch off via Settings > Advanced.

### Testing

Manually tested.

#### APK
[Download build #3586](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3586/artifact/build/artifact/wire-dev-PR3358-3586.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/WireApplication.scala
+++ b/app/src/main/scala/com/waz/zclient/WireApplication.scala
@@ -479,12 +479,21 @@ class WireApplication extends MultiDexApplication with WireContext with Injectab
         verbose(l"never checked for play services")
         googleApi.isGooglePlayServicesAvailable.head.foreach { gpsAvailable =>
           for {
-            _ <- prefs(GlobalPreferences.WsForegroundKey) := !gpsAvailable
+            _ <- if (!gpsAvailable) prefs(GlobalPreferences.WsForegroundKey) := true
+                 else updateWebSocketForBuildConfig(prefs)
             _ <- prefs(GlobalPreferences.CheckedForPlayServices) := true
           } yield ()
         }
-      case true =>
+      case true => updateWebSocketForBuildConfig(prefs)
     }
+
+  private def updateWebSocketForBuildConfig(prefs: GlobalPreferences): Future[Unit] =
+    for {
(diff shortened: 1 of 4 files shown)


app/src/main/scala/com/waz/zclient/WireApplication.scala
```diff
@@ -479,13 +479,22 @@ class WireApplication extends MultiDexApplication with WireContext with Injectab
...
     }
 
+  private def updateWebSocketForBuildConfig(prefs: GlobalPreferences): Future[Unit] =
+    for {
+      checked <- prefs(GlobalPreferences.CheckedWebSocketConfig).apply()
+      _       <- if (!checked && BuildConfig.KEEP_WEBSOCKET_ON) prefs(GlobalPreferences.WsForegroundKey) := true
+                 else Future.successful(())
+      _       <- prefs(GlobalPreferences.CheckedWebSocketConfig) := true
+    } yield ()
+
```
makingthematrix, line 497: How about this:
If we have `BuildConfig.KEEP_WEBSOCKET_ON` then we don't have to call `checkForPlayServices` at all. Instead we can set `prefs(GlobalPreferences.WsForegroundKey) := true` at once. And otherwise, if `KEEP_WEBSOCKET_ON` is false, then we can keep everything as before.

makingthematrix (changes requested):

app/src/main/scala/com/waz/zclient/WireApplication.scala
gizemb: if we only rely on `BuildConfig.KEEP_WEBSOCKET_ON`, when the user manually turns it off, the flag will be set to true next time the app opens again. I guess this would be a bad experience for the user.

gizemb: I don't understand why we don't have to call checkForPlayServices?

makingthematrix (approved):

PR #3375, opened 2021-06-17
fix: Send a message to a large conversation

makingthematrix:
fixes https://wearezeta.atlassian.net/browse/SQCORE-532

In fact, this fixes two problems:
1. In OtrClientsSyncHandler, when we load prekeys, there is a limit to the size of the request we can do. If we try to load prekeys for more than 128 clients at once, the backend will return an error "413 Request Entity Too Large" which the Android app can't handle. The fix divides the request into smaller chunks, send a request for each chunk, and then merges the responses.
2. In OtrSyncHandler, we encrypt the message for all clients - the more clients, the bigger will be the encrypted message. Then, if the encrypted message is bigger than the limit we set, instead of sending it in a regular way, we try to send it as "External" - a blob of data which we hope the backend will be able to handle. But here comes the bug: after turning the message into the "External" we again check its size against the limit, and if it's still bigger than the limit, we try again to turn it into an "External"... in the end, the procedure crashes and we never send the message. The fix breaks the loop - if we have the message as the "External" we don't care about the limit size anymore.

#### APK
[Download build #3642](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3642/artifact/build/artifact/wire-dev-PR3375-3642.apk)
[Download build #3643](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3643/artifact/build/artifact/wire-dev-PR3375-3643.apk)
[Download build #3644](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3644/artifact/build/artifact/wire-dev-PR3375-3644.apk)
[Download build #3650](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3650/artifact/build/artifact/wire-dev-PR3375-3650.apk)

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/otr/CryptoBoxService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/otr/CryptoBoxService.scala
@@ -39,7 +39,7 @@ import scala.concurrent.Future
 import scala.util.Try
 
 trait CryptoBoxService {
-  val sessions: CryptoSessionService
+  def sessions: CryptoSessionService
   def cryptoBox: Future[Option[CryptoBox]]
   def apply[A](f: CryptoBox => Future[A]): Future[Option[A]]
   def deleteCryptoBox(): Future[Unit]
--- a/zmessaging/src/main/scala/com/waz/service/otr/CryptoSessionService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/otr/CryptoSessionService.scala
@@ -70,7 +70,7 @@ class CryptoSessionServiceImpl(cryptoBox: CryptoBoxService)
     }
 
   def deleteSession(id: SessionId): Future[Unit] = dispatch(id) { cb =>
-    verbose(l"FIX deleteSession($id)")
+    verbose(l"deleteSession($id)")
     cb.foreach(_.deleteSession(id.toString))
   }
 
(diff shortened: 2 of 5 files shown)


vitorhugods (approved):
I was thinking about it a couple of days ago, when implementing/documenting external messages on Reloaded.

I think just accepting the bigger external message sounds sensible. I guess the server should accept it. 

-----
#### Side note

Maybe I'm missing something, as Scala is not my _Muttersprache_. 
What if the content of the message is actually too big, making the external message loop trigger without reaching the 128 clients? I assume this PR doesn't cover it also, right?

It would take about 1k characters (give it or take, depending on the characters themselves, as emoji are bigger, for instance) to be sent to 127 clients to reach 256kb (ignoring the rest of the envelope).

Very edge-case maybe, but something to think about, and would be good to align with server-side, on what's recommended here.

makingthematrix:
@vitorhugods : Thanks for the review :)

Answering your question - look at zmessaging/src/main/scala/com/waz/sync/otr/OtrSyncHandler.scala
The number of clients is important there because the more clients, the bigger will be the encrypted message (we need to encrypt it for each client). But in OtrSyncHandler we don't check the clients number limit, only the size limit. So the fix will work just as well in situations when the message is too big for another reason.

PR #3376, opened 2021-06-18
Feature: Deep Link - Part 4- Handle conversation/join response, open conversation after joining

gizemb:
### Issues

[SQSERVICES-500](https://wearezeta.atlassian.net/browse/SQSERVICES-500)

As a response to conversation/join BE call, 
- if the user is already a member, we get 204 with empty body. Show an error in that case. 
- if the user successfully joined, we get 200 with a MemberJoinEvent as body. Process that event as if it came from EventStream, and open the conversation afterwards.

### Solutions

Added a `MemberJoinEvent` response deserializer. Passed the response to `processConversationEvent` method, which is also used by `convStateEventProcessingStage`.

### Testing

Manually tested with using deep links
- for a new conversation
- for a conversation which the user is already a member of
- for a conversation which the user was used to be a member, but later left

#### APK
[Download build #3645](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3645/artifact/build/artifact/wire-dev-PR3376-3645.apk)
[Download build #3646](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3646/artifact/build/artifact/wire-dev-PR3376-3646.apk)

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/conversation/ConversationsService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/conversation/ConversationsService.scala
@@ -699,7 +699,7 @@ class ConversationsServiceImpl(teamId:          Option[TeamId],
     }
   }
 
-  override def joinConversation(key: String, code: String): Future[Either[GuestRoomStateError, Option[ConvId]]] = {
+  override def joinConversation(key: String, code: String): Future[Either[GuestRoomStateError, Option[ConvId]]] =
     client.postJoinConversation(key, code).future.flatMap {
       case Right(Some(event: MemberJoinEvent)) =>
         for {
@@ -715,7 +715,6 @@ class ConversationsServiceImpl(teamId:          Option[TeamId],
         warn(l"joinConversation(key: $key, code: $code) error: $error")
         Future.successful(Left(GeneralError))
     }
-  }
 }
 
 object ConversationsService {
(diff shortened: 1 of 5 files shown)


zmessaging/src/main/scala/com/waz/service/conversation/ConversationsService.scala
```diff
@@ -691,14 +699,21 @@ class ConversationsServiceImpl(teamId:          Option[TeamId],
     }
   }
 
-  override def joinConversation(key: String, code: String): Future[Either[GuestRoomStateError, Unit]] = {
-    client.postJoinConversation(key, code).future.map {
-      case Right(_)                                          => Right(())
-      case Left(ErrorResponse(_, _, "no-conversation-code")) => Left(NotAllowed)
-      case Left(ErrorResponse(_, _, "too-many-members"))     => Left(MemberLimitReached)
+  override def joinConversation(key: String, code: String): Future[Either[GuestRoomStateError, Option[ConvId]]] = {
```
mejdoo, line 702: Unnecessary `{}`

mejdoo (approved):
Left a minor comment

zmessaging/src/main/scala/com/waz/service/conversation/ConversationsService.scala
gizemb: done!

PR #3401, opened 2021-07-06
[Feature] Add compile time flag to force constant bitrate calls

johnxnguyen:
## What's new in this PR?

**Jira:** https://wearezeta.atlassian.net/browse/SQSERVICES-607

A white label customer requires that all calls use constant bitrate encoding. This PR adds a compile time flag to force CBR calls. When it is enabled, then the switch to use variable bitrate encoding (in the options screen) is not visible, and all calls will be started or joined with CBR enabled.

### Testing

Tested manually.

## Notes

Configurations that use the flags can be seen here: https://github.com/wireapp/android-customization/pull/8

#### APK
[Download build #3706](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3706/artifact/build/artifact/wire-dev-PR3401-3706.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/preferences/pages/OptionsView.scala
+++ b/app/src/main/scala/com/waz/zclient/preferences/pages/OptionsView.scala
@@ -25,7 +25,7 @@ import android.provider.Settings
 import android.text.TextUtils
 import android.util.AttributeSet
 import android.view.View
-import android.widget.LinearLayout
+import android.widget.{LinearLayout, TextView}
 import com.waz.content.UserPreferences._
 import com.waz.media.manager.context.IntensityLevel
 import com.waz.service.{UiLifeCycle, ZMessaging}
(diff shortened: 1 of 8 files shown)


app/src/main/scala/com/waz/zclient/preferences/pages/OptionsView.scala
```diff
@@ -129,6 +130,12 @@ class OptionsViewImpl(context: Context, attrs: AttributeSet, style: Int) extends
     hideScreenContentSwitch.pref.foreach(_ := true)
   }
 
+  if (BuildConfig.FORCE_CONSTANT_BITRATE_CALLS) {
+    callsCategoryTitle.setVisible(false)
```
johnxnguyen, line 134: We need to hide also the title because otherwise we will have an empty section.

mejdoo (approved):
Looks good to me 👍

mejdoo:
@johnxnguyen good to merge :)

PR #3413, opened 2021-07-14
fix: conversation-join deep link doesn't work

makingthematrix:
fixes https://wearezeta.atlassian.net/browse/SQSERVICES-638

The problem was that on Android we expected the deep link with the format
wire://conversation-join/?key=key&code=code
And on other platforms it's
wire://conversation-join?key=key&code=code
without the slash / after conversation-join

We needed that / to properly parse the deep link. I made some quick changes to make it work.
It would be nice to write it better in the future.

#### APK
[Download build #3735](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3735/artifact/build/artifact/wire-dev-PR3413-3735.apk)


mejdoo (approved):

makingthematrix:
build again

PR #3422, opened 2021-08-09
[Calling]: Large conference calls - Adjust incoming calling UI with large video conference calling

mejdoo:
## What's new in this PR?

Updated incoming calls UI to fit with large video conference calling design spec : https://zpl.io/VYq9DpJ
#### APK
[Download build #3750](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3750/artifact/build/artifact/wire-dev-PR3422-3750.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
@@ -244,14 +244,24 @@ class CallController(implicit inj: Injector, cxt: WireContext)
     case _ => Signal.const[Option[UserData]](None)
   }
 
-  val memberForPicture: Signal[Option[UserId]] = isGroupCall.flatMap {
-    case true  => Signal.const(None)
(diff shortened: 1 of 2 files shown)


app/src/main/scala/com/waz/zclient/calling/controllers/CallController.scala
```diff
@@ -244,14 +244,24 @@ class CallController(implicit inj: Injector, cxt: WireContext)
...
+        } yield member
+      }
+     else
+      isGroupCall.flatMap {
+        case true => Signal.const(None)
+        case false =>
+          for {
+            self <- callingZms.map(_.selfUserId)
+            member <- conversationMembers.map(_.find(m => m._1 != self).map(_._1))
+          } yield member
```
ohassine, line 263: Maybe we can extract `for .. yield` bloc to a separate method to avoid duplication

ohassine (approved):

PR #3434, opened 2021-08-16
Federation: Add the conversation domain and the sender ("from") domain to events

makingthematrix:
On new backends each conversation event will contain data about the domain its coming from:
the domain of the conversation and also the domain of the sender, if applicable.
This PR is mostly just rewriting the Event class where after adding these two fields a lot of other code
had to be changed. I wrote JSON parsing in such way that if for some reason the Android app is connected
to the old backend which sends events without domains, the event will be handled properly and
the domains will be set to None.

Changes in other classes are only side-effects, like in pattern matching where now we need to take into
account that there are more fields in the given event class.

#### APK
[Download build #3786](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3786/artifact/build/artifact/wire-dev-PR3434-3786.apk)


makingthematrix:
build again

mejdoo (approved):

PR #3435, opened 2021-08-16
File sharing restrictions (part 10)

johnxnguyen:
## What's new in this PR?

**Jira:** https://wearezeta.atlassian.net/browse/SQSERVICES-612

In this PR we add restriction views for file messages (see screen shot). 

I needed to add a new icon (blocked file) but I see that we don't seem to use paint code anymore and instead prefer adding SVG resources. However, a lot of UI still relies on paint code generated draw methods, so  using `ControlsView` as a guide, I manually created the draw method and placed it inside a helper class call `StyleKitMethods`.

### Testing

Manually tested

## Notes

Wire 2021-08-16 at 4 46 PM

#### APK
[Download build #3781](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3781/artifact/build/artifact/wire-dev-PR3435-3781.apk)
[Download build #3794](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3794/artifact/build/artifact/wire-dev-PR3435-3794.apk)
[Download build #3795](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3795/artifact/build/artifact/wire-dev-PR3435-3795.apk)
[Download build #3796](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3796/artifact/build/artifact/wire-dev-PR3435-3796.apk)


app/src/main/scala/com/waz/zclient/conversation/ReplyView.scala
```diff
@@ -71,27 +71,30 @@ class ReplyView(context: Context, attrs: AttributeSet, defStyle: Int) extends Fr
...
+          set(getString(R.string.reply_message_type_image), bold = true, Some(WireStyleKit.drawImage), messageData.assetId)
+        case Type.AUDIO_ASSET =>
+          set(getString(R.string.reply_message_type_audio), bold = true, Some(WireStyleKit.drawVoiceMemo), None)
+        case Type.ANY_ASSET =>
+          val assetName = asset match {
+            case Some(a: Asset) => a.name
+            case _ => getString(R.string.reply_message_type_asset)
+          }
+
+          set(assetName, bold = true, Some(if (isRestricted) StyleKitMethods().drawFileBlocked else WireStyleKit.drawFile), None)
```
makingthematrix, line 92: Since `isRestricted` is used only here, not for any other type of an asset, can we check if `isFileSharingRestricted` only here? 
Also, I think instead of having `isFileSharingRestricted`  as a signal, and then calling `.head`, you can check the property directly.

makingthematrix (changes requested):

app/src/main/scala/com/waz/zclient/conversation/ReplyView.scala
johnxnguyen: I've updated this: there is no long the signal, and I moved minimized the scope of `isRestricted`.

makingthematrix (approved):

PR #3445, opened 2021-08-20
Feature: Enforce ephemeral messages (part 0)

vitorhugods:
## What's new

Jira: https://wearezeta.atlassian.net/browse/SQSERVICES-572

A mechanism that allows team admins to disable or disable the self-deleting messages (ephemeral messages) and optionaly enforce a timeout in seconds for them.

This first PR aims to fetch and update the new flag together with the file sharing restrictions flag recently put in place by @johnxnguyen.

In the next PRs I will tackle the receiving of flag updates via notifications and the actual usage of this flag.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/SelfDeletingMessagesFeatureConfig.scala
+++ b/zmessaging/src/main/scala/com/waz/model/SelfDeletingMessagesFeatureConfig.scala
@@ -1,0 +1,26 @@
+package com.waz.model
+
+import com.waz.utils.JsonDecoder
+import org.json.JSONObject
+
+
+final case class SelfDeletingMessagesFeatureConfig(status: String, seconds: Int) {
+
+
+  def isEnabled: Boolean = status == "enabled"
+
+  def enforcedTimeoutInSeconds: Int = Math.max(0, seconds)
+
+}
+
+object SelfDeletingMessagesFeatureConfig {
+
+  val Default: SelfDeletingMessagesFeatureConfig = SelfDeletingMessagesFeatureConfig(status = "enabled", 0)
+
+  implicit object Decoder extends JsonDecoder[SelfDeletingMessagesFeatureConfig] {
+    override def apply(implicit js: JSONObject): SelfDeletingMessagesFeatureConfig =
+      if (!js.has("status") || !js.has("enforcedTimeoutSeconds")) Default
+      else SelfDeletingMessagesFeatureConfig(js.getString("status"), js.getInt("enforcedTimeoutSeconds"))
+  }
+
+}
--- a/zmessaging/src/main/scala/com/waz/model/SelfDeletingMessagesFeatureConfig.scala
+++ b/zmessaging/src/main/scala/com/waz/model/SelfDeletingMessagesFeatureConfig.scala
@@ -3,9 +3,7 @@ package com.waz.model
 import com.waz.utils.JsonDecoder
 import org.json.JSONObject
 
-
 final case class SelfDeletingMessagesFeatureConfig(status: String, seconds: Int) {
-
 
   def isEnabled: Boolean = status == "enabled"
 
--- a/zmessaging/src/main/scala/com/waz/service/teams/FeatureConfigsService.scala
+++ b/zmessaging/src/main/scala/com/waz/service/teams/FeatureConfigsService.scala
@@ -5,7 +5,7 @@ import com.waz.content.UserPreferences._
 import com.waz.log.BasicLogging.LogTag.DerivedLogTag
 import com.waz.sync.handler.FeatureConfigsSyncHandler
 import com.waz.log.LogSE._
-import com.waz.model.{FeatureConfigEvent, FeatureConfigUpdateEvent, FileSharingFeatureConfig}
+import com.waz.model.{FeatureConfigEvent, FeatureConfigUpdateEvent, FileSharingFeatureConfig, SelfDeletingMessagesFeatureConfig}
 import com.waz.service.EventScheduler
 import com.waz.service.EventScheduler.Stage
 import com.waz.utils.JsonDecoder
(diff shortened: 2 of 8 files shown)


zmessaging/src/main/scala/com/waz/service/teams/FeatureConfigsService.scala
```diff
@@ -69,4 +70,25 @@ class FeatureConfigsServiceImpl(syncHandler: FeatureConfigsSyncHandler,
...
+    for {
+      wasEnabled        <- userPrefs(AreSelfDeletingMessagesEnabled).apply()
+      lastKnownTimeout  <- userPrefs(SelfDeletingMessagesEnforcedTimeout).apply()
+      isNowEnabled      =  config.isEnabled
+      newTimeout        =  config.enforcedTimeoutInSeconds
+      _                 <- userPrefs(AreSelfDeletingMessagesEnabled) := isNowEnabled
+      _                 <- userPrefs(SelfDeletingMessagesEnforcedTimeout) := newTimeout
+      // Inform of new restrictions.
+      _                 <- userPrefs(ShouldInformSelfDeletingMessagesChanged) :=
+        (wasEnabled != isNowEnabled || lastKnownTimeout != newTimeout)
```
johnxnguyen, line 91: It may not be updated in the design specs yet, but I think we want to show an info alert for all changes:

- "self deleting messages are now disabled"
- "self deleting messages are now enabled"
- "self deleting messages are now enforced with a X second timeout"

So probably it means we should always set the flag `ShouldInformSelfDeletingMessagesChanged` to `true`, then determine when presenting the alert which of the three cases is currently relevant.

zmessaging/src/main/scala/com/waz/sync/client/FeatureConfigsClient.scala
```diff
@@ -34,12 +35,19 @@ class FeatureConfigsClientImpl(implicit
     .withResultType[FileSharingFeatureConfig]
     .withErrorType[ErrorResponse]
     .executeSafe
+
+  override def getSelfDeletingMessages(): ErrorOrResponse[SelfDeletingMessagesFeatureConfig] =
+    Request.Get(relativePath =  fileSharingPath)
```
johnxnguyen, line 40: I think you mean `relativePath = selfDeletingMessages`

zmessaging/src/test/scala/com/waz/model/SelfDeletingMessagesFeatureConfigSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+
+    scenario("Deserializing an enabled config with enforced timeout") {
+      // Given
+      val json =
+        """
+          |{
+          |  "status": "enabled",
+          |  "enforcedTimeoutSeconds": 30
+          |}
+          |""".stripMargin
```
johnxnguyen, line 18: As I mentioned in a previous comment, the structure here is not quite right. There are two scenarios:

```
{
    "status": "disabled"
}
```

and

```
{
    "status": "enabled",
    "config": {
        "enforcedTimeoutSeconds": 30
    }
}
```

zmessaging/src/main/scala/com/waz/service/teams/FeatureConfigsService.scala
vitorhugods: I'm only setting it to true if there's a change. Any change.

The only way to make it `false` is if `wasEnabled == isNowEnabled && lastKnownTimeout == newTimeout` _i.e._ nothing changed.

It's just to prevent the repeating display of the same configuration, without any changes.

zmessaging/src/main/scala/com/waz/model/SelfDeletingMessagesFeatureConfig.scala
```diff
@@ -0,0 +1,24 @@
...
+
+object SelfDeletingMessagesFeatureConfig {
+
+  val Default: SelfDeletingMessagesFeatureConfig = SelfDeletingMessagesFeatureConfig(status = "enabled", 0)
+
+  implicit object Decoder extends JsonDecoder[SelfDeletingMessagesFeatureConfig] {
+    override def apply(implicit js: JSONObject): SelfDeletingMessagesFeatureConfig =
+      if (!js.has("status") || !js.has("enforcedTimeoutSeconds")) Default
+      else SelfDeletingMessagesFeatureConfig(js.getString("status"), js.getInt("enforcedTimeoutSeconds"))
+  }
```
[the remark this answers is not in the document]
vitorhugods: I remembered that I missed the `config` object a few hours into a flight last friday. Thinking about life and stuff 😄 

Good eye, and thanks for clarifying that the object may be missing. That part I would've missed completely. I'll fix this now.

zmessaging/src/main/scala/com/waz/model/SelfDeletingMessagesFeatureConfig.scala
```diff
@@ -3,22 +3,26 @@ package com.waz.model
...
 
-  val Default: SelfDeletingMessagesFeatureConfig = SelfDeletingMessagesFeatureConfig(status = "enabled", 0)
+  val Default: SelfDeletingMessagesFeatureConfig = SelfDeletingMessagesFeatureConfig(isEnabled = true, 0)
 
   implicit object Decoder extends JsonDecoder[SelfDeletingMessagesFeatureConfig] {
     override def apply(implicit js: JSONObject): SelfDeletingMessagesFeatureConfig =
-      if (!js.has("status") || !js.has("enforcedTimeoutSeconds")) Default
-      else SelfDeletingMessagesFeatureConfig(js.getString("status"), js.getInt("enforcedTimeoutSeconds"))
+      if (!js.has("status")) Default
+      else if (js.getString("status") != "enabled") SelfDeletingMessagesFeatureConfig(isEnabled = true, 0)
```
johnxnguyen, line 19: If `status` is not `enabled` then `isEnabled` should be `false`.

vitorhugods: 🤦🏼 Fixed

johnxnguyen (approved):

ohassine (approved):

PR #3447, opened 2021-08-20
Feature: Enforce ephemeral messages (part 2)

vitorhugods:
Continuation of #3445

## What's new in this PR?

Display a dialog to the user informing changes in the SelfDeletingMessages feature config.

### Testing
 Tested manually

Screenshot_1629446853
Screenshot_1629447221

#### APK
[Download build #3876](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3876/artifact/build/artifact/wire-dev-PR3447-3876.apk)
[Download build #3894](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3894/artifact/build/artifact/wire-dev-PR3447-3894.apk)

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -1646,8 +1646,10 @@
     <string name="join_conversation_invalid_link_error_message">The conversation link is invalid.</string>
     <string name="join_conversation_member_limit_reached_error_message">The conversation is full.</string>
 
+    <!-- Feature Config Changes -->
+    <string name="feature_config_changed_info_dialog_title">There has been a change in Wire</string>
+
     <!-- File sharing feature config -->
-    <string name="file_sharing_restriction_info_dialog_title">There has been a change in Wire</string>
     <string name="file_sharing_restriction_info_dialog_message">Sharing and receiving files of any type is now disabled.</string>
 
     <string name="file_sharing_restriction_info_image">RECEIVING IMAGES RESTRICTED</string>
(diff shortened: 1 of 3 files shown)


app/src/main/res/values/strings.xml
```diff
@@ -1656,5 +1656,11 @@
     <string name="file_sharing_restriction_info_file">RECEIVING FILES RESTRICTED</string>
     <string name="file_sharing_restriction_info_generic">FILE SHARING RESTRICTED</string>
 
+    <!-- Self Deleting Messages feature config -->
+    <string name="self_deleting_messages_change_info_dialog_title">There has been a change in Wire</string>
```
johnxnguyen, line 1660: This title should be the same for all feature config changes, do you think we should have a single string entry rather than one for each feature?

vitorhugods: I was not sure if the title should be the same. I was locked out of the design specs at the time so I left it here mainly as a placeholder until later.

Looking at the design specs, I see the text is supposed to be "Self-deleting messages restrictions".

Now I have more questions actually. What if on top of this change, there were more config changes. Like disabling file sharing? 

Maybe from a user point of view we should take all these changes, and compile them together?

I'm not fond of dialogs stacking-up (especially if titles can be the same). May make things very unclear.

johnxnguyen: Yeah I think the design specs still need updating. I'm also not a fan of popups, especially when they stack, and neither is the design team. But I think for the moment it is just a simplification. 

I already spoke with design saying that this approach won't scale well and we need some other non intrusive way to do this.

PR #3450, opened 2021-08-20
[Fix] File sharing automation issues

johnxnguyen:
## What's new in this PR?

This PR includes fixes for the following automation issues:
- The header views in the collection screen can't be located, so an id has been added to this view.
- In the collection image view, automation can't tell when the image is restricted. The icon view that is displayed for the cases when the image is expired or restricted has be replaced with two icon views, one for expired images, and one for restricted images. This allows automation to know when the image is restricted.


johnxnguyen:
build again

PR #3457, opened 2021-08-23
Federation: Refactoring to use qualified ids in zmessaging (part 1)

makingthematrix:
You can follow titles of respective commits to see how the refactoring went from one stage to the other.
1. Use qualified ids only for syncing clients - I simplified a bit the code for syncing clients. I removed the old `SyncClients` command and the associated methods. Now in all cases the code for qualified ids will be used (but it still respects the federation flag - if it's off, we're only interested in `userId`s in qualified ids, not the domains. 
2. Use domains in generated qualified ids - Until now an empty string for the domain was the sign that we use the current domain. From now on, qualified ids will always carry the domain, also the domain of the current user. This is necessary, because new OTR events always carry the domain of the conversation and the sender and we need to be able to compare those domains with domains of other users. It's easier to do it if we can be sure that the domain is always there.
3. Migrate users data - There is already the "domain" field in `UserData` (also in DB) but until now it was empty for users from the current domain. Now, these will be changed to the current domain.
4. Wrap the map of user ids to client ids in `OtrClientIdMap` - In a lot of places in zmessaging we use maps of the type `Map[UserId, Set[ClientId]]` or `Map[UserId, Seq[ClientId]]`. I wrapped them into a new type, `OtrClientIdMap`. In future I want to change `UserId` in there to `QualifiedId` and it will be easier if I have this code in one place, so to say. But before I do that, I want to work on prekeys and basic messaging. So that's why refactoring stops at this point.

#### APK
[Download build #3831](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3831/artifact/build/artifact/wire-dev-PR3457-3831.apk)
[Download build #3845](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3845/artifact/build/artifact/wire-dev-PR3457-3845.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/common/controllers/global/ClientsController.scala
+++ b/app/src/main/scala/com/waz/zclient/common/controllers/global/ClientsController.scala
@@ -51,9 +51,7 @@ class ClientsController(implicit inj: Injector) extends Injectable with DerivedL
 
   def client(userId: UserId, clientId: ClientId): Signal[Option[Client]] = for {
     manager <- accountManager
-    _ = verbose(l"FIX client($userId, $clientId)")
     clients <- manager.storage.otrClientsStorage.signal(userId)
-    _ = verbose(l"FIX ${clients.clients.keys}")
   } yield clients.clients.get(clientId)
 
   def selfClient(clientId: ClientId): Signal[Option[Client]] = for {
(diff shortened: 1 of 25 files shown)


makingthematrix:
build again

mejdoo (approved):

PR #3504, opened 2021-09-14
[wip] fdroid build flavour

obfusk:


app/src/main/scala/com/waz/zclient/common/controllers/BrowserController.scala
```diff
@@ -51,12 +53,22 @@ class BrowserController(implicit context: Context, injector: Injector) extends I
...
+      location.getName
+    ) foreach { i =>
+      try {
+        Log.i("BrowserController", s"trying $i...")
+        context.startActivity(i)
+        Log.i("BrowserController", "OK")
+        return
+      } catch {
+        case _: ActivityNotFoundException => Log.i("BrowserController", "failed")
+      }
```
makingthematrix, line 70: Please change to something like:
```
Try {
        info(s"trying $i...")
        context.startActivity(i)
        info("OK")
}.recover { _ => info("failed" }
```

obfusk: ack.

obfusk: The logging was mostly for testing anyway (and thus logcat was more useful), so I could just remove it completely.

obfusk: I removed it. let me know if you disagree.

app/src/main/scala/com/waz/zclient/common/controllers/BrowserController.scala
```diff
@@ -51,12 +52,13 @@ class BrowserController(implicit context: Context, injector: Injector) extends I
...
   def openLocation(location: Location): Unit =
-    Option(IntentUtils.getGoogleMapsIntent(
+    IntentUtils.getMapIntents(
       context,
       location.getLatitude,
       location.getLongitude,
       location.getZoom,
-      location.getName)) foreach { context.startActivity }
+      location.getName
+    ) foreach { i => if (Try(context.startActivity(i)).isSuccess) return }
```
makingthematrix, line 61: Change to `.foreach(Try(context.startActivity(_)))`

app/src/main/scala/com/waz/zclient/pages/main/conversation/LocationFragment.scala
```diff
@@ -0,0 +1,483 @@
...
+import org.osmdroid.util.GeoPoint
+import org.osmdroid.views.CustomZoomButtonsController.Visibility
+import org.osmdroid.views.MapView
+import org.osmdroid.views.overlay.Marker
+
+import java.util._
+import scala.util.control.Exception._
+
+// FIXME: use more idiomatic scala (Option, lazy val etc.)
+class LocationFragment extends BaseFragment[LocationFragment.Container]
```
makingthematrix, line 62: This class needs refactoring and testing (at least manual tests to ensure all functions work).

makingthematrix (changes requested):

app/src/main/scala/com/waz/zclient/common/controllers/BrowserController.scala
obfusk: It's supposed to start either osmand+, osmand, google maps, or open a browser link; whichever succeeds first.

The suggested change would do *all of those* (that are possible, depending on what's installed) at the same time, not just the first successful one (and then stop trying the rest).

obfusk: The old code used `resolveActivity` to find out which of these can succeed without trying them, which turned out not to work and always end up using the browser fallback even if some of the other apps are installed. Hence this alternative.

app/src/main/scala/com/waz/zclient/pages/main/conversation/LocationFragment.scala
obfusk: > This class needs refactoring

I would have been happy to do that earlier. But my contract with wire ended yesterday.

Also: I think it's already a lot better than the current Java code that it replaces.

> and testing (at least manual tests to ensure all functions work).

Manual testing was done by myself and @mythsunwind, so all functions work.

app/src/main/scala/com/waz/zclient/common/controllers/BrowserController.scala
Gardosen: @makingthematrix @obfusk ' explanation kind of makes sense, we should leave it like it is.

Gardosen (approved):
i approve this as a side standing person over the entire time of the development of this. 

LGTM! 👍

app/src/main/scala/com/waz/zclient/common/controllers/BrowserController.scala
makingthematrix: Ah, okay, so if it should stop after the first successful one then instead of `foreach` we can go with `find`:
```
).find(i => Try(context.startActivity(i)).isSuccess))
```

Sorry, but `return` in Scala is really wrong. 
I can change it later.

app/src/main/scala/com/waz/zclient/pages/main/conversation/LocationFragment.scala
makingthematrix: @obfusk : Okay. I can refactor it later. Thanks for your work!

PR #3508, opened 2021-09-21
Federation: Simplify post conversation

makingthematrix:
I realized that the post conversation endpoint stays the same in federation and non-federation.
A new conversation is always created on the backend of the creator so there's no need to provide
the qualified conversation id. This lets me simplify the code.

On top of that, the fact that now we create new group conversations without members and only add members later - that also let me simplify the code a bit.
#### APK
[Download build #3962](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3962/artifact/build/artifact/wire-dev-PR3508-3962.apk)
[Download build #3975](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3975/artifact/build/artifact/wire-dev-PR3508-3975.apk)

### DIFF
--- a/zmessaging/src/main/java/com/waz/model/sync/SyncCommand.java
+++ b/zmessaging/src/main/java/com/waz/model/sync/SyncCommand.java
@@ -37,7 +37,6 @@ public enum SyncCommand {
     SyncQualifiedSearchResults("sync-qualified-search-results"),
     SyncProperties("sync-properties"),
     PostConv("post-conv"),
-    PostQualifiedConv("post-qualified-conv"),
     PostConvReceiptMode("post-conv-receipt-mode"),
     PostConvName("post-conv-name"),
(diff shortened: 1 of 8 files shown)


zmessaging/src/test/scala/com/waz/model/UserDataSpec.scala
```diff
@@ -54,7 +55,6 @@ class UserDataSpec extends AndroidFreeSpec {
 
       // THEN
       data.id.shouldEqual(referenceInfo.id)
-      data.domain.shouldEqual(referenceInfo.domain)
```
makingthematrix, line 57: This test failed, and rightly so, but the reason why it failed is addressed in another PR: https://github.com/wireapp/wire-android/pull/3507
So here I just delete this line to let the test pass.

mejdoo (approved):

PR #3514, opened 2021-09-22
[Calling] Large Conference calls - Bugfix : tiles are ordered per page but not globally

mejdoo:
### Issues

During the latest internal test for large conference calls, I realised that logic of ordering the tiles was applied per page which is different from the following acceptance criteria: 

Participants in the conference call should appear in the following order:
- me, myself & I (always, with or without the camera on)
- participants with video ON, ordered by alphabet
- participants with video OFF, ordered by alphabet

https://wearezeta.atlassian.net/browse/SQCALL-389

### Solution

Apply the order earlier before passing the list of participants to show in every page.

#### APK
[Download build #3970](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3970/artifact/build/artifact/wire-dev-PR3514-3970.apk)
[Download build #3974](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3974/artifact/build/artifact/wire-dev-PR3514-3974.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/calling/CallingGridFragment.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/CallingGridFragment.scala
@@ -109,7 +109,7 @@ class CallingGridFragment extends FragmentHelper {
           val startIndex = pageNumber * MAX_PARTICIPANTS_PER_PAGE
           val endIndex = startIndex + MAX_PARTICIPANTS_PER_PAGE
 
-          val participantsToShow = (orderedParticipants(participants, participantsInfo.toIdMap, selfClientId).slice(startIndex, endIndex), size) match {
+          val participantsToShow = (orderedParticipants(participants, participantsInfo.toIdMap, selfUserId, selfClientId).slice(startIndex, endIndex), size) match {
             case (ps, 2) => ps.filter(_.clientId != selfClientId)
             case (ps, _) => ps
           }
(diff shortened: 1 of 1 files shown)


app/src/main/scala/com/waz/zclient/calling/CallingGridFragment.scala
```diff
@@ -109,17 +109,26 @@ class CallingGridFragment extends FragmentHelper {
...
+          refreshVideoGrid(grid, selfUserId, selfClientId, participantsToShow, participantsInfo, participants, false)
         case _ =>
       }
     }
   }
 
+  private def orderedParticipants(participants : Set[Participant], infoMap : Map[UserId, CallParticipantInfo], selfClientId : ClientId): Seq[Participant] =
+    participants.toSeq.sortWith {
+      case (p, _) if p.clientId == selfClientId => true
+      case (_, p) if p.clientId == selfClientId => false
```
vitorhugods, line 126: What's the expected behaviour when a user has multiple clients?

Should we consider `p.userId == selfUserId` and order it differently in any way?

I can imagine that being useful if you want to show your face using one device's camera, and something in the office (like a meeting room full of people) using another. Thinking like this it can make sense to put these two clients side by side.

vitorhugods (commented):
Just raised one comment about the possibility of having multiple clients for the same user.

app/src/main/scala/com/waz/zclient/calling/CallingGridFragment.scala
mejdoo: ok good point, I refactored the logic and tested it :
- First : Self user in current device
- Second: Self user in an other device
- Third : Video ON users
- the rest is sorted alphabetically.

makingthematrix (approved):

PR #3516, opened 2021-09-23
[Calling] Bugfix - Calling flickering- part 1

mejdoo:
## What's new in this PR?

This PR aims to reduce calling flickering issues detected during the large conference calls internal tests. The idea here is to avoid useless refresh of the views when the number of pages is stable.
#### APK
[Download build #3977](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3977/artifact/build/artifact/wire-dev-PR3516-3977.apk)
[Download build #3978](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3978/artifact/build/artifact/wire-dev-PR3516-3978.apk)
[Download build #3979](http://10.10.124.11:8080/job/Pull%20Request%20Builder/3979/artifact/build/artifact/wire-dev-PR3516-3979.apk)

### DIFF
--- a/app/src/main/scala/com/waz/zclient/calling/AllParticipantsAdapter.scala
+++ b/app/src/main/scala/com/waz/zclient/calling/AllParticipantsAdapter.scala
@@ -31,15 +31,23 @@ class AllParticipantsAdapter(implicit context: Context, eventContext: EventConte
 
   private val callController = inject[CallController]
 
-  var numberOfParticipants : Int = 0
+  var numberOfParticipants: Int = 0
+  var previousPagesCount: Int = 0
 
-  callController.allParticipants.map(_.size).foreach { numberOfParticipants = _ }
+  callController.allParticipants.map(_.size).foreach { size =>
+    numberOfParticipants = size
+    if (getPagesCount() != previousPagesCount) notifyDataSetChanged()
+    previousPagesCount = getPagesCount()
+  }
 
-  override def getItemCount(): Int = if (numberOfParticipants == 0) 0
-  else if (numberOfParticipants % MAX_PARTICIPANTS_PER_PAGE == 0) numberOfParticipants / MAX_PARTICIPANTS_PER_PAGE
-  else (numberOfParticipants / MAX_PARTICIPANTS_PER_PAGE) + 1
+  override def getItemCount(): Int = getPagesCount()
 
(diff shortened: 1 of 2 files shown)


app/src/main/scala/com/waz/zclient/calling/AllParticipantsAdapter.scala
```diff
@@ -31,16 +31,24 @@ class AllParticipantsAdapter(implicit context: Context, eventContext: EventConte
...
 
-  var numberOfParticipants : Int = 0
+  var numberOfParticipants: Int = 0
+  var previousPagesCount: Int = 0
 
-  callController.allParticipants.map(_.size).foreach { numberOfParticipants = _ }
+  callController.allParticipants.map(_.size).foreach { size =>
+    numberOfParticipants = size
+    if (getPagesCount() != previousPagesCount) notifyDataSetChanged()
+    previousPagesCount = getPagesCount()
```
makingthematrix, line 40: To avoid calling twice `getPagesCount`:
```
getPagesCount match {
  case n if n != previousPagesCount =>
    previousPagesCount = n
    notifyDataChanged()
  case _ =>
}
```

Also, how about changing the name of that method to `countPages`?

makingthematrix (changes requested):

app/src/main/scala/com/waz/zclient/calling/AllParticipantsAdapter.scala
mejdoo: I want to apply always `previousPagesCount = n` even when this condition is not met` if n != previousPagesCount`. Does you solution still correct with this requirement?

mejdoo: What do you think about?

```
getPagesCount match {
  case n if n != previousPagesCount => notifyDataChanged()
case n  =>  previousPagesCount = n
  case _ =>
}
```

mejdoo: To answer your second question, `countPages` sounds good to me.

makingthematrix (approved):

app/src/main/scala/com/waz/zclient/calling/AllParticipantsAdapter.scala
mejdoo: ok got it thanks.

PR #3582, opened 2021-11-03
fix: Users unable to send messages in some conversations

makingthematrix:
This is a hotfix and it's based on manual tests results and code review.
With the Federation feature we added fields for "domains" to the Conversations and to the Users table in db. In the db schema we allow the fields to be null, but when deserializing conversations and users we expect them to be empty strings at least, or to have the actual domain string, but we don't expect null. This might cause the bug and it might also explain why it's not 100% reproducible - the problem would appear only for old conversations and users (which now got those fields initialized to null), and in non-federated environments.

The fix introduces checks if the fields are null and if yes then they return an empty string instead.
This way we avoid another migration. If this is indeed the cause of the bug then after the update the affected apps should start working again.


AndroidBob:
Build Success
ID:1
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3582/1/)

PR #3585, opened 2021-11-09
federation: consider domain when checking if user is a guest

ohassine:
## What's new in this PR?

Now we are considering domain when checking if the user is a guest

### DIFF
--- a/zmessaging/src/main/scala/com/waz/model/UserData.scala
+++ b/zmessaging/src/main/scala/com/waz/model/UserData.scala
@@ -137,9 +137,17 @@ final case class UserData(override val id:       UserId,
     }
   }
 
-  def isGuest(ourTeamId: TeamId): Boolean = isGuest(Some(ourTeamId))
+  def isGuest(ourTeamId: TeamId, ourDomain: Domain = Domain.Empty): Boolean = isGuest(Some(ourTeamId), Some(ourDomain))
 
-  def isGuest(ourTeamId: Option[TeamId]): Boolean = ourTeamId.isDefined && teamId != ourTeamId
+  def isGuest(ourTeamId: Option[TeamId], ourDomain: Option[Domain]): Boolean = {
+    val isSameTeam = teamId == ourTeamId
+    if (BuildConfig.FEDERATION_USER_DISCOVERY) {
+      val isSameDomain = ourDomain.isDefined && ourDomain.get == domain
+      ourTeamId.isDefined && (!isSameTeam || (isSameTeam &&  !isSameDomain))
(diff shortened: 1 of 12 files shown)


AndroidBob:
Build Success
ID:1
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3585/1/)

zmessaging/src/main/scala/com/waz/model/UserData.scala
```diff
@@ -137,9 +137,17 @@ final case class UserData(override val id:       UserId,
     }
   }
 
-  def isGuest(ourTeamId: TeamId): Boolean = isGuest(Some(ourTeamId))
-
-  def isGuest(ourTeamId: Option[TeamId]): Boolean = ourTeamId.isDefined && teamId != ourTeamId
+  def isGuest(ourTeamId: TeamId, ourDomain: Domain = Domain.Empty): Boolean = isGuest(Some(ourTeamId), Some(ourDomain))
```
makingthematrix, line 140: I think using `Option[Domain]` makes the code unnecessarily complicated.
You can always use `Domain.Empty` as default. So, for example, it's safe to use `inject[Signal[Domain]]` even on a private account - in that case you will get a constant signal with the value `Domain.Empty`.

And then you can just compare domain directly: 
```
ourTeamId.isDefined && (teamId != ourTeamId || domain != ourDomain)
```

zmessaging/src/main/scala/com/waz/model/UserData.scala
```diff
@@ -137,9 +137,17 @@ final case class UserData(override val id:       UserId,
...
+
+  def isGuest(ourTeamId: Option[TeamId], ourDomain: Option[Domain]): Boolean = {
```
makingthematrix, line 142: So it means this method probably shouldn't be necessary. Try to use the above one instead of this one everywhere.

makingthematrix (changes requested):

zmessaging/src/main/scala/com/waz/model/UserData.scala
ohassine: Got it

AndroidBob:
Build Success
ID:2
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3585/2/)

makingthematrix (approved):

AndroidBob:
Build Success
ID:3
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3585/3/)

PR #3594, opened 2021-11-11
fix: consider domain only for federation

ohassine:
## What's new in this PR?
Jira : https://wearezeta.atlassian.net/browse/SQCORE-1164

### Issues

User is unable to create 1:1 conversation from search

### Causes

This is due to invalid response from `isGuest()`

When federation is turned off, we still do have a value for domain. So comparing the `domain` with `ourDomain` (which is Empty) invalidates the test.

### Solutions

Consider the domain only for federation 

### Testing

Manually tested


AndroidBob:
Build Success
ID:1
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3594/1/)

makingthematrix (approved):

PR #3599, opened 2021-11-13
fix(config): Change default of conference calling enabled to true JCT-70

tmspzz:
For on-prem installations conference calls should always be possible (unless deliberately configured otherwise). Therefore, the default value should be true.

### Issues

Conference calling is at the moment disabled by default on the client. It should instead be enabled by default.

### Causes

A missing endpoint on an older on-prem installation can trigger the client to behave in such fashion that conference calling is disabled.

### Solutions

Change the client default to be true, thus enabling conference calling for older on-prem installations.

### Testing

GIVEN: An older backend
WHEN: Using the android client
THEN: I am able to make conference calls

### DIFF
--- a/zmessaging/src/main/scala/com/waz/content/Preferences.scala
+++ b/zmessaging/src/main/scala/com/waz/content/Preferences.scala
@@ -446,7 +446,7 @@ object UserPreferences {
   lazy val SelfDeletingMessagesEnforcedTimeout: PrefKey[Int] = PrefKey[Int]("self_deleting_messages_enforced_timeout", customDefault = 0)
   lazy val ShouldInformSelfDeletingMessagesChanged: PrefKey[Boolean] = PrefKey[Boolean]("self_deleting_messages_enforced_changed", customDefault = false)
 
-  lazy val ConferenceCallingFeatureEnabled: PrefKey[Option[Boolean]] = PrefKey[Option[Boolean]]("conference_calling_feature_enabled", customDefault = None)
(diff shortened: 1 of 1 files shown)


AndroidBob:
Build Failure
ID:1
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3599/1/)

AndroidBob:
Build Success
ID:2
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3599/2/)

AndroidBob:
Build Success
ID:3
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3599/3/)

ohassine:
Changing the default value to true will cause another problem.
If for some reason, the app would not able to update `ConferenceCallingFeatureEnabled` from backend, then personal users will be able to start conference calls (which is not allowed)

tmspzz:
> Changing the default value to true will cause another problem.
> If for some reason, the app would not able to update `ConferenceCallingFeatureEnabled` from backend, then personal users will be able to start conference calls (which is not allowed)

That's fine

ohassine (approved):

Gardosen (commented):
why is Some used here in this case?
the declaration is a hardcoded value. 

isn't this performance wise impractical to use Some in a case which can never be null?

tmspzz:
> why is Some used here in this case? the declaration is a hardcoded value.
> 
> isn't this performance wise impractical to use Some in a case which can never be null?

It can be null. The default value was in fact null, which via `null.asInstanceOf[T]` where T is Boolean translated to false.

PR #3611, opened 2021-11-30
Notifications: Refactoring to make initialization lighter when it's triggered by a notification (SQCORE-1146)

makingthematrix:
https://wearezeta.atlassian.net/browse/SQCORE-1146

When the app starts, either because the user clicked on the icon, or because of an incoming notification, the method `WireApplication.ensureInitialized()` is called.

The original idea was to split the initialization into two parts - one with the all necessary components that need to be initialized immediately, the other with components that are needed only when the app goes to the front and can be initialized as a response to that. 
But it turns out the initialization is too complicated. Components depend on each other in a non-trivial way, and this ticket shouldn't try to untangle and fix it all. So instead I looked into logs to learn what happens when the app is initialized through a notification and tried to minimize what I saw there while not risking that one of the components will stop working because I turned off something it depended on.

I also included a small fix in `NotificationHandlerService` for when a notification is cancelled by swiping it out from the screen. When the bug was there, it was possible that a specific notification was removed but the "bundle" notification wrapped around it stayed in the tray, displayed as an empty rectangle.

### DIFF
--- a/app/src/main/scala/com/waz/services/notifications/NotificationsHandlerService.scala
+++ b/app/src/main/scala/com/waz/services/notifications/NotificationsHandlerService.scala
@@ -17,7 +17,7 @@
  */
 package com.waz.services.notifications
 
-import android.app.{NotificationManager, PendingIntent}
+import android.app.PendingIntent
 import android.content.{Context, Intent}
 import androidx.core.app.RemoteInput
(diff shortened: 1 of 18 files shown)


AndroidBob:
Build Success
ID:1
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3611/1/)

PR #3662, opened 2022-02-08
fix(build): remove 'null' from F-Droid build version strings.

sysvinit:
----

- The **PR Title**
  - [ ] contains a reference JIRA issue number like `SQPIT-764`

- The **PR Description**
----

# What's new in this PR?

### Issues

The version string embedded in the F-Droid flavoured APK's contains a trailing "null", which does not match with the versions which are published on GitHub. This causes the F-Droid project's automated builds to fail on a lint pass which checks whether the versions match -- see [this F-Droid merge request](https://gitlab.com/fdroid/fdroiddata/-/merge_requests/10237) for more context.

### Causes

The build configuration for the F-Droid flavour uses a `version_name_postfix` field which is uninitialised in the build configuration, which is then converted into "null" due to a string type conversion.

### Solutions

Removing the `version_name_postfix` from the F-Droid build, as this field is unused for this flavour.

### Testing

I've disassembled an APK built with and without this change applied to inspect the embedded version metadata, and the trailing "null" is no longer present after this change.

### Notes

Thanks to @Gardosen for helping to find the cause of this bug.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -242,7 +242,6 @@ android {
 
         fdroid {
             applicationId buildtimeConfiguration.configuration.fdroid.applicationId
-            versionName android.defaultConfig.versionName + buildtimeConfiguration.configuration.fdroid.version_name_postfix
             manifestPlaceholders = [applicationLabel       : buildtimeConfiguration.configuration.fdroid.appName,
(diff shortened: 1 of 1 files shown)


vitorhugods (approved):

AndroidBob:
Build Success
ID:1
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3662/1/)

PR #3690, opened 2022-04-05
fix: display domain in handle if federation is enabled FS-506

marcoconti83:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-506" title="FS-506" target="_blank">BugFS-506</a>  [Android] Issues with API versioning staging
  </td></table>
  <br />
 

----

- The **PR Title**
  - [x] contains a reference JIRA issue number like `SQPIT-764`

- The **PR Description**
----

# What's new in this PR?

### Issues

There was an automation failure: 
`C352687 - We do also see [xyz@staging.zinfra.io](mailto:xyz@staging.zinfra.io) in settings/account`

The @handle in the profile screen contained the domain even for backends where the federation was not enabled. This is because the code was checking if the domain of the user displayed is the same as the domain of the user logged in. However, sometimes (probably when the federation is disabled?) the domain of the logged in user is empty, which is of course not equal to any domain that is not empty. So users with a domain would always display a domain, even if it's in theory the same domain as the user and federation is disabled.

### Solutions

There is a function to compute the display handle (`displayHandle`) with a boolean parameter `forceDomain` . I replaced it with a policy enumeration:

- `NeverShowDomain`
- `AlwaysShowDomain`
- `ShowIfNotSame`

The `displayHandle` function was called in various parts of the code. I followed this strategy:
- When federation is not enabled, I will pass in `NeverShowDomain`. Users are guaranteed to be from the same backend anyway.
- When federation is enabled, if it's a UI component that is supposed to show the domain only if it's not the same as the logged in user (e.g. conversations, search results), I pass `ShowIfNotSame`. Hopefully in this case, the logged in user has a domain as federation is enabled.
- When federation is enabled, if it's a UI component that is always supposed to show the domain (e.g. profile view), I pass `AlwaysShowDomain`

I'm not 100% sure that I got all the specs right on where we should display the domain even if it's the same as the logged in user. So after we merge this PR we should be on the lookout for QA automation breaking in other places for @handles. Good news is that now the code is more clear (at least to me :D ) so I can fix it easily if it breaks somewhere else.

### Testing

I did some manual testing of the UI, as it's not possible to automate UI testing with the current setup. Nightly QA automation will tell us if anything else is broken.

### DIFF
--- a/app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
+++ b/app/src/main/scala/com/waz/zclient/common/views/SingleUserRowView.scala
@@ -26,7 +26,7 @@ import android.view.{Gravity, View, ViewGroup}
 import android.widget.{CompoundButton, ImageView, LinearLayout, RelativeLayout}
 import androidx.appcompat.widget.AppCompatCheckBox
 import com.waz.model.otr.ClientId
-import com.waz.model.{Availability, IntegrationData, UserData, UserId}
+import com.waz.model.{Availability, DisplayHandleDomainPolicies, IntegrationData, UserData, UserId}
 import com.waz.service.BackendConfig
 import com.waz.threading.Threading
 import com.waz.threading.Threading._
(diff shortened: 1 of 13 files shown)


AndroidBob:
Build Failure
ID:1
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3690/1/)

marcoconti83:
Looks like I need to fix a couple more call sites for the function that I changed. It's not compiling on the build machine. It compiled on my machine... until I did a `gradle clean` :-(

AndroidBob:
Build Success
ID:2
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3690/2/)

vitorhugods (approved):
👨🏼‍🍳🤌🏼

PR #3695, opened 2022-04-07
fix: live update of federation configuration FS-512

marcoconti83:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-512" title="FS-512" target="_blank">BugFS-512</a>  [Android] Issues with API versioning federation
  </td></table>
  <br />
 

----

- The **PR Title**
  - [x] contains a reference JIRA issue number like `SQPIT-764`

- The **PR Description**
----

# What's new in this PR?

Following up on https://github.com/wireapp/wire-android/pull/3686, there are still some instances in the code that, as described in the aforementioned PR, capture the value of the "federation enabled" flag at creation time, when that value might be not updated yet.

### Issues

This had many symptoms:

- Could not send connection requests to a remote user when the app is first started
- Could not place a call to the webapp
- ... other minor effects mentioned in FS-512 and FS-506

### Causes (Optional)

Often, a class constructor would be passed a `FederationSupported` instance, which is just a wrapper for a boolean initially set to false until we get the API information from the backend. 

The `FederationSupported` instance is then never modified even if we later discover that the backend supports federation.

An example of this is visible in `zmessaging/src/main/scala/com/waz/service/ConnectionService.scala`.

There were other ways in which the `FederationSupported` in `GlobalModule` was accessed, which might have had the same issue. Therefore I removed them.

### Solutions

Pass a `Signal` of the backend configuration rather than a value. When using the signal, the call site has the most updated value of the federation flag.

### Testing

I ran the app on production, staging and Anta environment. Calling the webapp worked, and the UI seemed to show the right values, with the exception of the automation issue C925125, which I will keep investigating and address in another PR.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/GlobalModule.scala
+++ b/zmessaging/src/main/scala/com/waz/service/GlobalModule.scala
@@ -201,10 +201,9 @@ final class GlobalModuleImpl(val context:             AContext,
 
   lazy val httpProxy:           Option[Proxy]                    = HttpProxy(metadata, defaultProxyDetails).proxy
 
-  private def _backendConfiguration = new SourceSignal[BackendConfig](None)
+  private val _backendConfiguration = new SourceSignal[BackendConfig](Option(backend))
   override def backendConfiguration: Signal[BackendConfig] = _backendConfiguration
 
-  _backendConfiguration ! backend
 }
 
 class EmptyGlobalModule extends GlobalModule {
--- a/zmessaging/src/main/scala/com/waz/service/ZMessaging.scala
+++ b/zmessaging/src/main/scala/com/waz/service/ZMessaging.scala
@@ -111,9 +111,7 @@ class ZMessaging(val teamId:    Option[TeamId],
   val clock = ZMessaging.clock
 
   val global     = account.global
-
-  //def federation2: FederationSupport = global.backend.federationSupport
-
+  
   lazy val selfUserId: UserId = account.userId
   lazy val selfDomain: Domain = account.domain
 
--- a/zmessaging/src/main/scala/com/waz/service/ZMessaging.scala
+++ b/zmessaging/src/main/scala/com/waz/service/ZMessaging.scala
@@ -112,7 +112,7 @@ class ZMessaging(val teamId:    Option[TeamId],
 
   val global     = account.global
 
-  def federation: FederationSupport = global.backend.federationSupport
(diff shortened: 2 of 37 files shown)


zmessaging/src/main/scala/com/waz/service/GlobalModule.scala
```diff
@@ -201,10 +201,9 @@ final class GlobalModuleImpl(val context:             AContext,
 
   lazy val httpProxy:           Option[Proxy]                    = HttpProxy(metadata, defaultProxyDetails).proxy
 
-  private def _backendConfiguration = new SourceSignal[BackendConfig](None)
+  private val _backendConfiguration = new SourceSignal[BackendConfig](Option(backend))
```
marcoconti83, line 204: This was an interesting mistake: being a `def`, it would re-create an empty signal every time it was accessed. Now a `val` with a non-empty initial state, this behaves as expected.

AndroidBob:
Build Failure
ID:1
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3695/1/)

marcoconti83:
... and again, it compiles locally but not on the build machine, because I did not run a `gradle clean` after every change 😠 .
I am fixing the compilation now.

AndroidBob:
Build Failure
ID:2
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3695/2/)

zmessaging/src/main/scala/com/waz/service/ZMessaging.scala
```diff
@@ -112,7 +112,7 @@ class ZMessaging(val teamId:    Option[TeamId],
 
   val global     = account.global
 
-  def federation: FederationSupport = global.backend.federationSupport
+  //def federation2: FederationSupport = global.backend.federationSupport
```
johnxnguyen, line 115: To delete?

johnxnguyen (approved):

zmessaging/src/main/scala/com/waz/service/ZMessaging.scala
marcoconti83: Yes, thanks!

AndroidBob:
Build Success
ID:4
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3695/4/)

PR #3701, opened 2022-04-14
fix: do not crash at startup if the backend URL is not set FS-421

marcoconti83:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/FS-421" title="FS-421" target="_blank">TaskFS-421</a>  [Android] Clients need to know if the backend support federated endpoints or not
  </td></table>
  <br />
 

# What's new in this PR?

### Issues

Some build configuration have no backend API URL set, because it will be set anyway to a custom backend later via deep link or other configuration mechanism.

The app would crash when trying to parse the URL for the `/api-version` request made at startup, because the host is null (since the backend URL is empty).

### Solutions

If the backend URL is empty, do not attempt to make a request. The request will be made later anyway when the backend is set to another value.

I also removed an unnecessary, unused field `domain` from the backend config. If we need a domain, we should NOT infer it from the backend URL but take the one from the `api-version` response.

### DIFF
--- a/zmessaging/src/main/scala/com/waz/service/BackendConfig.scala
+++ b/zmessaging/src/main/scala/com/waz/service/BackendConfig.scala
@@ -39,7 +39,6 @@ final class BackendConfig(private var _environment: String,
 
   def environment: String = _environment
   def baseUrl: URI = _baseUrl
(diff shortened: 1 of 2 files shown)


AndroidBob:
Build Success
ID:1
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3701/1/)

ohassine (approved):
👍

PR #3710, opened 2022-06-07
fix: properly handle SSO types where the user has a password [SQSERVICES-1565]

vitorhugods:
----

- The **PR Title**
  - [X] conforms to the style of semantic commits messages¹ supported in Wire's Github Workflow²
  - [X] contains a reference JIRA issue number like `SQPIT-764`
  - [X] answers the question: _If merged, this PR will: ..._ ³

- The **PR Description**
  - [X] is free of optional paragraphs and you have filled the relevant parts to the best of your ability
----

# What's new in this PR?

### Issues

- App asks for passwords and when it shouldn't, for example when removing devices.

### Causes

App was only using one check (`isSSO`, or `ssoId.isDefined`) in order to make this kind of decisions.
But some kinds of SSO don't have SAML-based credentials, for example when the email is whitelisted by some external entity, but the password is managed within wire.

### Solutions

Instead of only checking `isSSO`, we need two different checks:
- `sso.id != null` to know if the account is managed by a 3rd party
- `sso.subject != null && sso.subject.isNotEmpty` to know if the password is also managed by a 3rd party

The first condition is to be used on things like Delete account or change e-mail (we don't want any SSOO users to be able to do that).

The second one is for password-related stuff. Making it mandatory or not when removing clients, showing or not the option to reset the password, etc.

### Testing

Manually tested

#### Test Coverage (Optional)

- [ ] I have added automated test to this contribution

#### How to Test

Log-in with a different accounts:
* SCIM-managed (SSO with password)
* SAML-managed (SSO without password)
* Regular account (non-SSO)

Attempt to do the following on each:
- Remove another device
- Reset your password
- Delete your account
- Logout

----
#### PR Post Merge Checklist for internal contributors

 - [X] If any soft of configuration variable was introduced by this PR, it has been added to the relevant documents and the CI jobs have been updated.
----
##### References
1. https://sparkbox.com/foundry/semantic_commit_messages
1. https://github.com/wireapp/.github#usage
1. E.g. `feat(conversation-list): Sort conversations by most emojis in the title #SQPIT-764`.

### DIFF
--- a/app/src/main/scala/com/waz/zclient/MainActivity.scala
+++ b/app/src/main/scala/com/waz/zclient/MainActivity.scala
@@ -354,10 +354,10 @@ class MainActivity extends BaseActivity
               isNewClient <- z.userPrefs(IsNewClient).apply()
               pendingPw <- z.userPrefs(PendingPassword).apply()
               pendingEmail <- z.userPrefs(PendingEmail).apply()
-              ssoLogin <- accountsService.activeAccount.map(_.exists(_.ssoId.isDefined)).head
+              passwordManagedBySSO <- accountsService.activeAccountHasCompanyManagedPassword.head
             } yield {
               val (f, t) =
-                if (ssoLogin) {
+                if (passwordManagedBySSO) {
                   if (self.handle.isEmpty) (SetHandleFragment(), SetHandleFragment.Tag)
                   else (new MainPhoneFragment, MainPhoneFragment.Tag)
                 }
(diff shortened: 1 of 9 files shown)


AndroidBob:
Build Success
ID:1
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3710/1/)

AndroidBob:
Build Success
ID:2
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3710/2/)

tmspzz (approved):

PR #3731, opened 2022-09-03
fix: long delays when updating version AN-7156

marcoconti83:
<table>
<td>
  <a href="https://wearezeta.atlassian.net/browse/AN-7156" title="AN-7156" target="_blank">BugAN-7156</a>  App stuck on syncing
  </td></table>
  <br />
 

# What's new in this PR?

This is a follow up to https://github.com/wireapp/wire-android/pull/3730.

### Issues

As mentioned in #3730 

> It may be possible that there are users who have accumulated so many rows in the `EncryptedPushNotificationEvents` table that the application will hang before the above fix has a chance to start deleting events after they are decrypted (and then processed).

The app might take a long time to recover after migrating to a new version. This looks to the user like the app is stuck (while it is in fact processing a LOT of messages in the background).

### Causes

This happens when a LOT of messages have accumulated over time in the `EncryptedPushNotificationEvents` table. This is because before #3730 those events were never discarded, they kept accumulating over time.

### Solutions

This PR introduces a DB migration that simply deletes the content of `EncryptedPushNotificationEvents`. The migration is run only once, upon update. There is no data loss: if the events are still in `EncryptedPushNotificationEvents`, they have either:

- (correctly) not been processed
- have been processed but are (incorrectly) still in the table due to the bug addressed in #3730

If they have not been processed, then the ID of the last event that was processed has not been updated to include those. This means that the next time the app is opened, it will use an older last event ID to fetch new events, causing those events to be fetched again. Therefore deleting them is no problem.

If the event have been processed already, then deleting them is no problem.

The PR leaves the content of `DecryptedPushNotificationEvents` as is. This table might also have grown substantially due to the `EncryptedPushNotificationEvents` being copied over, however deleting messages from this table might result in them being lost forever (since they can't be decrypted again). Therefore we will wipe this table only if it turns out that deleting the encrypted table is not enough to reduce the delay.

### Testing

Pending. We want to see how this PR affects the performance of devices that are affected by the issue, which we can't easily reproduce in a test setup. Therefore we will test on real devices.

#### How to Test

Install this build as a replacement of a build that was showing symptoms of taking a long time to synchronize on start.

### DIFF
--- a/storage/src/main/kotlin/com/waz/zclient/storage/db/UserDatabase.kt
+++ b/storage/src/main/kotlin/com/waz/zclient/storage/db/UserDatabase.kt
@@ -97,7 +97,7 @@ abstract class UserDatabase : RoomDatabase() {
     abstract fun buttonsDao(): ButtonsDao
 
     companion object {
-        const val VERSION = 139
+        const val VERSION = 140
 
         @JvmStatic
         val migrations = arrayOf(
@@ -112,7 +112,8 @@ abstract class UserDatabase : RoomDatabase() {
             USER_DATABASE_MIGRATION_135_TO_136,
             USER_DATABASE_MIGRATION_136_TO_137,
             USER_DATABASE_MIGRATION_137_TO_138,
-            USER_DATABASE_MIGRATION_138_TO_139
+            USER_DATABASE_MIGRATION_138_TO_139,
+            USER_DATABASE_MIGRATION_139_TO_140
         )
     }
 }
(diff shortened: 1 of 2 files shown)


AndroidBob:
Build Failure
ID:1
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3731/1/)

AndroidBob:
Build Failure
ID:2
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3731/2/)

AndroidBob:
Build Failure
ID:3
URL:[Link to Buildjob](https://10.10.124.134/job/wire-android/job/PR-3731/3/)

typfel (approved):
👍

